PR #16, opened 2020-07-02
R4R: system contract upgrade framework and upgrade tokenhub

HaoyangLiu:


core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+	systemRewardContract       = common.HexToAddress("0x0000000000000000000000000000000000001002")
+	lightClientContract        = common.HexToAddress("0x0000000000000000000000000000000000001003")
+	tokenHubContract           = common.HexToAddress("0x0000000000000000000000000000000000001004")
+	relayerIncentivizeContract = common.HexToAddress("0x0000000000000000000000000000000000001005")
+	relayerHubContract         = common.HexToAddress("0x0000000000000000000000000000000000001006")
+	govHubContract             = common.HexToAddress("0x0000000000000000000000000000000000001007")
+	tokenManagerContract       = common.HexToAddress("0x0000000000000000000000000000000000001008")
+	crossChainContract         = common.HexToAddress("0x0000000000000000000000000000000000002000")
+
+	mainnetUpgrapdeConfig = make(map[int64]Upgrade)
```
guagualvcha, line 41: please reuse the params here. https://github.com/binance-chain/bsc/blob/b2874e27e2116f69294a366fdb38c6ccba3bfca0/params/config.go#L216

Chapel is testnet, rialto is dev net

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+	// setup mainnet upgrade config
+
+	/* mainnet upgrade config */
+
+	// setup testnet upgrade config
+
+	/* testnet upgrade config */
+
+	// setup testnet upgrade config
+	devnetUpgrapdeConfig[20000] = Upgrade{
```
guagualvcha, line 56: can not just simple define `devnetUpgrapdeConfig[20000]` here.  Because `20000` is treat as fork height here, we must define the fork height on chainConfig, like  dao fork get `DAOForkBlock` in config. Otherwise will break the handshake protocol about forkId.

guagualvcha: We need define a map like  `map[forkname] []*UpgradeConfig` here. Then define different height for the fork for different network on `Params` file

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,103 @@
...
+		return
+	}
+
+	upgrade, ok := upgradeCfg[blockNumber.Int64()]
+	if !ok {
+		return
+	}
+	fmt.Println(fmt.Sprintf("Apply upgrade %s at height %d", upgrade.UpgradeName, blockNumber.Int64()))
+	for _, cfg := range upgrade.Configs {
+		fmt.Println(fmt.Sprintf("Upgrade contract %s to commit %s", cfg.ContractAddr.String(), cfg.CommitUrl))
```
guagualvcha, line 95: try use log instead of `fmt`

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+
+const (
+	mainnet = "Mainnet"
+	chapel  = "Chapel"
+	rialto  = "Rialto"
+)
+
+var (
+	// genesis contracts
+	validatorContract          = common.HexToAddress("0x0000000000000000000000000000000000001000")
```
guagualvcha, line 39: we have redefined the system contract address here, since parlia have done it too. But I don't know where to put it better either.

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+func UpgradeBuildInSystemContract(config *params.ChainConfig, blockNumber *big.Int, statedb *state.StateDB) {
+	if config == nil || blockNumber == nil || statedb == nil {
+		return
+	}
+	var network string
+	switch core.GenesisHash {
+	/* Add mainnet genesis hash */
+	case params.ChapelGenesisHash:
+		network = chapel
+	case params.RialtoGenesisHash:
```
guagualvcha, line 78: if no match with any network, just return

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func applySystemContractUpgrade(upgrade *Upgrade, blockNumber *big.Int, statedb *state.StateDB, logger log.Logger) {
+	if upgrade == nil {
+		return
+	}
+
+	logger.Info(fmt.Sprintf("Apply upgrade %s at height %d", upgrade.UpgradeName, blockNumber.Int64()))
+	for _, cfg := range upgrade.Configs {
+		logger.Info(fmt.Sprintf("Upgrade contract %s to commit %s", cfg.ContractAddr.String(), cfg.CommitUrl))
```
guagualvcha, line 99: ==> `		logger.Info(fmt.Sprintf("Upgrade contract  to commit ", "address", cfg.ContractAddr.String(),"commit url", cfg.CommitUrl))
`

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+		apply other upgrades
+	*/
+}
+
+func applySystemContractUpgrade(upgrade *Upgrade, blockNumber *big.Int, statedb *state.StateDB, logger log.Logger) {
+	if upgrade == nil {
+		return
+	}
+
+	logger.Info(fmt.Sprintf("Apply upgrade %s at height %d", upgrade.UpgradeName, blockNumber.Int64()))
```
guagualvcha, line 97: The log style here violate how the log being used.

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+	CommitUrl     string
+	Code          string
+}
+
+type Upgrade struct {
+	UpgradeName string
+	Configs     []*UpgradeConfig
+}
+
+type beforeUpgrade func(blockNumber *big.Int, contractAddr common.Address, statedb *state.StateDB)
```
guagualvcha, line 28: there is no need to define type `beforeUpgrade ` and `afterUpgrade`, `upgradeHook` is enough.

core/systemcontractupgrade/system-contract-upgrade.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, cfg := range upgrade.Configs {
+		logger.Info(fmt.Sprintf("Upgrade contract %s to commit %s", cfg.ContractAddr.String(), cfg.CommitUrl))
+
+		if cfg.BeforeUpgrade != nil {
+			cfg.BeforeUpgrade(blockNumber, cfg.ContractAddr, statedb)
+		}
+
+		newContractCode, err := hex.DecodeString(cfg.Code)
+		if err != nil {
+			logger.Info(fmt.Sprintf("failed to decode new contract code: %s", err.Error()))
```
guagualvcha, line 107: Let's panic here if there is an error. It is a very serious bug and need stop the node if happened

core/systemcontractupgrade/system-contract-upgrade.go
guagualvcha: please add `return (error)` to this hook.

HaoyangLiu: > there is no need to define type `beforeUpgrade ` and `afterUpgrade`, `upgradeHook` is enough.

I don't think upgradeHook is enough. Sometime we have to do cleanup work before update code, and sometime we have to do init work after update code

core/systemcontractupgrade/system-contract-upgrade.go
HaoyangLiu: `commit url` contains space. Besides, I think no need to follow this style.

core/systemcontractupgrade/system-contract-upgrade.go
HaoyangLiu: System addresses in parlia.go are not completed. Besides, I think it would be better to defind all addresses in here ranther than in consensus engine.

core/systemcontractupgrade/system-contract-upgrade.go
HaoyangLiu: Sorry for misunderstand. We can just define one function type`upgradeHook` but multiple instances
```
type UpgradeConfig struct {
	BeforeUpgrade ugradeHook
	AfterUpgrade  ugradeHook
        ...
}
```

core/systemcontractupgrade/system-contract-upgrade.go
HaoyangLiu: In the further development, we might need to start a local network. This local network might also need system upgrade, so I added a default network type.

core/genesis.go
```diff
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
```
guagualvcha, line 25: imports order here.

PR #36, opened 2020-09-10
R4R: add max gas allwance calculation

HaoyangLiu:


guagualvcha (approved):

internal/ethapi/api.go
```diff
@@ -887,6 +887,11 @@ func DoEstimateGas(ctx context.Context, b Backend, args CallArgs, blockNrOrHash
 		hi  uint64
 		cap uint64
 	)
+	// Use zero address if sender unspecified.
+	if args.From == nil {
+		args.From = new(common.Address)
```
thanhnguyennguyen, line 892: should `return 0, err `
from == nil || from == common.HexToAddress('0x0')
both are meaningless

guagualvcha (approved):

yutianwu (approved):

Issue #52, opened 2020-11-24
Fatal: ./config.toml, line 46: field 'omitempty' is not defined in node.Config


alberentsen:
I follow the instruction to set-up a node on https://docs.binance.org/smart-chain/developer/fullnode.html and get the following fatal error:

Fatal: ./config.toml, line 46: field 'omitempty' is not defined in node.Config, see https://godoc.org/github.com/ethereum/go-ethereum/node#Config for available fields

Here are the steps that I follow from the website:
## Install BSC node
git clone https://github.com/binance-chain/bsc
## Enter the folder bsc was cloned into
cd bsc
## Compile and install bsc
make geth
## mainet
wget --no-check-certificate  $(curl -s https://api.github.com/repos/binance-chain/bsc/releases/latest |grep browser_ |grep mainnet |cut -d\" -f4)
unzip mainnet.zip
geth --datadir node init genesis.json
## start a full node
geth --config ./config.toml --datadir ./node --pprofaddr 0.0.0.0 --metrics --pprof
Here the error occurs.

## System information
Get Version: 1.9.24-stable
Git Commit: cc05b050df5f88e80bb26aaf6d2f339c49c2d702
Architecture: amd64
Protocol Versions: [65 64 63]
Go Version: go1.15.5
Operating System: linux
GOPATH=/home/aber/go
GOROOT=/usr/local/go

guagualvcha:
I try to follow your steps, and the version is:

```
Geth
Version: 1.0.4
Git Commit: ed10dfa6bf8d1175e6fb4d2e79997f60fb9627a5
Architecture: amd64
Protocol Versions: [65 64 63]
Go Version: go1.15.5
Operating System: darwin
GOPATH=/Users/baifudong/workspace
GOROOT=/usr/local/go
```
guess something wrong when build, the binary should under `build/bin`.

And we did have a false config in  `mainnet.zip` and correct it now.

Issue #51, opened 2020-11-23
Is there any interest in a ethstats-like system to track node performance?


jimtje:
Hi there,

Since this is more or less of an interest check I hope you forgive m for eschewing the template.

Out of frustration as available monitoring tools for comparison purposes to debug and diagnose issues I've forked the aleth.io set up clients to adapt it to BSC. Of course. Of course, 1 node (mine) isn't a great sample size. Is there a need in the community for this? On the eth side I've found great value for this but I literally cannot tell what the BSC lanscape is truly like since it's, frankly, not entirely transparent and albeit being a Chinese national who has been taught (brainwashed?) into thinking that the process doesn't matter, it does, and this is transparency and process at work. (And don't worry, I'm in the state with no plans to leave, have a law degree, and am thoroughly disgusted at what they do, particularly coming from a place of privilege over there that taught me that decentralized self-deployable small clusters of monitoring beats a.SAAS solution for privacy concerns.

But thre's no point in working on this if nobody care so this is an interest check really. There are bugs to be worked out and I'm admittedly much more a backend guy than a frontend guy. This is really a quick strawpoll to see what sentiments are out there. And my apologies in advance to the Chinese nationals still in China seeing this, but I don't think any of this is consider 敏感。 Comments in Chinese or English are both cool, thanks.

huangsuyu:
I am not sure I understand you correctly. 请问能不能用中文复述一遍。

jimtje:
因为现在好像没有什么服务可以让我从网上及时看到我的full node的情况，而且币安从来不回我的email, 也不提供任何监视办法。同时，它们尽然还通过我跟它们联系的机会来找我麻烦，把我的账户关了一周，莫名其妙，因为我在国外但是我身份证留在了国内。因为我不认为它们对辅助系统有太多兴趣我也不想多跟她们打交道，但是想问问如果我自己开了想ethstats.io一样的服务或者其他的类似系统，有没有人真的会用，还就是我一个？

AlphaStop:
Yes, absolutely! I could definitely use this

PR #53, opened 2020-12-08
[R4R]support fork id in header; elegant upgrade

guagualvcha:
# BEP-87: Visual Fork of Binance Smart Chain

- BEP-87: Visual Fork of Binance Smart Chain
  - [1. Summary](#1-summary)
  - [2. Abstract](#2-abstract)
  - [3. Status](#3-status)
  - [4. Motivation](#4-motivation)
  - [5. Specification](#5-specification)
    - [5.1 Fork Hash](#51-fork-hash)
    - [5.2 Vanity](#52-vanity)
    - [5.3 Vanity](#53-clients)
    - [5.4 Backwards Compatibility](#54-backwards-compatibility)
  - [6. License](#6-license)

## 1.  Summary 

This BEP describes a proposal to enable the chain to display the whole view of validators that on different upcoming forks.  

## 2.  Abstract

The four bytes of `Header.Extra[28:32]` will be fulfilled with `NEXT_FORK_HASH`. The `NEXT_FORK_HASH` indicates which fork the signer of this block is going to stay on. By analysing `N` (`N` is the amount of validators) continuous block headers, we are able to know which fork is supported by the majority of validators and exact which validator has not upgraded yet. 

## 3.  Status

Draft.

## 4.  Motivation

Binance Smart Chain will have some hard forks inevitably in the long run. Binance Smart Chain takes a risk of halt during hard fork once the validators can not reach a consensus. A validator could be slashed if it is not on the canonical fork which could be avoidable if the maintainer received alerts in time. A new protocol should be introduced to enable the chain to display the whole view of validators that on different upcoming forks. Any nodes/validators can decide to upgrade/fork or not accordingly.

###  5.1 Fork Hash

Fork Hash is introduced in [EIP-2124](https://eips.ethereum.org/EIPS/eip-2124). It is a mechanism that can let ethereum nodes precisely identify whether another node will be useful. 

- `FORK_HASH`. IEEE CRC32 checksum ([4]byte) of the genesis hash and fork blocks numbers that already passed. E.g. Fork Hash for mainnet would be: `forkhash₂ = 0x91d1f948 (DAO fork) = CRC32(<genesis-hash> || uint64(1150000) || uint64(1920000))`.
- `FORK_NEXT`. Block number (uint64) of the next upcoming fork, or 0 if no next fork is known.
- `NEXT_FORK_HASH`, whose algorithm is same with `FORK_HASH`, but `FORK_NEXT` will be included as well if it is not 0.

### 5.2 Vanity

Format of `Header.Extra`: 

```
|   32 bytes       |       20 * N bytes      |   65 bytes     |
|   extraVanity    |   validatorSetBytes     |   extraSeal    |
```

`extraVanity` field is customized now, validator can use it to record `NEXT_FORK_HASH` of itself. The `NEXT_FORK_HASH` will only use the last 4 bytes of `extraVanity`.

### 5.3 Clients

- For validator. It will fill in `Header.Extra` with `NEXT_FORK_HASH` during preparing block header.
- For witness. It will log a warning message if the majority `NEXT_FORK_HASH` is different from local.
- For light client. No impact.

### 5.4 Backwards Compatibility

- This BEP itself is not a hardfork one, it breaks nothing of consensus. 
- Downstream service is completely compatible with this BEP.

## 6. License

The content is licensed under [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

### DIFF
--- a/consensus/parlia/snapshot.go
+++ b/consensus/parlia/snapshot.go
@@ -18,6 +18,7 @@ package parlia
 
 import (
 	"bytes"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"math/big"
@@ -38,10 +39,11 @@ type Snapshot struct {
 	ethAPI   *ethapi.PublicBlockChainAPI
 	sigCache *lru.ARCCache // Cache of recent block signatures to speed up ecrecover
 
-	Number     uint64                      `json:"number"`     // Block number where the snapshot was created
-	Hash       common.Hash                 `json:"hash"`       // Block hash where the snapshot was created
-	Validators map[common.Address]struct{} `json:"validators"` // Set of authorized validators at this moment
-	Recents    map[uint64]common.Address   `json:"recents"`    // Set of recent validators for spam protections
+	Number           uint64                      `json:"number"`             // Block number where the snapshot was created
+	Hash             common.Hash                 `json:"hash"`               // Block hash where the snapshot was created
+	Validators       map[common.Address]struct{} `json:"validators"`         // Set of authorized validators at this moment
+	Recents          map[uint64]common.Address   `json:"recents"`            // Set of recent validators for spam protections
+	RecentForkHashes map[uint64]string           `json:"recent_fork_hashes"` // Set of recent forkHash
 }
 
 // newSnapshot creates a new snapshot with the specified startup parameters. This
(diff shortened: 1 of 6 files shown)


consensus/parlia/snapshot.go
```diff
@@ -121,9 +125,22 @@ func (s *Snapshot) copy() *Snapshot {
 	for block, v := range s.Recents {
 		cpy.Recents[block] = v
 	}
+	for block, id := range s.RecentForkHashes {
+		cpy.RecentForkHashes[block] = id
+	}
 	return cpy
 }
 
+func (s *Snapshot) isMajorityFork(header *types.Header, forkHash string) bool {
```
HaoyangLiu, line 134: `header *types.Header` seems useless here.

guagualvcha: fixed.

yutianwu (dismissed):

consensus/parlia/snapshot.go
```diff
@@ -121,9 +125,22 @@ func (s *Snapshot) copy() *Snapshot {
...
 }
 
+func (s *Snapshot) isMajorityFork(forkHash string) bool {
+	ally := 0
+	for _, h := range s.RecentForkHashes {
+		if h == forkHash {
+			ally++
+		}
+	}
+	return ally > len(s.RecentForkHashes)/2
```
HaoyangLiu, line 141: Do we need to ensure if the length of `s.RecentForkHashes` is closed to the length of current validatorset?

guagualvcha: updated.

HaoyangLiu (approved):

yutianwu (approved):

PR #56, opened 2020-12-18
R4R: apply mirror sync upgrade

HaoyangLiu:
### Description

Implement genesis contract upgrade for  https://github.com/binance-chain/bsc-genesis-contract/pull/85 and https://github.com/binance-chain/BEPs/pull/84

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues


params/config.go
```diff
@@ -213,6 +214,26 @@ var (
 		Threshold: 2,
 	}
 
+	BSCChainConfig = &ChainConfig{
+		ChainID:             big.NewInt(97),
```
guagualvcha, line 218: Chain id 56 for main-net

guagualvcha:
Because we hardcode the byte code into the file. We are not quite confident that the byte code is exactly what we want, we will take a great risk(just like the hard fork we have done on testnet while the chain id of the contract is wrong). We need to come up with a solution to double/triple  verify the byte code is correct before performing the hardfork.

guagualvcha (dismissed):

yutianwu (approved):

guagualvcha (approved):

yutianwu (approved):

PR #101, opened 2021-03-04
R4R: limit range get log

guagualvcha:
### Description

`GetLog` API with unlimited block range will consume huge CPU and IO resource, which has import great pressure to our data seed nodes. As a result, our data seed node always can't sync recent blocks.

### Rationale

Add `--rangelimit` flag to our geth. If a node started with this flag, a default block range limit will be applied to `GetLog` API, thus the pressure to BSC node will be reduced.

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -504,7 +504,7 @@ func (b *SimulatedBackend) FilterLogs(ctx context.Context, query ethereum.Filter
 			to = query.ToBlock.Int64()
 		}
 		// Construct the range filter
-		filter = filters.NewRangeFilter(&filterBackend{b.database, b.blockchain}, from, to, query.Addresses, query.Topics)
+		filter = filters.NewRangeFilter(&filterBackend{b.database, b.blockchain}, from, to, query.Addresses, query.Topics, false)
 	}
 	// Run the filter and return all the logs
 	logs, err := filter.Logs(ctx)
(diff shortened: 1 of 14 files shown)


yutianwu (approved):

saurik:
This essentially breaks the ability to write sane GUIs using your chain, at least if the user is going to use your public endpoints :(. I do understand the interest in limiting the amount of I/O from these calls, but I will *strongly* argue that the correct way to approach this problem is to limit the number of results or the complexity of the query, not having a hard cap on the range of requests.

The way this works on Alchemy's Ethereum endpoints, for example, is that they have a hybrid strategy: they have an even tighter range if you expect to get a *lot* (and it is a lot...) of logs, but if you aren't getting many logs (which is going to, by far, be the common case: these queries should be "easy") you have no range limit (allowing you to build a way to quickly scour history for rare events, or do "paging" to find all events).

> You can make eth_getLogs requests with up to a 2K block range and no limit on the response size. You can also request any block range with a cap of 10K logs in the response.

To put it strongly: if my contract has logged five events ever for some specific log index, *the whole point of this feature* is that I am paying a non-trivial amount of gas (aka, money ;P these logs cost 375 gas per index and 256 gas per word) for nodes to index this information and let me find it later: if Binance is going to run public endpoints for clients, but not let me find that information, they just aren't really correctly functional.

Think about what this means for dapps that are trying to show you the history of multisig transactions on your contract: BSC's 5000 block "range limit", given your very fast three second block time, means that it could only look at just over four hours of data per query... to find some confirmation event that happened even just in the past month would require almost two *hundred* requests!

PR #102, opened 2021-03-05
add metrics for distinct call

yutianwu:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -19,8 +19,10 @@ package ethapi
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	lru "github.com/hashicorp/golang-lru"
 	"math/big"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/ethapi/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package ethapi
+
+import (
+"fmt"
+
+"github.com/ethereum/go-ethereum/metrics"
```
HaoyangLiu, line 22: go fmt

internal/ethapi/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func newDistinctRpcCallRequestGauge() metrics.Gauge {
+	m := fmt.Sprintf("rpc/distinct_count/call")
```
HaoyangLiu, line 26: Why not just `m := "rpc/distinct_count/call"`

internal/ethapi/api.go
```diff
@@ -19,8 +19,10 @@ package ethapi
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	lru "github.com/hashicorp/golang-lru"
```
HaoyangLiu, line 25: go fmt

internal/ethapi/api.go
```diff
@@ -876,6 +887,30 @@ func (s *PublicBlockChainAPI) Call(ctx context.Context, args CallArgs, blockNrOr
...
+
+	type CallReq struct {
+		Args CallArgs
+		BlockNrOrHash rpc.BlockNumberOrHash
+		Accounts map[common.Address]account
+	}
+	cr := &CallReq{
+		args,
+		blockNrOrHash,
+		accounts,
```
HaoyangLiu, line 899: `accounts` is a map. As we know, the iterator sequence is random for map in golang. Does this affect the hash result event if the accounts are the same?

PR #112, opened 2021-03-15
[R4R]apply max commit tx time for miner worker

guagualvcha:
### Description
The `miner worker` together with `parlia consensus` has a defect that if the previous validator produces a large block or propagate the block to the next validator later than 3 seconds,  if the next validator still tries to include tx as much as possible, it will eventually produce an empty block. This PR target to solve this issue.

### Rationale
We add a new interface of consensus engine: `Delay()`, which is special designed for the consensus engine that has a fixed blocking time. If the miner is working on commit tx and found that it reach the `Delay()-delayLeftOver` time, it will stop commit tx and seal a block, so that the empty block can be avoid.

### Example
an new option is added:
```
--miner.delayleftover 50000000
```

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -737,6 +737,13 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 	}
 
 	var coalescedLogs []*types.Log
+	var stopTimer *time.Timer
+	delay := w.engine.Delay(w.chain, w.current.header)
+	if delay != nil {
+		stopTimer = time.NewTimer(*delay - w.config.DelayLeftOver)
+		log.Info("Time left for mining work", "left", (*delay - w.config.DelayLeftOver).String())
+		defer stopTimer.Stop()
+	}
 
 	for {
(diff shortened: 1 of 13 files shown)


miner/worker.go
```diff
@@ -737,7 +737,14 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 	}
 
 	var coalescedLogs []*types.Log
-
+	var stopTimer *time.Timer
+	delay := w.engine.Delay(w.chain, w.current.header)
+	if delay != nil {
+		stopTimer = time.NewTimer(*delay - w.config.DelayLeftOver)
```
HaoyangLiu, line 743: Is it possible that `*delay - w.config.DelayLeftOver` is negative?

guagualvcha: Yes, if it is negative, the channel will receive a signal immediately. Just tested it.

HaoyangLiu (approved):

yutianwu (approved):

PR #333, opened 2021-07-29
[R4R]fix: block fetcher efficiency

j75689:
### Description

If a fork block is generated, the fork block is write faster than the original block, which will cause the subsequent block write delay.

### Rationale

Put the blocks that cannot find the parent block into the queue, wait for the parent block to write, and then import it.

### Preflight checks

### DIFF
--- a/eth/fetcher/block_fetcher.go
+++ b/eth/fetcher/block_fetcher.go
@@ -32,10 +32,12 @@ import (
 )
 
 const (
-	lightTimeout  = time.Millisecond       // Time allowance before an announced header is explicitly requested
-	arriveTimeout = 500 * time.Millisecond // Time allowance before an announced block/transaction is explicitly requested
-	gatherSlack   = 100 * time.Millisecond // Interval used to collate almost-expired announces with fetches
-	fetchTimeout  = 5 * time.Second        // Maximum allotted time to return an explicitly requested block/transaction
+	lightTimeout        = time.Millisecond       // Time allowance before an announced header is explicitly requested
+	arriveTimeout       = 500 * time.Millisecond // Time allowance before an announced block/transaction is explicitly requested
+	gatherSlack         = 100 * time.Millisecond // Interval used to collate almost-expired announces with fetches
+	fetchTimeout        = 5 * time.Second        // Maximum allotted time to return an explicitly requested block/transaction
+	reQueueBlockTimeout = 500 * time.Millisecond // Time allowance before blocks are requeued for import
+
 )
 
 const (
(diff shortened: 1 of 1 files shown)


eth/fetcher/block_fetcher.go
```diff
@@ -795,6 +806,7 @@ func (f *BlockFetcher) importBlocks(peer string, block *types.Block) {
 		parent := f.getBlock(block.ParentHash())
 		if parent == nil {
 			log.Debug("Unknown parent of propagated block", "peer", peer, "number", block.Number(), "hash", hash, "parent", block.ParentHash())
```
guagualvcha, line 808: You should sleep for a while before enquene again.

eth/fetcher/block_fetcher.go
```diff
@@ -760,6 +768,7 @@ func (f *BlockFetcher) importHeaders(peer string, header *types.Header) {
 		parent := f.getHeader(header.ParentHash)
 		if parent == nil {
 			log.Debug("Unknown parent of propagated header", "peer", peer, "number", header.Number, "hash", hash, "parent", header.ParentHash)
+			f.requeue <- op
```
guagualvcha, line 771: so do here.

eth/fetcher/block_fetcher.go
```diff
@@ -166,6 +166,8 @@ type BlockFetcher struct {
 	done chan common.Hash
 	quit chan struct{}
 
+	requeue chan *blockOrHeaderInject
```
guagualvcha, line 169: You need initial this requeue.

j75689: fixup by 2ef48de

guagualvcha (approved):

yutianwu (approved):

eth/fetcher/block_fetcher.go
```diff
@@ -415,6 +420,10 @@ func (f *BlockFetcher) loop() {
 				f.rescheduleFetch(fetchTimer)
 			}
 
+		case op := <-f.requeue:
+			// Re-queue blocks that have not been written due to fork block competition
```
guagualvcha, line 424: Please add an info log here.

j75689: added

eth/fetcher/block_fetcher.go
j75689: added

eth/fetcher/block_fetcher.go
j75689: added

yutianwu (approved):

PR #355, opened 2021-08-09
[R4R] the miner module should propose block on a proper fork

guagualvcha:
### Description

### Rationale
When there are multi forks of the same height, the miner module may give up the correct fork in the current implementation.

### Example
For validator, the forks who has the same TD but:
1. Has smaller block timestamp
2. Is not produced by the current valdiator
get more chance to be included in the canonical chain.

### Changes
The recent sign validator will not try to propose a block anymore.

### Preflight checks

### Already reviewed by

...

### Related issues


yutianwu (approved):

consensus/parlia/parlia.go
```diff
@@ -882,6 +882,34 @@ func (p *Parlia) EnoughDistance(chain consensus.ChainReader, header *types.Heade
...
+	if _, authorized := snap.Validators[p.val]; !authorized {
+		return true, errUnauthorizedValidator
+	}
+
+	// If we're amongst the recent signers, wait for the next block
+	number := parent.Number.Uint64() + 1
+	for seen, recent := range snap.Recents {
+		if recent == p.val {
+			// Signer is among recents, only wait if the current block doesn't shift it out
+			if limit := uint64(len(snap.Validators)/2 + 1); number < limit || seen > number-limit {
```
kyrie-yl, line 905: please put the "limit := " expression before the if expression, it can make the 'if' expression more concise.

guagualvcha: `if {statement};  {condition} ` is a widely used grammar, I think it is concise enough for reviewers.

kyrie-yl (approved):

Issue #341, opened 2021-07-31
Pending Block is null after upgrade to Latest Version (1.1.1-beta)


zcrypt0:
#### System information

Geth version: `1.1.1-beta`
OS & Version: Linux
Commit hash : (if `develop`)

#### Expected behaviour

`eth.getBlock('pending')` returns the pending block when `--mine` is specified on the command line.

#### Actual behaviour

`eth.getBlock('pending')` returns null.

#### Backtrace

N/A

### More Details

I upgraded from 1.10.2 (https://github.com/binance-chain/bsc/commit/032970b2deeffe3b7f87703753d7be9a09906758) to 1.1.1-beta and now the node no longer generates pending blocks.

I resynced the node entirely using `--syncmode=snap` with version 1.1.1-beta to see if it would solve the issue. The pending block is still null. 

Snapshot generation is on and the snapshot is completely generated (if applicable).

I downgraded back to 1.10.2 (with the same db synced on 1.1.1-beta) and confirmed that pending blocks are generated, so this is definitely some type of issue with 1.1.1-beta. Note I am specifying the same command line in both versions which includes `--mine`.

guagualvcha:
hi, @zcrypt0 Thanks for reporting this issue. It is introduced in release v1.1.1-beta, we will fix it in v1.1.1 stable release

zcrypt0:
Thanks @guagualvcha 

This is becoming somewhat of a priority for my use case, and I expect others who depend on timely generation of pending blocks. Given the number of transactions on the blockchain right now, I cannot generate pending blocks in a timely manner even on relatively high end hardware. 

I will probably attempt an interim hacky patch myself. I wonder if you might point out relevant sections in the code to help accelerate this, since I am not familiar with the geth codebase.

Alternatively, a dev branch that could be built manually would be helpful as well if you are already making progress on it.

guagualvcha:
> Hi, @guagualvcha
> When v1.1.1 stable will be release?

It will be done within this week.

lovelyrrg51:
Great
So at that time, will only update geth on Node?
@guagualvcha

lovelyrrg51:
And I've another question.
Currently, BSC Node size is fullfilled 100GB every day.
Is this normal?
@guagualvcha

Undead8:
I think that this issue should be reopened because it has been only partially fixed, as mentioned [here](https://github.com/binance-chain/bsc/pull/358#issuecomment-898466311).

PR #358, opened 2021-08-10
[R3R]fix pending block null issue

guagualvcha:
### Description
Fix https://github.com/binance-chain/bsc/issues/341

### Rationale
release v1.1.1-beta will not copy state anymore during proposing block to improve the performance, so that the pending block is null, this PR target to fix it.

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -184,19 +184,21 @@ func (miner *Miner) SetRecommitInterval(interval time.Duration) {
 // Pending returns the currently pending block and associated state.
 func (miner *Miner) Pending() (*types.Block, *state.StateDB) {
 	if miner.worker.isRunning() {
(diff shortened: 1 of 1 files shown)


yutianwu (approved):

j75689 (approved):

Undead8:
@guagualvcha  Unfortunately, this has not fixed issue #341.
The pending block is no longer null as intended. However, now both `eth.getBlock('pending')` and `eth.getBlock('latest')` will return the **latest** block, even if the `--mine` flag is specified.
The pending block should be accessible somehow and it does not seem to be with this release.

PR #376, opened 2021-08-22
[R4R]implement diff sync

guagualvcha:
#### Abstract

The increasing adoption of BSC leads to a more active network. Blocks on BSC start hitting the gasceil daily and we are planning to increase the capacity of BSC further.  On the other hand, the node maintainer had a hard time keeping their node catching up with the chain. We need a light sync mode to lower the hardware requirement for running a bsc fullnode.

Currently bsc has three kinds of sync mode: 1.Snap sync. 2. Fast sync. 3. Full sync.

1 and 2 are used for the initial synchronization, once the client has the entire state and all historical block data, it will switch to full sync automatically. 

It takes several steps to process a block when doing full sync: 
1. Block fetcher or downloader get blocks from other peers.  
2. Verify header and block body. 
3. Execute transactions within EVM, including read data from cache/disk.
4. Calculate the root hash of MPT.
5. Commit MPT to memory db,  persist snapshot to disk.

According to our profile, step 3 occupied 70+% of the block processing time.  

This design wants to propose a diff sync protocol without executing transactions, in exchange, the security of a fullnode will degrade to light client.

#### WorkFlow
diffsync

#### Cache Difflayer and Persist Difflayer
The size of one difflayer is around 100K bytes, it is totally fine to cache 10K diff layers, which means a client can still diff sync if it is down for 8 hours.

We have to persist difflayer to a new database so that a node can still get difflayer even after a long downtime.

#### Fallback to Full Sync && Switch to Diff Sync

When fallback to full sync:
1. Can not receive difflayer in time.
2. No peers have that diff layer.
3. Randomly switch to full sync(every 21 blocks) to ensure security.

When switch diff sync:
1. Periodically detect the network, whether any peer has the latest difflayer.
#### Delay of Diff Sync
The verification of difflayer takes time, we adopt an optimistic strategy to broadcast the difflayer. A node will cache two kinds of difflayers: 1. Trusted difflayer which have been verified; 2. Untrusted difflayer received from other peers.  A node always tries to respond with trusted difflayer first, and only responds with untrusted difflayer if it is missed in cache and database.  The client can disconnect the peer once receive an invalid diff layer. In this way, we dismiss the delay of diff sync. 

#### Security
The diff sync guarantee that:
1. Light client security
2. State consistency. 

It sustains:
validator collusion
Fork the chain with an invalid state 

So do not enable diff sync when it requires high security.

For a validator, it can only use diff sync when it is the inturn validator for the next block. It guarantee that at most one validator is doing diff sync at the same time, the whole network is still secure.

### Preflight checks

### Already reviewed by

...

### Related issues


OhArthits (approved):

kyrie-yl (approved):

cmd/utils/flags.go
```diff
@@ -117,6 +117,15 @@ var (
 		Name:  "directbroadcast",
 		Usage: "Enable directly broadcast mined block to all peers",
 	}
+	DisableSnapProtocolFlag = cli.BoolFlag{
+		Name:  "disablesnapprotocol",
+		Usage: "Disable snap protocol",
+	}
+	DiffSyncFlag = cli.BoolFlag{
+		Name: "diffsync",
+		Usage: "Enable difflayer sync, Please note that enable diffsync will improve the syncing speed, " +
```
yutianwu, line 126: can we change difflayer to `diff layer`

core/state_processor.go
```diff
@@ -49,20 +66,315 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+}
+
+func (p *LightStateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg vm.Config) (*state.StateDB, types.Receipts, []*types.Log, uint64, error) {
+	allowLightProcess := true
+	if posa, ok := p.engine.(consensus.PoSA); ok {
+		allowLightProcess = posa.AllowLightProcess(p.bc, block.Header())
+		log.Error("===debug, allow to light process?", "allow", allowLightProcess)
+	}
+	// random fallback to full process
+	if check := p.randomGenerator.Int63n(fullProcessCheck); allowLightProcess && check != 0 && len(block.Transactions()) != 0 {
```
yutianwu, line 89: do we need to update random seed here

eth/handler.go
```diff
@@ -453,9 +482,16 @@ func (h *handler) BroadcastBlock(block *types.Block, propagate bool) {
 		} else {
 			transfer = peers[:int(math.Sqrt(float64(len(peers))))]
 		}
+		diff := h.chain.GetDiffLayerRLP(block.Hash())
 		for _, peer := range transfer {
+			if len(diff) != 0 && peer.diffExt != nil {
+				// difflayer should send before block
+				log.Error("===debug Broadcast block", "number", block.Number(), "hash", hash)
```
yutianwu, line 489: remove debug log

eth/protocols/diff/handshake.go
```diff
@@ -0,0 +1,81 @@
...
+		return err
+	}
+	if msg.Code != DiffCapMsg {
+		return fmt.Errorf("%w: first msg has code %x (!= %x)", errNoCapMsg, msg.Code, DiffCapMsg)
+	}
+	if msg.Size > maxMessageSize {
+		return fmt.Errorf("%w: %v > %v", errMsgTooLarge, msg.Size, maxMessageSize)
+	}
+	// Decode the handshake and make sure everything matches
+	if err := msg.Decode(cap); err != nil {
```
yutianwu, line 77: do we need to consider the possibility of adding new params in handshake

eth/protocols/diff/protocol.go
```diff
@@ -0,0 +1,120 @@
...
+	return diffLayers, nil
+}
+
+type DiffCapPacket struct {
+	DiffSync bool
+}
+
+type DiffLayersPacket []rlp.RawValue
+
+type FullDiffLayersPacket struct {
```
yutianwu, line 105: do we have an estimate size of `FullDiffLayersPacket`

eth/handler.go
guagualvcha: please ignore this. I add it for testing, and it is not done yet.

eth/protocols/diff/handshake.go
guagualvcha: good idea. Let me think about this further.

core/state_processor.go
guagualvcha: no need. The seed is depends on the start time of nodes, which is quite random enough.

eth/protocols/diff/protocol.go
guagualvcha: Around 100K bytes on main-net for each block.

core/state_processor.go
```diff
@@ -49,20 +66,320 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+				parent := p.bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
+				statedb, err = state.New(parent.Root, p.bc.stateCache, p.bc.snaps)
+				if err != nil {
+					return statedb, nil, nil, 0, err
+				}
+				// Enable prefetching to pull in trie node paths while processing transactions
+				statedb.StartPrefetcher("chain")
+			}
+		}
+	}
```
yutianwu, line 129: do we need to return here or we will always fall back to full process

core/state_processor.go
```diff
@@ -49,20 +66,320 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+	// fallback to full process
+	return p.StateProcessor.Process(block, statedb, cfg)
+}
+
+func (p *LightStateProcessor) LightProcess(diffLayer *types.DiffLayer, block *types.Block, statedb *state.StateDB, cfg vm.Config) (types.Receipts, []*types.Log, uint64, error) {
```
yutianwu, line 134: unused `cfg`

core/state_processor.go
```diff
@@ -49,20 +66,320 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+		fullDiffCode[c.Hash] = c.Code
+	}
+
+	for des := range snapDestructs {
+		statedb.Trie().TryDelete(des[:])
+	}
+	threads := 1
+	if len(snapAccounts)/runtime.NumCPU() > minNumberOfAccountPerTask {
+		threads = runtime.NumCPU()
+	}
```
yutianwu, line 155: `threads` will have only two options: 1 or `runtime.NumCPU()`

core/state_processor.go
```diff
@@ -49,20 +66,320 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+				var previousAccount state.Account
+				stateMux.Lock()
+				enc, err := statedb.Trie().TryGet(diffAccount[:])
+				stateMux.Unlock()
+				if err != nil {
+					errChan <- err
+					return
+				}
+				if len(enc) != 0 {
+					if err := rlp.DecodeBytes(enc, &previousAccount); err != nil {
```
yutianwu, line 201: why not use `snapshot.FullAccount` here

cmd/utils/flags.go
guagualvcha: change `difflayer sync` into `diff sync`

core/state_processor.go
guagualvcha: Already fallback to full process.

core/state_processor.go
guagualvcha: removed

core/state_processor.go
guagualvcha: yes, I just use a simple strategy here.

core/state_processor.go
guagualvcha: This is state.Account

yutianwu (approved):

core/state/statedb.go
```diff
@@ -1080,12 +1250,26 @@ func (s *StateDB) Commit(deleteEmptyObjects bool) (common.Hash, error) {
 									log.Crit("Failed to commit dirty codes", "error", err)
 								}
 							}
+							wg.Done()
```
j75689, line 1253: I prefer to use `defer wg.Done()`

core/state_processor.go
```diff
@@ -49,20 +68,323 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+		}
+		if diffLayer != nil {
+			if err := diffLayer.Receipts.DeriveFields(p.bc.chainConfig, block.Hash(), block.NumberU64(), block.Transactions()); err != nil {
+				log.Error("Failed to derive block receipts fields", "hash", block.Hash(), "number", block.NumberU64(), "err", err)
+				// fallback to full process
+				return p.StateProcessor.Process(block, statedb, cfg)
+			}
+
+			receipts, logs, gasUsed, err := p.LightProcess(diffLayer, block, statedb)
+			if err == nil {
```
j75689, line 116: It is recommended to use the same code style
```go
if err != nil {
	log.Error("do light process err at block", "num", block.NumberU64(), "err", err)
	p.bc.removeDiffLayers(diffLayer.DiffHash)
	// prepare new statedb
	statedb.StopPrefetcher()
	parent := p.bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
	statedb, err = state.New(parent.Root, p.bc.stateCache, p.bc.snaps)
	if err != nil {
		return statedb, nil, nil, 0, err
	}
	// Enable prefetching to pull in trie node paths while processing transactions
	statedb.StartPrefetcher("chain")
}
log.Info("do light process success at block", "num", block.NumberU64())
return statedb, receipts, logs, gasUsed, nil
```

core/state_processor.go
```diff
@@ -49,20 +68,323 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+					accountTrie, err := statedb.Database().OpenStorageTrie(addrHash, previousAccount.Root)
+					if err != nil {
+						errChan <- err
+						return
+					}
+					snapMux.RLock()
+					storageChange, exist := snapStorage[diffAccount]
+					snapMux.RUnlock()
+
+					if exist {
```
j75689, line 272: suggestion
```go
if !exist {
	errChan <- errors.New("missing storage change in difflayer")
	return
}

for k, v := range storageChange {
	if len(v) != 0 {
		accountTrie.TryUpdate([]byte(k), v)
	} else {
		accountTrie.TryDelete([]byte(k))
	}
}
```

eth/downloader/downloader.go
```diff
@@ -220,8 +220,43 @@ type BlockChain interface {
...
+	GetDiffPeer(string) IDiffPeer
+}
+
+func EnableDiffFetchOp(peers IPeerSet) DownloadOption {
+	return func(dl *Downloader) *Downloader {
+		var hook = func(headers []*types.Header, args ...interface{}) {
+			if len(args) < 2 {
+				return
+			}
+			if mode, ok := args[0].(SyncMode); ok {
```
j75689, line 239: suggestion
```go
peerID, ok := args[1].(string)
if !ok {
	return
}
mode, ok := args[0].(SyncMode)
if !ok {
	return
}
if ep := peers.GetDiffPeer(peerID); mode == FullSync && ep != nil {
	hashes := make([]common.Hash, 0, len(headers))
	for _, header := range headers {
		hashes = append(hashes, header.Hash())
	}
	ep.RequestDiffLayers(hashes)
}
```

eth/downloader/downloader.go
```diff
@@ -220,8 +220,43 @@ type BlockChain interface {
 	Snapshots() *snapshot.Tree
 }
 
+type DownloadOption func(downloader *Downloader) *Downloader
```
j75689, line 223: returning `*Downloader` is unnecessary

eth/handler_diff.go
```diff
@@ -0,0 +1,89 @@
...
+	if err := peer.Handshake(h.diffSync); err != nil {
+		return err
+	}
+	defer h.chain.RemoveDiffPeer(peer.ID())
+	return (*handler)(h).runDiffExtension(peer, hand)
+}
+
+// PeerInfo retrieves all known `diff` information about a peer.
+func (h *diffHandler) PeerInfo(id enode.ID) interface{} {
+	if p := h.peers.peer(id.String()); p != nil {
```
j75689, line 44: can be combined into one line
```go
if p := h.peers.peer(id.String()); p != nil && p.diffExt != nil {
		return p.diffExt.info()
	}
```

eth/peerset.go
```diff
@@ -131,9 +172,50 @@ func (ps *peerSet) waitSnapExtension(peer *eth.Peer) (*snap.Peer, error) {
...
+	if diff, ok := ps.diffPend[id]; ok {
+		delete(ps.diffPend, id)
+
+		ps.lock.Unlock()
+		return diff, nil
+	}
+	// Otherwise wait for `diff` to connect concurrently
+	wait := make(chan *diff.Peer)
+	ps.diffWait[id] = wait
+	ps.lock.Unlock()
```
j75689, line 204: move this line to :184 `defer ps.lock.Unlock`

eth/protocols/diff/handler.go
```diff
@@ -0,0 +1,181 @@
...
+		return backend.Handle(peer, res)
+	case msg.Code == FullDiffLayerMsg:
+		// A batch of trie nodes arrived to one of our previous requests
+		res := new(FullDiffLayersPacket)
+		if err := msg.Decode(res); err != nil {
+			return fmt.Errorf("%w: message %v: %v", errDecode, msg, err)
+		}
+		if fulfilled := requestTracker.Fulfil(peer.id, peer.version, FullDiffLayerMsg, res.RequestId); fulfilled {
+			return backend.Handle(peer, res)
+		} else {
```
j75689, line 146: this `else` is unnecessary

node/node.go
```diff
@@ -611,6 +628,30 @@ func (n *Node) OpenDatabaseWithFreezer(name string, cache, handles int, freezer,
...
+	defer n.lock.Unlock()
+	if n.state == closedState {
+		return nil, ErrNodeStopped
+	}
+
+	var db *leveldb.Database
+	var err error
+	if n.config.DataDir == "" {
+		panic("datadir is missing")
+	} else {
```
j75689, line 642: this `else` is unnecessary

cmd/faucet/faucet.go
```diff
@@ -139,7 +139,7 @@ func main() {
 		log.Crit("Length of bep2eContracts, bep2eSymbols, bep2eAmounts mismatch")
 	}
 
-	bep2eInfos := make(map[string]bep2eInfo, 0)
+	bep2eInfos := make(map[string]bep2eInfo)
```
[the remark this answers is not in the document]
realuncle: Fixed.

core/blockchain.go
```diff
@@ -142,6 +156,8 @@ var defaultCacheConfig = &CacheConfig{
 	SnapshotWait:   true,
 }
 
+type BlockChainOption func(*BlockChain) *BlockChain
```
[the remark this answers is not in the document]
realuncle: It is true there is no need to return in this case. But in some other cases, we may wrapper the Object which is an interface, and return the wrapper. It impact little to performance, just personal code style. If you insist, I can prune the return objects.

core/blockchain.go
```diff
@@ -375,6 +419,10 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 		}
 		bc.snaps, _ = snapshot.New(bc.db, bc.stateCache.TrieDB(), bc.cacheConfig.SnapshotLimit, int(bc.cacheConfig.TriesInMemory), head.Root(), !bc.cacheConfig.SnapshotWait, true, recover)
 	}
+	// do options before start any routine
+	for _, option := range options {
+		bc = option(bc)
```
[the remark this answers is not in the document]
realuncle: check above

eth/downloader/downloader.go
realuncle: check above

eth/handler_diff.go
realuncle: fixed.

eth/peerset.go
realuncle: I just follow what `waitSnapExtension` did, it has been proven quite safe. So I think we don't have motivation to modify this.

eth/protocols/diff/handler.go
realuncle: fixed.

node/node.go
realuncle: fixed

consensus/parlia/parlia.go
```diff
@@ -882,6 +886,18 @@ func (p *Parlia) EnoughDistance(chain consensus.ChainReader, header *types.Heade
...
+func (p *Parlia) AllowLightProcess(chain consensus.ChainReader, currentHeader *types.Header) bool {
+	snap, err := p.snapshot(chain, currentHeader.Number.Uint64()-1, currentHeader.ParentHash, nil)
+	if err != nil {
+		return true
+	}
+
+	idx := snap.indexOfVal(p.val)
+	// validator is not allowed to diff sync
+	return idx < 0
+
```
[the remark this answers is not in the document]
realuncle: fixed.

consensus/parlia/parlia.go
```diff
@@ -799,6 +799,10 @@ func (p *Parlia) Delay(chain consensus.ChainReader, header *types.Header) *time.
 		return nil
 	}
 	delay := p.delayForRamanujanFork(snap, header)
+	// The blocking time should be no more than half of period
+	if delay > time.Duration(p.config.Period)*time.Second/2 {
```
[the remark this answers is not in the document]
realuncle: fixed

core/blockchain.go
```diff
@@ -2292,6 +2401,280 @@ func (bc *BlockChain) update() {
...
+func (bc *BlockChain) GetUnTrustedDiffLayer(blockHash common.Hash, pid string) *types.DiffLayer {
+	bc.diffMux.RLock()
+	defer bc.diffMux.RUnlock()
+	if diffs, exist := bc.blockHashToDiffLayers[blockHash]; exist && len(diffs) != 0 {
+		if len(diffs) == 1 {
+			// return the only one diff layer
+			for _, diff := range diffs {
+				return diff
+			}
+		} else {
```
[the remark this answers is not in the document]
realuncle: fixed

core/blockchain.go
```diff
@@ -2292,6 +2401,280 @@ func (bc *BlockChain) update() {
...
+				}
+				delete(bc.diffHashToBlockHash, invalidDiffHash)
+			}
+		}
+		delete(bc.diffPeersToDiffHashes, pid)
+	}
+}
+
+func (bc *BlockChain) untrustedDiffLayerPruneLoop() {
+	recheck := time.Tick(diffLayerPruneRecheckInterval)
```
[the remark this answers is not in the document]
realuncle: Make sense. Fixed.

core/blockchain.go
```diff
@@ -2292,6 +2401,280 @@ func (bc *BlockChain) update() {
...
+		if number < currentHeight-maxDiffForkDist {
+			affectedHashes := bc.diffNumToBlockHashes[number]
+			if affectedHashes != nil {
+				for affectedHash := range affectedHashes {
+					staleBlockHashes[affectedHash] = struct{}{}
+				}
+				delete(bc.diffNumToBlockHashes, number)
+			}
+		} else {
+			break
```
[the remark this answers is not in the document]
realuncle: good suggestion. Fixed.

core/blockchain.go
```diff
@@ -2292,6 +2401,280 @@ func (bc *BlockChain) update() {
...
+	}
+	if diffLayer.Number < currentHeight && currentHeight-diffLayer.Number > maxDiffForkDist {
+		log.Error("diff layers too old from current", "pid", pid)
+		return nil
+	}
+
+	bc.diffMux.Lock()
+	defer bc.diffMux.Unlock()
+
+	if !fulfilled {
```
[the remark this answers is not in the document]
guagualvcha: fixed

core/blockchain.go
```diff
@@ -2541,3 +2924,19 @@ func (bc *BlockChain) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscript
 func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscription {
 	return bc.scope.Track(bc.blockProcFeed.Subscribe(ch))
 }
+
+// Options
+func EnableLightProcessor(bc *BlockChain) *BlockChain {
```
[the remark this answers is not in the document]
guagualvcha: check above

core/state/statedb.go
realuncle: good idea. Fixed.

core/state_processor.go
realuncle: `p.StateProcessor.Process(block, statedb, cfg)` is also needed if 
`if check := p.randomGenerator.Int63n(fullProcessCheck); allowLightProcess && check != 0 && len(block.Transactions()) != 0 {` can not meet.

realuncle: I agree with you that handle `err!=nil` first, but this is a special case.

core/state_processor.go
realuncle: fixed

eth/downloader/downloader.go
realuncle: godd suggestion. Fixed.

core/state_processor.go
j75689: ok, but maybe it can be written like:
```go
if err == nil {
	log.Info("do light process success at block", "num", block.NumberU64())
	return statedb, receipts, logs, gasUsed, nil
} 
log.Error("do light process err at block", "num", block.NumberU64(), "err", err)
p.bc.removeDiffLayers(diffLayer.DiffHash)
// prepare new statedb
statedb.StopPrefetcher()
parent := p.bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
statedb, err = state.New(parent.Root, p.bc.stateCache, p.bc.snaps)
if err != nil {
	return statedb, nil, nil, 0, err
}
// Enable prefetching to pull in trie node paths while processing transactions
statedb.StartPrefetcher("chain")
```
the `else` is unnecessary

core/state_processor.go
```diff
@@ -49,20 +68,322 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
+
+	for des := range snapDestructs {
+		statedb.Trie().TryDelete(des[:])
+	}
+	threads := len(snapAccounts) / minNumberOfAccountPerTask
+	if threads > runtime.NumCPU() {
+		threads = runtime.NumCPU()
+	} else if threads == 0 {
+		threads = 1
+	}
```
j75689, line 160: I saw that this logic had been written in two places, and the only parameter is the difference.
If it is just a strategy for determining the number of threads, we can make a unique function.
It will be easier to maintain.

realuncle: fixed

PR #384, opened 2021-08-26
[WIP]Optimized db storage

Mercybudda:
### Description

This is the optimization for database storage

### Rationale
1. Block data is written to leveldb for two times. It occupied too much disk bandwidth. And as block data is sequential and less read operations, compaction to block data also takes disk bandwidth/ cpu/ memory.
2. State data in KV Store needs to do compaction in order to get better read performance. If block data and state data in the same KV Store, compaction will take more time(more data/ cpu/ memory/ disk bandwidth). Indeed we just need current world state(or latest world state) to execute. In future we needs other solution to get better performance.

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -738,12 +738,16 @@ func (bc *BlockChain) ResetWithGenesisBlock(genesis *types.Block) error {
 	defer bc.chainmu.Unlock()
 
 	// Prepare the genesis block and reinitialise the chain
-	batch := bc.db.NewBatch()
-	rawdb.WriteTd(batch, genesis.Hash(), genesis.NumberU64(), genesis.Difficulty())
-	rawdb.WriteBlock(batch, genesis)
-	if err := batch.Write(); err != nil {
-		log.Crit("Failed to write genesis block", "err", err)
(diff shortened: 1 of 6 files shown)


core/blockchain.go
```diff
@@ -1342,7 +1344,8 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 				}
 			}
 			// Write all the data out into the database
-			rawdb.WriteBody(batch, block.Hash(), block.NumberU64(), block.Body())
+			//rawdb.WriteBody(batch, block.Hash(), block.NumberU64(), block.Body())
+			rawdb.WriteAncientBlock(bc.db, block, receiptChain[i], bc.GetTd(block.Hash(), block.NumberU64()))
 			rawdb.WriteReceipts(batch, block.Hash(), block.NumberU64(), receiptChain[i])
```
RenRick, line 1349: L1349: Receipts don't need to be persiste to rawdb.

Issue #477, opened 2021-10-24
issue of prefetcher


guagualvcha:
`github.com/ethereum/go-ethereum/core/state/trie_prefetcher.go:86 +0x96` leak

nil point at line 106:

![image](https://user-images.githubusercontent.com/7310198/138591369-5e9e4617-2f5c-4ddf-81ec-576ea8a5b5da.png)

guagualvcha:
fixed

PR #491, opened 2021-10-29
[R4R]fix prefetcher related bugs

KeefeL:
### Description

1.Fix panic in trie_prefetcher close
2.Fix trie_prefetcher abortLoop goroutine leak 

### Rationale

1.In ValidateState, there are three goroutines running concurrently, if the previous one return error, the upper caller function InsertChain would return and call it's defer function to close prefetcher, while the IntermediateRoot goroutine still run and when it done call its defer function to close prefetcher, this time the prefetcher is nil, so it will trigger a panic. So we should wait all the goroutines done in ValidateState to avoid this panic.
2.In InsertChain, if we enable diffsync, and insert two more blocks here, then all the blocks except the last one wound't do call StopPrefetcher, so these blocks' abortLoop goroutine will leak. So we should do the StopPrefetcher in IntermediateRoot to avoid this leak. 

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

#477

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -144,13 +144,15 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 			validateRes <- tmpFunc()
 		}()
 	}
+
+	var err error
 	for i := 0; i < len(validateFuns); i++ {
 		r := <-validateRes
-		if r != nil {
-			return r
+		if r != nil && err == nil {
+			err = r
 		}
 	}
-	return nil
+	return err
 }
 
 // CalcGasLimit computes the gas limit of the next block after parent. It aims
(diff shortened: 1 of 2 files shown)


core/block_validator.go
```diff
@@ -144,13 +144,15 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 			validateRes <- tmpFunc()
 		}()
 	}
+
+	var err error = nil
```
guagualvcha, line 148: `var err error` is enough

core/state/statedb.go
```diff
@@ -958,6 +958,7 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 // goes into transaction receipts.
 func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
 	if s.lightProcessed {
+		s.StopPrefetcher()
```
guagualvcha, line 961: can we just move above code to the top of this func?
```
	prefetcher := s.prefetcher
	if s.prefetcher != nil {
		defer func() {
			s.prefetcher.close()
			s.prefetcher = nil
		}()
	}
```

core/block_validator.go
KeefeL: done

yutianwu (approved):

guagualvcha (approved):

PR #493, opened 2021-10-29
Accelerate get diff accounts with scope

KeefeL:
### Description

Add needToReplay condition to accelerate eth_getDiffAccountsWithScope

### Rationale

Replay the whole block spend too much time, actually we need't do that, so add an quick way to decide whether to replay the block.

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1101,6 +1101,66 @@ func (s *PublicBlockChainAPI) GetDiffAccounts(ctx context.Context, blockNr rpc.B
 	return s.b.Chain().GetDiffAccounts(header.Hash())
 }
 
+func (s *PublicBlockChainAPI) needToReplay(ctx context.Context, block *types.Block, accounts []common.Address) (bool, error) {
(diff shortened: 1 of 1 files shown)


internal/ethapi/api.go
```diff
@@ -1111,6 +1171,21 @@ func (s *PublicBlockChainAPI) GetDiffAccountsWithScope(ctx context.Context, bloc
...
 		return nil, fmt.Errorf("block not found for block number (%d): %v", blockNr, err)
 	}
+
+	result := &types.DiffAccountsInBlock{
+		Number:       uint64(blockNr),
+		BlockHash:    block.Hash(),
+		Transactions: make([]types.DiffAccountsInTx, 0),
+	}
+
+	if needReplay, err := s.needToReplay(ctx, block, accounts); !needReplay {
```
yutianwu, line 1181: can we test err first

yutianwu (approved):

guagualvcha (approved):

PR #536, opened 2021-11-10
get diff accounts by replaying block when diff layer not found

KeefeL:
Signed-off-by: Keefe-Liu <bianze.kernel@gmail.com>

### Description

Get block's diff accounts by replaying block when diff layer not found.

### Rationale

The diff layer just can cache for a few days, when there is no diff layer, the eth_getDiffAccounts API couldn't work, in this case we should replay the block and get the diff accounts, it may slow but i can work right.

### Example

### Changes

Notable changes: 
* ...

### Preflight checks

### Already reviewed by

...

### Related issues

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1493,3 +1493,11 @@ func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addre
 	}
 	return s.accessList.Contains(addr, slot)
 }
+
+func (s *StateDB) GetDirtyAccounts() []common.Address {
+	accounts := make([]common.Address, 0, len(s.stateObjectsDirty))
+	for account := range s.stateObjectsDirty {
+		accounts = append(accounts, account)
+	}
+	return accounts
+}
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1098,7 +1098,21 @@ func (s *PublicBlockChainAPI) GetDiffAccounts(ctx context.Context, blockNr rpc.B
 		return nil, fmt.Errorf("block not found for block number (%d): %v", blockNr, err)
 	}
 
-	return s.b.Chain().GetDiffAccounts(header.Hash())
+	accounts, err := s.b.Chain().GetDiffAccounts(header.Hash())
+	if err == nil || !errors.Is(err, core.ErrDiffLayerNotFound) {
+		return accounts, err
+	}
+
+	// Replay the block when diff layer not found, it is very slow.
+	block, err := s.b.BlockByNumber(ctx, blockNr)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


core/state/statedb.go
```diff
@@ -1493,3 +1493,11 @@ func (s *StateDB) SlotInAccessList(addr common.Address, slot common.Hash) (addre
 	}
 	return s.accessList.Contains(addr, slot)
 }
+
+func (s *StateDB) GetDirtyAccounts() []common.Address {
+	accounts := make([]common.Address, 0)
```
guagualvcha, line 1498: The capacity should be `len(s.stateObjectsDirty)`

guagualvcha:
Try fixing the CI as well. Seems integration test stuck

internal/ethapi/api.go
```diff
@@ -1259,7 +1257,32 @@ func (s *PublicBlockChainAPI) GetDiffAccountsWithScope(ctx context.Context, bloc
...
+	if s.b.Chain() == nil {
+		return nil, fmt.Errorf("blockchain not support get diff accounts")
+	}
+
+	block, err := s.b.BlockByNumber(ctx, blockNr)
+	if err != nil {
+		return nil, fmt.Errorf("block not found for block number (%d): %v", blockNr, err)
+	}
+
+	if needReplay, err := s.needToReplay(ctx, block, accounts); err != nil {
```
j75689, line 1274: I prefer this style 😂
```go
needReplay, err := s.needToReplay(ctx, block, accounts)
if err != nil {
	return nil, err
}
if !needReplay {
	return &types.DiffAccountsInBlock{
		Number:       uint64(blockNr),
		BlockHash:    block.Hash(),
		Transactions: make([]types.DiffAccountsInTx, 0),
	}, nil
}
```

internal/ethapi/api.go
```diff
@@ -1098,7 +1098,21 @@ func (s *PublicBlockChainAPI) GetDiffAccounts(ctx context.Context, blockNr rpc.B
 		return nil, fmt.Errorf("block not found for block number (%d): %v", blockNr, err)
 	}
 
-	return s.b.Chain().GetDiffAccounts(header.Hash())
+	accounts, err := s.b.Chain().GetDiffAccounts(header.Hash())
+	if err == nil || err.Error() != "no diff layer found" {
```
j75689, line 1102: Can we define this error in `core/error.go`, and compare it with `errors.Is(err, core.ErrDiffLayerNofFound)` instead of string comparison?

j75689:
> Try fixing the CI as well. Seems integration test stuck

Sometimes public machines on GitHub are busy, which will cause timeout errors...
just need to re-run it

guagualvcha (approved):

j75689 (approved):

PR #543, opened 2021-11-14
[R4R]offline block prune

Mercybudda:
### Description
This is for the off-line block prune, in bsc storage, there's db wrapper of ancientDb(freezer) and kvDb(levelDB), the space was mainly occupied by ancientDb, the purpose of this is to prune and remain the specific number of old blocks in ancientDb by specifying the amount of blocks you'd like to remain after pruning in the CLI flag. Then the pruner will prune and only remain this specified amount backward, this first block is not necessarily the genesis block because we may prune many rounds. Actually we can prune and remain any amount of blocks as long as the amount specified did exceed the number of blocks inside ancientDb
###Example:
Assume all blocks in ancientDb are [0,1,... 1000], if pruning amount of 500 blocks, then [0,...499] will be pruned, the data in ancientDb will be [500, 1000]. The next round for pruning will start with BlockNumber 500.
### Brief Coding workflow is as follows
1. Create a new ancientDb without a kvdb wrapper, it will create a new folder ./node/geth/chaindata/ancient_backup for initializing new ancientDb.
2. Back-up all the blocks at and after oldOffSet + frozen - BlockPruneAmountLeft from old ancientDb into new ancientBackup. In this process, pruner will read one by one for block, receipts, difficulty from old ancientDb and write to new backup ancientDb. To be noted that there’ll be a new offset of new ancientDb, this offset will be recorded into kvdb.
3. Delete directly for the old folder./node/geth /chaindata/ancient/, i.e. the old ancientDb.
4. Rename the new folder of ./node/geth/chaindata/ancient_backup into ./node/geth/chaindata/ancient, then assemble with the same kvdb of the original DB wrapper in which old ancientDb resides.
5. Finally the ancientDb in DB wrapper is replaced successfully, all the block data we want to reserve were in this new ancientDb.

### Self Test
1. Unit test: Initialized a blockchain of 500 blocks, testing the pruning of 328 amount of blocks remaining. passed
2. Locally geth test, start a full node: `./geth --config ./config.toml --datadir ./node --cache 12000 --rpc.allow-unprotected-txs --txlookuplimit 0 --syncmode full --usb`, after a while until there's enough old block for pruning in ancientDb. 
3. 1st round of prune test: Start the prune process:  `./geth snapshot prune-block --datadir ./node --datadir.ancient  ./chaindata/ancient --block-amount-reserved 10000`. This will prune and reserve 10000 blocks.
Passed and get a new ancientDb with amount of 10000 blocks
4. 2nd round of prune test, start the geth again, repeat the step 2 & 3.
5. Introduced the flag `inspect-reserved-oldest-blocks`,  can be used for checking information of ancientStore like offset and reserved items amount after pruning done. 
`./geth db inspect-reserved-oldest-blocks --datadir ./node`
+--------------------------------+--------------------------------+--------+
|            DATABASE            |            CATEGORY            | ITEMS  |
+--------------------------------+--------------------------------+--------+
| Offset/StartBlockNumber        | Offset/StartBlockNumber after  | 302672 |
|                                                    | BlockPrune                                   |               |
| Amount of remained items in  | remaining items after                   |  10000  |
| AncientStore                             | BlockPrune                                   |               |
+--------------------------------+--------------------------------+--------+
|                                  ANCIENTSTORE INFORMATION AFTER        |
|                                        OFFLINE BLOCKPRUNE                            |
+--------------------------------+--------------------------------+--------+
### Running command
e.g.
`./geth snapshot prune-block --datadir ./node --datadir.ancient ./chaindata/ancient  --block-amount-reserved 9000 --triesInMemory 32 --check-snapshot-with-mpt`
--block-amount-reserved: Amount of blocks want to reserve.
--triesInMemory(optional): The layer of tries trees that keep in memory.
--check-snapshot-with-mpt(optinal): Indicate if enable/disable checking for the match between MPT and snapshot.

Check the info of ancientDB
`./geth db inspect-reserved-oldest-blocks --datadir ./node`
### Result
1. The offset of ancientDb changed as expected.
2. Geth client restarted successfully.
3. new block data freeze() moving from kvDb to ancientDb as expected.
### Preflight checks

### DIFF
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -62,6 +62,7 @@ Remove blockchain and state databases`,
 			dbPutCmd,
 			dbGetSlotsCmd,
 			dbDumpFreezerIndex,
+			blockPruneInspectCmd,
 		},
 	}
 	dbInspectCmd = cli.Command{
@@ -195,6 +196,16 @@ WARNING: This is a low-level operation which may cause database corruption!`,
 		},
 		Description: "This command displays information about the freezer index.",
 	}
+	blockPruneInspectCmd = cli.Command{
+		Action: utils.MigrateFlags(blockPruneInspect),
+		Name:   "inspect-reserved-oldest-blocks",
+		Flags: []cli.Flag{
+			utils.DataDirFlag,
+		},
+		Usage: "Inspect the ancientStore information after block-prune",
+		Description: `This commands will read current offset from kvdb, which is the current offset and starting BlockNumber
+		of ancientStore, will also displays the reserved number of blocks in ancientStore `,
+	}
 )
 
 func removeDB(ctx *cli.Context) error {
@@ -282,10 +293,19 @@ func inspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
+}
+
+func blockPruneInspect(ctx *cli.Context) error {
+	stack, _ := makeConfigNode(ctx)
+	defer stack.Close()
+
+	db := utils.MakeChainDatabase(ctx, stack, true, true)
+	defer db.Close()
+	return rawdb.InspectBlockPrune(db)
 }
 
 func showLeveldbStats(db ethdb.Stater) {
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -282,7 +282,7 @@ func inspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, true)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -282,7 +282,7 @@ func inspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
@@ -305,7 +305,7 @@ func dbStats(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	showLeveldbStats(db)
@@ -316,7 +316,7 @@ func dbCompact(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	log.Info("Stats before compaction")
@@ -340,7 +340,7 @@ func dbGet(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -365,7 +365,7 @@ func dbDelete(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -392,7 +392,7 @@ func dbPut(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	var (
@@ -426,7 +426,7 @@ func dbDumpTrie(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 	var (
 		root  []byte
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -204,7 +204,7 @@ WARNING: This is a low-level operation which may cause database corruption!`,
 		},
 		Usage: "Inspect the ancientStore information",
 		Description: `This commands will read current offset from kvdb, which is the current offset and starting BlockNumber
-		of ancientStore, will also displays the reserved number of blocks in ancientStore `,
+of ancientStore, will also displays the reserved number of blocks in ancientStore `,
 	}
 )
 
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -293,7 +293,7 @@ func inspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
@@ -303,7 +303,7 @@ func ancientInspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, true, false)
 	defer db.Close()
 	return rawdb.AncientInspect(db)
 }
@@ -325,7 +325,7 @@ func dbStats(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
 	defer db.Close()
 
 	showLeveldbStats(db)
@@ -336,7 +336,7 @@ func dbCompact(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
 	defer db.Close()
 
 	log.Info("Stats before compaction")
@@ -360,7 +360,7 @@ func dbGet(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -385,7 +385,7 @@ func dbDelete(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -412,7 +412,7 @@ func dbPut(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
 	defer db.Close()
 
 	var (
@@ -446,7 +446,7 @@ func dbDumpTrie(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
 	defer db.Close()
 	var (
 		root  []byte
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -293,7 +293,7 @@ func inspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
@@ -303,7 +303,7 @@ func ancientInspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, true, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, true)
 	defer db.Close()
 	return rawdb.AncientInspect(db)
 }
@@ -325,7 +325,7 @@ func dbStats(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	showLeveldbStats(db)
@@ -336,7 +336,7 @@ func dbCompact(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	log.Info("Stats before compaction")
@@ -360,7 +360,7 @@ func dbGet(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -385,7 +385,7 @@ func dbDelete(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	key, err := hexutil.Decode(ctx.Args().Get(0))
@@ -412,7 +412,7 @@ func dbPut(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, false, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
 	defer db.Close()
 
 	var (
@@ -446,7 +446,7 @@ func dbDumpTrie(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true, false, false)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 	var (
 		root  []byte
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -62,7 +62,7 @@ Remove blockchain and state databases`,
 			dbPutCmd,
 			dbGetSlotsCmd,
 			dbDumpFreezerIndex,
-			blockPruneInspectCmd,
+			ancientInspectCmd,
 		},
 	}
 	dbInspectCmd = cli.Command{
@@ -196,13 +196,13 @@ WARNING: This is a low-level operation which may cause database corruption!`,
 		},
 		Description: "This command displays information about the freezer index.",
 	}
-	blockPruneInspectCmd = cli.Command{
-		Action: utils.MigrateFlags(blockPruneInspect),
+	ancientInspectCmd = cli.Command{
+		Action: utils.MigrateFlags(ancientInspect),
 		Name:   "inspect-reserved-oldest-blocks",
 		Flags: []cli.Flag{
 			utils.DataDirFlag,
 		},
-		Usage: "Inspect the ancientStore information after block-prune",
+		Usage: "Inspect the ancientStore information",
 		Description: `This commands will read current offset from kvdb, which is the current offset and starting BlockNumber
 		of ancientStore, will also displays the reserved number of blocks in ancientStore `,
 	}
@@ -299,13 +299,13 @@ func inspect(ctx *cli.Context) error {
 	return rawdb.InspectDatabase(db, prefix, start)
 }
 
-func blockPruneInspect(ctx *cli.Context) error {
+func ancientInspect(ctx *cli.Context) error {
 	stack, _ := makeConfigNode(ctx)
 	defer stack.Close()
 
 	db := utils.MakeChainDatabase(ctx, stack, true, true)
 	defer db.Close()
-	return rawdb.InspectBlockPrune(db)
+	return rawdb.AncientInspect(db)
 }
 
 func showLeveldbStats(db ethdb.Stater) {
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -165,7 +165,7 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
-		utils.BlockPruneAmountLeft,
+		utils.BlockAmountReserved,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -165,6 +165,7 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
+		utils.BlockPruneQuantity,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -165,8 +165,6 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
-		utils.AncientBackUpFlag,
-		utils.GenesisFlag,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -165,7 +165,7 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
-		utils.BlockPruneQuantity,
+		utils.BlockPruneAmountLeft,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -165,6 +165,8 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
+		utils.AncientBackUpFlag,
+		utils.GenesisFlag,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -166,6 +166,7 @@ var (
 		configFileFlag,
 		utils.CatalystFlag,
 		utils.BlockAmountReserved,
+		utils.CheckSnapshotWithMPT,
 	}
 
 	rpcFlags = []cli.Flag{
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -64,6 +64,13 @@ var (
 )
 
 func TestOfflineBlockPrune(t *testing.T) {
+	//Corner case for 0 remain in ancinetStore.
+	testOfflineBlockPruneWithAmountLeft(t, 0)
+	//General case.
+	testOfflineBlockPruneWithAmountLeft(t, 100)
+}
+
+func testOfflineBlockPruneWithAmountLeft(t *testing.T, amountLeft uint64) {
 	datadir, err := ioutil.TempDir("", "")
 	if err != nil {
 		t.Fatalf("Failed to create temporary datadir: %v", err)
@@ -74,12 +81,12 @@ func TestOfflineBlockPrune(t *testing.T) {
 	oldAncientPath := filepath.Join(chaindbPath, "ancient")
 	newAncientPath := filepath.Join(chaindbPath, "ancient_back")
 
-	db, blocks, blockList, receiptsList, externTdList, startBlockNumber, _ := BlockchainCreator(t, chaindbPath, oldAncientPath)
+	db, blocks, blockList, receiptsList, externTdList, startBlockNumber, _ := BlockchainCreator(t, chaindbPath, oldAncientPath, amountLeft)
 	node, _ := startEthService(t, gspec, blocks, chaindbPath)
 	defer node.Close()
 
-	//Initialize a block pruner for pruning amount of 328 blocks
-	testBlockPruner, err := pruner.NewBlockPruner(db, node, oldAncientPath, newAncientPath, 328)
+	//Initialize a block pruner for pruning, only remain amountLeft blocks backward.
+	testBlockPruner, err := pruner.NewBlockPruner(db, node, oldAncientPath, newAncientPath, amountLeft)
 	if err != nil {
 		t.Fatalf("failed to make new blockpruner: %v", err)
 	}
@@ -94,7 +101,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 	defer dbBack.Close()
 
 	//check against if the backup data matched original one
-	for blockNumber := startBlockNumber; blockNumber < startBlockNumber+328; blockNumber++ {
+	for blockNumber := startBlockNumber; blockNumber < startBlockNumber+amountLeft; blockNumber++ {
 		blockHash := rawdb.ReadCanonicalHash(dbBack, blockNumber)
 		block := rawdb.ReadBlock(dbBack, blockHash, blockNumber)
 		if reflect.DeepEqual(block, blockList[blockNumber-startBlockNumber]) {
@@ -128,7 +135,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 	}
 }
 
-func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Database, []*types.Block, []*types.Block, []types.Receipts, []*big.Int, uint64, *core.BlockChain) {
+func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string, blockRemain uint64) (ethdb.Database, []*types.Block, []*types.Block, []types.Receipts, []*big.Int, uint64, *core.BlockChain) {
 	//create a database with ancient freezer
 	db, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, AncientPath, "", false, false)
 	if err != nil {
@@ -168,13 +175,13 @@ func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Dat
 	if err != nil || frozen == 0 {
 		t.Fatalf("Failed to import canonical chain start: %v", err)
 	}
-	if frozen < 328 {
+	if frozen < blockRemain {
 		t.Fatalf("block amount is not enough for pruning: %v", err)
 	}
 
 	oldOffSet := rawdb.ReadOffSetOfAncientFreezer(db)
 	// Get the actual start block number.
-	startBlockNumber := frozen - 328 + oldOffSet
+	startBlockNumber := frozen - blockRemain + oldOffSet
 	// Initialize the slice to buffer the block data left.
 	blockList := make([]*types.Block, 0, blockPruneBackUpBlockNumber)
 	receiptsList := make([]types.Receipts, 0, blockPruneBackUpBlockNumber)
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -78,7 +78,8 @@ func TestOfflineBlockPrune(t *testing.T) {
 	node, _ := startEthService(t, gspec, blocks, chaindbPath)
 	defer node.Close()
 
-	testBlockPruner, err := pruner.NewBlockPruner(db, node, oldAncientPath, newAncientPath)
+	//Initialize a block pruner for pruning amount of 328 blocks
+	testBlockPruner, err := pruner.NewBlockPruner(db, node, oldAncientPath, newAncientPath, 328)
 	if err != nil {
 		t.Fatalf("failed to make new blockpruner: %v", err)
 	}
@@ -93,7 +94,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 	defer dbBack.Close()
 
 	//check against if the backup data matched original one
-	for blockNumber := startBlockNumber; blockNumber < startBlockNumber+128; blockNumber++ {
+	for blockNumber := startBlockNumber; blockNumber < startBlockNumber+328; blockNumber++ {
 		blockHash := rawdb.ReadCanonicalHash(dbBack, blockNumber)
 		block := rawdb.ReadBlock(dbBack, blockHash, blockNumber)
 		if reflect.DeepEqual(block, blockList[blockNumber-startBlockNumber]) {
@@ -129,7 +130,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 
 func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Database, []*types.Block, []*types.Block, []types.Receipts, []*big.Int, uint64, *core.BlockChain) {
 	//create a database with ancient freezer
-	db, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, AncientPath, "", false, true)
+	db, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, AncientPath, "", false, false)
 	if err != nil {
 		t.Fatalf("failed to create database with ancient backend")
 	}
@@ -142,8 +143,7 @@ func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Dat
 	}
 
 	// Make chain starting from genesis
-
-	blocks, _ := core.GenerateChain(gspec.Config, genesis, ethash.NewFaker(), db, 200, func(i int, block *core.BlockGen) {
+	blocks, _ := core.GenerateChain(gspec.Config, genesis, ethash.NewFaker(), db, 500, func(i int, block *core.BlockGen) {
 		block.SetCoinbase(common.Address{0: byte(canonicalSeed), 19: byte(i)})
 		tx, err := types.SignTx(types.NewTransaction(block.TxNonce(address), common.Address{0x00}, big.NewInt(1000), params.TxGas, nil, nil), signer, key)
 		if err != nil {
@@ -168,11 +168,14 @@ func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Dat
 	if err != nil || frozen == 0 {
 		t.Fatalf("Failed to import canonical chain start: %v", err)
 	}
+	if frozen < 328 {
+		t.Fatalf("block amount is not enough for pruning: %v", err)
+	}
 
 	oldOffSet := rawdb.ReadOffSetOfAncientFreezer(db)
 	// Get the actual start block number.
-	startBlockNumber := frozen - 128 + oldOffSet
-	// Initialize the slice to buffer the 128 block data.
+	startBlockNumber := frozen - 328 + oldOffSet
+	// Initialize the slice to buffer the block data left.
 	blockList := make([]*types.Block, 0, blockPruneBackUpBlockNumber)
 	receiptsList := make([]types.Receipts, 0, blockPruneBackUpBlockNumber)
 	externTdList := make([]*big.Int, 0, blockPruneBackUpBlockNumber)
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -86,7 +86,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 		t.Fatalf("Failed to back up block: %v", err)
 	}
 
-	dbBack, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, newAncientPath, "", false)
+	dbBack, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, newAncientPath, "", false, true)
 	if err != nil {
 		t.Fatalf("failed to create database with ancient backend")
 	}
@@ -129,7 +129,7 @@ func TestOfflineBlockPrune(t *testing.T) {
 
 func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string) (ethdb.Database, []*types.Block, []*types.Block, []types.Receipts, []*big.Int, uint64, *core.BlockChain) {
 	//create a database with ancient freezer
-	db, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, AncientPath, "", false)
+	db, err := rawdb.NewLevelDBDatabaseWithFreezer(chaindbPath, 0, 0, AncientPath, "", false, true)
 	if err != nil {
 		t.Fatalf("failed to create database with ancient backend")
 	}
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -17,10 +17,16 @@
 package main
 
 import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
 	"io/ioutil"
 	"math/big"
 	"os"
+	"path/filepath"
+	"reflect"
 	"testing"
+	"time"
 
 	"github.com/ethereum/go-ethereum/cmd/utils"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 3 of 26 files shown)


cmd/geth/snapshot.go
```diff
@@ -149,6 +166,55 @@ It's also usable without snapshot enabled.
 	}
 )
 
+func pruneBlock(ctx *cli.Context) error {
```
guagualvcha, line 169: We need some protection here. Make sure the MPT and snapshot matches before when doing prune, otherwise, the node can not start.

Mercybudda: Using the check like this?
`snapshot.New(db, trie.NewDatabase(db), 256, int(triesInMemory), headBlock.Root(), false, false, false)`

cmd/geth/snapshot.go
```diff
@@ -78,6 +79,21 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
 If you specify another directory for the trie clean cache via "--cache.trie.journal"
 during the use of Geth, please also specify it here for correct deletion. Otherwise
 the trie clean cache with default directory will be deleted.
+`,
+			},
+			{
+				Name:      "prune-block",
+				Usage:     "Prune block data offline",
+				ArgsUsage: "<root>",
```
guagualvcha, line 87: `<root>` is not needed here.

cmd/geth/snapshot.go
```diff
@@ -78,6 +79,21 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
...
+				Action:    utils.MigrateFlags(pruneBlock),
+				Category:  "MISCELLANEOUS COMMANDS",
+				Flags: []cli.Flag{
+					utils.DataDirFlag,
+					utils.AncientFlag,
+				},
+				Description: `
+Offline prune for block data.
+For AncientFlag, please specify the absolute path of node/geth/chaindata.
```
guagualvcha, line 96: Why it have to be `absolute` path?

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+	if err := pruner.BlockPrune(oldAncientPath, newAncientPath); err != nil {
+		utils.Fatalf("Failed to prune block", err)
+		return err
+	}
+	//TODO lock.Release()
+	log.Info("Block prune successfully")
+	return nil
+}
+
+func getAncientPath(ctx *cli.Context) string {
```
guagualvcha, line 226: let us follow the same code style. Please remove this func, but use `ctx.GlobalString(utils.AncientFlag.Name)` directly.

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
 	}
 )
 
+func pruneBlock(ctx *cli.Context) error {
+	stack, config := makeConfigNode(ctx)
+	defer stack.Close()
+
+	chaindb := utils.MakeChainDatabaseForBlockPrune(ctx, stack, false)
+	chaindb.Close()
```
guagualvcha, line 173: Are you meaning ` defer chaindb.Close()`?

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+
+	var oldAncientPath, newAncientPath string
+	if path := getAncientPath(ctx); path != "" {
+		oldAncientPath = path + "/ancient"
+		newAncientPath = path + "/ancient_back"
+	} else {
+		utils.Fatalf("Prune failed, did not specify the AncientPath %v")
```
guagualvcha, line 180: Better have a default `ancient ` path.

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+		return errors.New("Prune failed, did not specify the AncientPath")
+	}
+
+	//TODO for filelock
```
guagualvcha, line 184: Will we have file lock?

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+	//lock, _, err := fileutil.Flock(filepath.Join(oldAncientPath, "FLOCK"))
+	// lock, _, err := fileutil.Flock(oldAncientPath)
+	// if err != nil {
+	// 	return err
+	// }
+	// _, err = os.Open(oldAncientPath)
+	// if err != nil {
+	// 	utils.Fatalf("Failed to read genesis file: %v", err)
+	// }
+	for _, name := range []string{"chaindata"} {
```
guagualvcha, line 194: why range a slice that only have one const item?

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+		root := stack.ResolvePath(name)
+		switch {
+		case oldAncientPath == "":
+			oldAncientPath = filepath.Join(root, "ancient")
```
guagualvcha, line 198: It is not possible to happen, since you forbidden it to be empty in previous logic.

cmd/geth/snapshot.go
guagualvcha: You have `%v` here , but no arg.

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+		case !filepath.IsAbs(oldAncientPath):
+			oldAncientPath = stack.ResolvePath(oldAncientPath)
+		}
+		pruner, err := pruner.NewBlockPruner(chaindb, stack, oldAncientPath)
+		if err != nil {
+			utils.Fatalf("Failed to create block pruner", err)
```
guagualvcha, line 204: Get err  arg here, but no `%v`

cmd/geth/snapshot.go
```diff
@@ -149,6 +165,68 @@ It's also usable without snapshot enabled.
...
+		if err := pruner.BlockPruneBackUp(name, config.Eth.DatabaseCache, utils.MakeDatabaseHandles(), newAncientPath, oldAncientPath, "", false); err != nil {
+			log.Error("Failed to back up block", "err", err)
+			return err
+		}
+	}
+
+	log.Info("geth block offline pruning backup successfully")
+
+	//After backing up successfully, rename the new ancientdb name to the original one, and delete the old ancientdb
+	if err := pruner.BlockPrune(oldAncientPath, newAncientPath); err != nil {
```
guagualvcha, line 217: It is quite weried you already pass `oldAncientPath` when `NewBlockPruner`, but pass it again when doing `BlockPruneBackUp`, it is duplicated.

cmd/utils/flags.go
```diff
@@ -1904,6 +1904,28 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.
...
+// MakeChainDatabase open an LevelDB using the flags passed to the client and will hard crash if it fails.
+func MakeChainDatabaseForBlockPrune(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.Database {
+	var (
+		cache   = ctx.GlobalInt(CacheFlag.Name) * ctx.GlobalInt(CacheDatabaseFlag.Name) / 100
+		handles = MakeDatabaseHandles()
+
+		err     error
+		chainDb ethdb.Database
+	)
+	if ctx.GlobalString(SyncModeFlag.Name) == "light" {
```
guagualvcha, line 1916: if sync mode is `light`, there is no need to do block prune

node/node.go
```diff
@@ -634,6 +634,73 @@ func (n *Node) OpenDatabaseWithFreezer(name string, cache, handles int, freezer,
 	return db, err
 }
 
+// The big difference is that the freezer inside was not running to avoid the mess between kvdb and freezer
+// while opening/closing db.
+// OpenDatabaseWithFreezerForPruneBlock opens an existing database with the given name (or
+// creates one if no previous can be found) from within the node's data directory,
+// also attaching a chain freezer to it. If the node is an ephemeral one, a memory database is returned.
+func (n *Node) OpenDatabaseWithFreezerForPruneBlock(name string, cache, handles int, freezer, namespace string, readonly bool) (ethdb.Database, error) {
```
guagualvcha, line 642: Please try not to invent new function, too many duplicated code.

node/node.go
```diff
@@ -634,6 +634,73 @@ func (n *Node) OpenDatabaseWithFreezer(name string, cache, handles int, freezer,
...
+	return db, err
+}
+
+// The big difference is that the freezer inside was not running to avoid the mess between kvdb and freezer
+// while opening/closing db.
+// OpenDatabaseWithFreezerBackup opens an existing database with the given name (or
+// creates one if no previous can be found) from within the node's data directory,
+// also attaching a chain freezer to it. If the node is an ephemeral one, a
+// memory database is returned.
+func (n *Node) OpenDatabaseWithFreezerBackup(offset uint64, name string, cache, handles int, freezer, namespace string, readonly bool) (ethdb.Database, error) {
```
guagualvcha, line 676: Same here

core/state/pruner/pruner.go
```diff
@@ -233,6 +251,87 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+	// Get the number of items in ancient db, frozen it is.
+	frozen, err := chainDb.Ancients()
+
+	// Write the latest 128 blocks data of the ancient db
+	// If we can't access the freezer or it's empty, abort.
+	if err != nil || frozen == 0 {
+		return errors.New("can't access the freezer or it's empty, abort")
+	}
+	// Get the actual start block number.
+	startBlockNumber := frozen - 128 + oldOffSet
```
guagualvcha, line 274: `128` is a magic number here, please use constant value.

core/state/pruner/pruner.go
```diff
@@ -233,6 +251,87 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+		log.Error("Failed to open ancient database: %v", err)
+		return err
+	}
+
+	// Write into ancient_backup
+	for id := 0; id < len(blockList); id++ {
+		rawdb.WriteAncientBlock(chainDbBack, blockList[id], receiptsList[id], externTdList[id])
+	}
+
+	chainDbBack.Close()
```
guagualvcha, line 315: try to use `defer`

cmd/geth/snapshot.go
Mercybudda: No,  this is for closing it on purpose, because in the later code, there'll be many times of open/close the same db, crash will occur if did not close it.

node/node.go
Mercybudda: These new functions were the function without freeze go-routine running, to avoid the case that freeze go-routine did not stop while db closing.

cmd/geth/pruneblock_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Fatalf("failed to create database with ancient backend")
+	}
+
+	testBlockPruneForOffSetZero(t, true, db, newAncientPath, oldAncientPath, chaindbPath)
+
+}
+
+func testBlockPruneForOffSetZero(t *testing.T, full bool, db ethdb.Database, backFreezer, oldfreezer, chaindbPath string) error {
+	// Make chain starting from genesis
+	blockchain, n, err := newCanonical(t, db, ethash.NewFaker(), 92000, full, chaindbPath)
```
guagualvcha, line 75: This is a block chain with empty blocks, it can not test prune  block body, transaction receipt. A block chain with transactions is required.

cmd/geth/pruneblock_test.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		t.Fatalf("failed to make new canonical chain: %v", err)
+	}
+	defer blockchain.Stop()
```
guagualvcha, line 79: Since prune do not need blockchain to be alive, we can stop it rather than defer. No sure, but have a try.

cmd/geth/pruneblock_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	testBlockPruner, err := pruner.NewBlockPruner(db, n, oldfreezer, backFreezer)
+	if err != nil {
+		t.Fatalf("failed to make new blockpruner: %v", err)
+	}
+
+	db.Close()
```
guagualvcha, line 86: we can delete `db.Close()` if we stop the blockchain before

cmd/geth/pruneblock_test.go
guagualvcha, line 75: rename `n` to `node`

cmd/geth/pruneblock_test.go
```diff
@@ -0,0 +1,151 @@
...
+	if err := testBlockPruner.BlockPruneBackUp(chaindbPath, 512, utils.MakeDatabaseHandles(), "", false); err != nil {
+		log.Error("Failed to back up block", "err", err)
+		return err
+	}
+
+	return nil
+
+}
+
```
guagualvcha, line 95: The testcase is not enough. We need query header, receipt, block, transaction before prune, and do query again to verify we do the prune correctly.

cmd/geth/snapshot.go
```diff
@@ -78,6 +82,19 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
...
+				Name:     "prune-block",
+				Usage:    "Prune block data offline",
+				Action:   utils.MigrateFlags(pruneBlock),
+				Category: "MISCELLANEOUS COMMANDS",
+				Flags: []cli.Flag{
+					utils.DataDirFlag,
+					utils.AncientFlag,
+				},
+				Description: `
+Offline prune for block data.
```
guagualvcha, line 97: This is too simple. Please describe it better.

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
 
+func accessDb(ctx *cli.Context, stack *node.Node) (ethdb.Database, error) {
+	chaindb := utils.MakeChainDatabase(ctx, stack, false, true)
+	defer chaindb.Close()
+	headBlock := rawdb.ReadHeadBlock(chaindb)
+	if headBlock == nil {
+		return nil, errors.New("failed to load head block")
+	}
+	//Make sure the MPT and snapshot matches before pruning, otherwise the node can not start.
+	_, err := snapshot.New(chaindb, trie.NewDatabase(chaindb), 256, int(ctx.GlobalUint64(utils.TriesInMemoryFlag.Name)), headBlock.Root(), false, false, false)
```
guagualvcha, line 177: This can no ensure MPT and snapshot matches.

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+		utils.Fatalf("MPT and snapshot sanity check failed: %v", err)
+		return err
+	}
+	var oldAncientPath, newAncientPath string
+	path, _ := filepath.Split(ctx.GlobalString(utils.AncientFlag.Name))
+	if path != "" {
+		if !filepath.IsAbs(path) {
+			path = stack.ResolvePath(path)
+		}
+		oldAncientPath = path + "ancient"
```
guagualvcha, line 198: Can we try use `join` than `+`

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+		newAncientPath = path + "ancient_back"
+	} else {
+		utils.Fatalf("Prune failed, did not specify the AncientPath")
+		return errors.New("Prune failed, did not specify the AncientPath")
```
guagualvcha, line 202: `Fatalf` will do `os.Exit(1)`, please just return, do not `Fatalf`

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+	}
+
+	lock, _, err := fileutil.Flock(filepath.Join(oldAncientPath, "RPUNEFLOCK"))
+	if err != nil {
+		return err
```
guagualvcha, line 207: Try not simply return the err here. But wrap the error with more message, like how can people handle with the Lock exist error

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+	}
+
+	name := "chaindata"
+	if !filepath.IsAbs(oldAncientPath) {
+		oldAncientPath = stack.ResolvePath(oldAncientPath)
```
guagualvcha, line 212: I think `oldAncientPath` is already abs path, because `path` is abs path

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+	}
+
+	blockpruner, err := pruner.NewBlockPruner(chaindb, stack, oldAncientPath, newAncientPath)
+	if err != nil {
+		utils.Fatalf("Failed to create block pruner: %v", err)
```
guagualvcha, line 217: same here. Please do not use `Fatal` if you want to return the err.

cmd/geth/snapshot.go
```diff
@@ -149,6 +166,76 @@ It's also usable without snapshot enabled.
...
+
+	if err := blockpruner.BlockPruneBackUp(name, config.Eth.DatabaseCache, utils.MakeDatabaseHandles(), "", false); err != nil {
+		log.Error("Failed to back up block", "err", err)
+		return err
+	}
+
+	log.Info("geth block offline pruning backup successfully")
+
+	//After backing up successfully, rename the new ancientdb name to the original one, and delete the old ancientdb
+	if err := pruner.BlockPrune(oldAncientPath, newAncientPath); err != nil {
```
guagualvcha, line 229: It is really confusing between `blockpruner.BlockPruneBackUp` and `pruner.BlockPrune`

core/rawdb/database.go
```diff
@@ -151,12 +151,16 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
 // NewDatabaseWithFreezer creates a high level database on top of a given key-
 // value data store with a freezer moving immutable chain segments into cold
 // storage.
-func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace string, readonly bool) (ethdb.Database, error) {
+func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace string, args ...bool) (ethdb.Database, error) {
 	// Create the idle freezer instance
-	frdb, err := newFreezer(freezer, namespace, readonly)
+	frdb, err := newFreezer(freezer, namespace, args[0])
```
guagualvcha, line 156: Please try not use some var like `args`, especially the type is bool

core/rawdb/database.go
```diff
@@ -179,50 +183,54 @@ func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace st
...
+				// Otherwise, the head header is still the genesis, we're allowed to init a new
+				// feezer.
 			}
-			// Otherwise, the head header is still the genesis, we're allowed to init a new
-			// feezer.
 		}
 	}
 	// Freezer is consistent with the key-value database, permit combining the two
-	if !frdb.readonly {
+	if len(args) == 1 && !frdb.readonly {
```
guagualvcha, line 233: This is tricky.

core/state/pruner/pruner.go
```diff
@@ -85,6 +91,13 @@ type Pruner struct {
 	triesInMemory uint64
 }
 
+type BlockPruner struct {
+	db             ethdb.Database
+	oldAncientPath string
+	newAncientPath string
+	n              *node.Node
```
guagualvcha, line 98: `n` --> `node`

core/rawdb/database.go
Mercybudda: Do you mean rename this variable?

yutianwu: you don't use like args[1] in this function, why do you change the signature of this function

core/state/pruner/pruner.go
```diff
@@ -233,6 +256,96 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
 	return nil
 }
 
+func (p *BlockPruner) backUpOldDb(name string, cache, handles int, freezer, namespace string, args ...bool) ([]*types.Block, []types.Receipts, []*big.Int, error) {
```
yutianwu, line 259: `backUp` -> `backup`

core/state/pruner/pruner.go
```diff
@@ -233,6 +256,96 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+	return blockList, receiptsList, externTdList, nil
+}
+
+// Backup the ancient data for the old ancient db, i.e. the most recent 128 blocks in ancient db.
+func (p *BlockPruner) BlockPruneBackUp(name string, cache, handles int, namespace string, readonly bool) error {
+
+	start := time.Now()
+	isBlockPruner := true
+
+	blockList, receiptsList, externTdList, err := p.backUpOldDb(name, cache, handles, p.oldAncientPath, namespace, readonly, isBlockPruner)
```
yutianwu, line 313: `isBlockPruner` is always true and `backUpOldDb` is called only here, why bother add `args`

core/rawdb/database.go
```diff
@@ -151,12 +151,16 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace string, args ...bool) (ethdb.Database, error) {
 	// Create the idle freezer instance
-	frdb, err := newFreezer(freezer, namespace, readonly)
+	frdb, err := newFreezer(freezer, namespace, args[0])
 	if err != nil {
 		return nil, err
 	}
+	offset := ReadOffSetOfAncientFreezer(db)
+
+	frdb.offset = offset
```
yutianwu, line 162: why not read offset in `newFreezer` above

core/rawdb/database.go
yutianwu: wouldn't it be bettter to add another parameter here

core/rawdb/database.go
```diff
@@ -179,50 +183,54 @@ func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace st
...
-			// are contiguous, otherwise we might end up with a non-functional freezer.
-			if kvhash, _ := db.Get(headerHashKey(frozen)); len(kvhash) == 0 {
-				// Subsequent header after the freezer limit is missing from the database.
-				// Reject startup is the database has a more recent head.
-				if *ReadHeaderNumber(db, ReadHeadHeaderHash(db)) > frozen-1 {
-					return nil, fmt.Errorf("gap (#%d) in the chain between ancients and leveldb", frozen)
+	// Only to check the followings when offset equal to 0, otherwise the block number
+	// in ancientdb did not start with 0, no genesis block in ancientdb as well.
+	if offset == 0 {
+		if kvgenesis, _ := db.Get(headerHashKey(0)); len(kvgenesis) > 0 {
```
yutianwu, line 189: what if we change this to `if kvgenesis, _ := db.Get(headerHashKey(0)); len(kvgenesis) > 0 && offset == 0`

core/rawdb/database.go
```diff
@@ -256,12 +264,12 @@ func NewLevelDBDatabase(file string, cache int, handles int, namespace string, r
 
 // NewLevelDBDatabaseWithFreezer creates a persistent key-value database with a
 // freezer moving immutable chain segments into cold storage.
-func NewLevelDBDatabaseWithFreezer(file string, cache int, handles int, freezer string, namespace string, readonly bool) (ethdb.Database, error) {
-	kvdb, err := leveldb.New(file, cache, handles, namespace, readonly)
+func NewLevelDBDatabaseWithFreezer(file string, cache int, handles int, freezer string, namespace string, args ...bool) (ethdb.Database, error) {
```
yutianwu, line 267: can you think twice about all functions with this `args` parameter

cmd/geth/snapshot.go
yutianwu: and can you add the side effects if we prune the blocks like we can not query the history of blocks

cmd/geth/snapshot.go
yutianwu: right, and what if `path` don't have `/`, `path.Join` should help

core/rawdb/database.go
Mercybudda: This is to assign offset to `newFreezer`, the above `newFreezer` is the initialization without offset.

core/rawdb/database.go
Mercybudda: I'm worried that if adding another parameter obviously, there will be too many places to update beyond this feature, will be kind of messy

cmd/geth/pruneblock_test.go
Mercybudda: Done

Mercybudda: Done

cmd/geth/pruneblock_test.go
Mercybudda: Done

cmd/geth/pruneblock_test.go
Mercybudda: Done

cmd/geth/pruneblock_test.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Good point! Done

core/state/pruner/pruner.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: The reason for introducing `args` param in is that I don't want the goroutine of frdb.freeze(db) running,  we can distinguish the offline block prune scenario if len(args) > 1. For this feature, there'll be many opening the same kvdb and two ancientdb, so running of this goroutine will bring unnecessary trouble even crash like db closed but freezer didn't at the same time. In addition, for geth offline case, it did not make sense that the item still move between db.

yutianwu: I mean the `args` itself is meaningless, we don's know what the `args` are at the first glance

core/rawdb/database.go
yutianwu: I mean why not init the offset in `newFreezer`

Mercybudda: because I think`newFreezer` func did not have the parameter of db to read offset through, right?

core/rawdb/database.go
Mercybudda: Done

core/state/pruner/pruner.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Done for adding another parameter here to indicate blockprune feature

cmd/geth/snapshot.go
```diff
@@ -78,6 +83,26 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
...
+				Usage:    "Prune block data offline",
+				Action:   utils.MigrateFlags(pruneBlock),
+				Category: "MISCELLANEOUS COMMANDS",
+				Flags: []cli.Flag{
+					utils.DataDirFlag,
+					utils.AncientFlag,
+					utils.BlockPruneQuantity,
+				},
+				Description: `
+geth offline prune-block for block data in ancientdb.
```
yutianwu, line 99: is there any side effect that we can not query old blocks any more

Mercybudda: Have added plus feature for specifying the amount of old blocks that we want to prune, I think this will avoid the side effect you mentioned.

RenRick (approved):
LGTM

core/state/pruner/pruner.go
```diff
@@ -233,6 +257,111 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+	}
+
+	//write back-up data to new chainDb
+	// Restore the last known head block
+
+	//write genesis block firstly
+	genesis := p.genesis
+	if _, _, err := core.SetupGenesisBlock(chainDbBack, genesis); err != nil {
+		log.Error("Failed to write genesis block: %v", err)
+		return err
```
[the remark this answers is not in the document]
Mercybudda: The code in this commit was already outdated and invalid. There's no manually db closing anymore for updated code, it will close automatically by `defer db.Close()`

cmd/utils/flags.go
```diff
@@ -827,6 +827,11 @@ var (
 		Name:  "catalyst",
 		Usage: "Catalyst mode (eth2 integration testing)",
 	}
+
+	BlockPruneAmountLeft = cli.Uint64Flag{
```
guagualvcha, line 831: Flag need match with name

core/rawdb/database.go
```diff
@@ -256,12 +274,12 @@ func NewLevelDBDatabase(file string, cache int, handles int, namespace string, r
 
 // NewLevelDBDatabaseWithFreezer creates a persistent key-value database with a
 // freezer moving immutable chain segments into cold storage.
-func NewLevelDBDatabaseWithFreezer(file string, cache int, handles int, freezer string, namespace string, readonly bool) (ethdb.Database, error) {
+func NewLevelDBDatabaseWithFreezer(file string, cache int, handles int, freezer string, namespace string, readonly, isPruneBlock bool) (ethdb.Database, error) {
```
guagualvcha, line 277: This is a `rawdb`, it is not proper to see any wording like `isPruneBlock`

core/rawdb/database.go
```diff
@@ -221,8 +238,9 @@ func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace st
 			// feezer.
 		}
 	}
+
 	// Freezer is consistent with the key-value database, permit combining the two
-	if !frdb.readonly {
+	if !isPruneBlock && !frdb.readonly {
```
guagualvcha, line 243: Please come up with better name rather than `isPruneBlock`.

guagualvcha (changes requested):

core/rawdb/database.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Done

cmd/utils/flags.go
Mercybudda: Done

cmd/geth/dbcmd.go
```diff
@@ -282,12 +293,21 @@ func inspect(ctx *cli.Context) error {
...
 	defer stack.Close()
 
-	db := utils.MakeChainDatabase(ctx, stack, true)
+	db := utils.MakeChainDatabase(ctx, stack, true, false)
 	defer db.Close()
 
 	return rawdb.InspectDatabase(db, prefix, start)
 }
 
+func blockPruneInspect(ctx *cli.Context) error {
```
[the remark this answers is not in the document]
Mercybudda: Done

cmd/geth/main.go
```diff
@@ -165,6 +165,7 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
+		utils.BlockPruneAmountLeft,
```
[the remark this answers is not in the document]
Mercybudda: Updated

cmd/geth/snapshot.go
```diff
@@ -78,6 +83,28 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
...
+The amount of blocks expected for remaining after prune can be specified via block-amount-reserved in this command,
+will prune and only remain the specified amount of old block data in ancientdb.
+the brief workflow is to backup the the number of this specified amount blocks backward in original ancientdb 
+into new ancient_backup, then delete the original ancientdb dir and rename the ancient_backup to original one for replacement,
+finally assemble the statedb and new ancientDb together.
+The purpose of doing it is because the block data will be moved into the ancient store when it
+becomes old enough(exceed the Threshold 90000), the disk usage will be very large over time, and is occupied mainly by ancientDb,
+so it's very necessary to do block data prune, this feature takes very short time in even seconds level.
```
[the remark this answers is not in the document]
Mercybudda: Updated

cmd/geth/snapshot.go
```diff
@@ -78,6 +83,28 @@ WARNING: It's necessary to delete the trie clean cache after the pruning.
...
+			},
+			{
+				Name:     "prune-block",
+				Usage:    "Prune block data offline",
+				Action:   utils.MigrateFlags(pruneBlock),
+				Category: "MISCELLANEOUS COMMANDS",
+				Flags: []cli.Flag{
+					utils.DataDirFlag,
+					utils.AncientFlag,
+					utils.BlockPruneAmountLeft,
```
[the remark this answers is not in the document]
Mercybudda: Done

cmd/geth/snapshot.go
```diff
@@ -149,11 +176,155 @@ It's also usable without snapshot enabled.
...
+	}
+	headHeader := headBlock.Header()
+	//Make sure the MPT and snapshot matches before pruning, otherwise the node can not start.
+	snaptree, err := snapshot.New(chaindb, trie.NewDatabase(chaindb), 256, int(ctx.GlobalUint64(utils.TriesInMemoryFlag.Name)), headBlock.Root(), false, false, false)
+	if err != nil {
+		return nil, err // The relevant snapshot(s) might not exist
+	}
+
+	var targetRoot common.Hash
+	if ctx.NArg() == 1 {
```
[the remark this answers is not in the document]
Mercybudda: Have deleted this check, will use HEAD-(n-1) as the target root

cmd/geth/snapshot.go
```diff
@@ -149,11 +176,155 @@ It's also usable without snapshot enabled.
...
+		targetRoot, err = parseRoot(ctx.Args()[0])
+		if err != nil {
+			log.Error("Failed to resolve state root", "err", err)
+			return nil, err
+		}
+	}
+
+	triesInMemory := ctx.GlobalUint64(utils.TriesInMemoryFlag.Name)
```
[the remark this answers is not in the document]
Mercybudda: Updated

cmd/geth/snapshot.go
```diff
@@ -149,11 +176,155 @@ It's also usable without snapshot enabled.
...
+	// If the target state root is not specified, use the HEAD-(n-1) as the
+	// target. The reason for picking it is:
+	// - in most of the normal cases, the related state is available
+	// - the probability of this layer being reorg is very low
+	var layers []snapshot.Snapshot
+	if targetRoot == (common.Hash{}) {
```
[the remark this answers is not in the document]
Mercybudda: Updated, will not check targetRoot but use HEAD-(n-1) as the target root

cmd/geth/snapshot.go
```diff
@@ -28,9 +31,12 @@ import (
 	"github.com/ethereum/go-ethereum/core/state/pruner"
 	"github.com/ethereum/go-ethereum/core/state/snapshot"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
+	"github.com/prometheus/tsdb/fileutil"
```
yutianwu, line 39: can you fix the imports here

Mercybudda: Done

cmd/utils/flags.go
```diff
@@ -1883,7 +1888,7 @@ func SplitTagsFlag(tagsFlag string) map[string]string {
 }
 
 // MakeChainDatabase open an LevelDB using the flags passed to the client and will hard crash if it fails.
-func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly bool) ethdb.Database {
+func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly, disablFreeze bool) ethdb.Database {
```
yutianwu, line 1891: typo `disablFreeze` -> disableFreeze

Mercybudda: Updated. I'm aware of this and my intention was to simplify the variable, but anyway I updated.

yutianwu (approved):

cmd/geth/snapshot.go
```diff
@@ -40,6 +47,9 @@ var (
 
 	// emptyCode is the known hash of the empty EVM bytecode.
 	emptyCode = crypto.Keccak256(nil)
+
+	//the layer of tries trees that keep in memory
+	TriesInMemory = 128
```
guagualvcha, line 52: `TriesInMemory` is not a fixed value,  default value is 128, but some node use other values.

cmd/geth/main.go
```diff
@@ -165,6 +165,7 @@ var (
 		utils.MinerNotifyFullFlag,
 		configFileFlag,
 		utils.CatalystFlag,
+		utils.BlockAmountReserved,
```
guagualvcha, line 168: unnecessry?

cmd/geth/snapshot.go
```diff
@@ -149,11 +181,163 @@ It's also usable without snapshot enabled.
...
+			return nil, fmt.Errorf("associated state[%x] is not present", targetRoot)
+		}
+	} else {
+		if len(layers) > 0 {
+			log.Info("Selecting bottom-most difflayer as the pruning target", "root", targetRoot, "height", headHeader.Number.Uint64()-127)
+		} else {
+			log.Info("Selecting user-specified state as the pruning target", "root", targetRoot)
+		}
+	}
+	if err := snaptree.Verify(targetRoot); err != nil {
```
guagualvcha, line 260: No need to verify

cmd/geth/snapshot.go
```diff
@@ -149,11 +181,163 @@ It's also usable without snapshot enabled.
...
+		return err
+	}
+
+	lock, exist, err := fileutil.Flock(filepath.Join(oldAncientPath, "PRUNEFLOCK"))
+	if err != nil {
+		log.Error("file lock error", "err", err)
+		return err
+	}
+	if exist {
+		log.Debug("file lock existed, waiting for prune recovery and continue", "err", err)
```
guagualvcha, line 298: Please change `Debug` to `Warn`, no err is need for the log because it will always be nil

cmd/geth/snapshot.go
```diff
@@ -149,11 +181,163 @@ It's also usable without snapshot enabled.
...
+		log.Info("Block prune successfully")
+		return nil
+	}
+	name := "chaindata"
+	if err := blockpruner.BlockPruneBackUp(name, config.Eth.DatabaseCache, utils.MakeDatabaseHandles(), "", false, false); err != nil {
+		log.Error("Failed to back up block", "err", err)
+		return err
+	}
+
+	log.Info("geth block offline pruning backup successfully")
```
guagualvcha, line 324: `"geth block offline pruning backup successfully"` ==> `back up block successfully` is fine.

core/rawdb/database.go
```diff
@@ -148,15 +151,89 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+}
+
+//NewFreezerDb only create a freezer without statedb.
+func NewFreezerDb(db ethdb.KeyValueStore, frz, namespace string, readonly bool, lastOffSet, newOffSet uint64) (*freezer, error) {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	// Create the idle freezer instance, this operation should be atomic to avoid mismatch between offset and acientDB.
+	var frdb *freezer
+	var errors error
+	go func() {
```
guagualvcha, line 188: Why need another go routine here?

core/rawdb/database.go
```diff
@@ -148,15 +151,89 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+func NewDatabaseWithFreezer(db ethdb.KeyValueStore, freezer string, namespace string, readonly, disableFreeze bool) (ethdb.Database, error) {
 	// Create the idle freezer instance
 	frdb, err := newFreezer(freezer, namespace, readonly)
 	if err != nil {
 		return nil, err
 	}
+
+	var offset uint64
+	path, name := filepath.Split(freezer)
+	if _, err := os.Stat(filepath.Join(path, "ancient_back")); os.IsNotExist(err) {
```
guagualvcha, line 221: try remove `ancient_back` and `ancient` here.

core/rawdb/database.go
Mercybudda: Updated

core/rawdb/database.go
Mercybudda: I hardcoded "ancient_back" within the code logic in the same directory tree of old ancient, no need to be specified by user for the new ancientdb directory, so we can assume `ancient_back` as new ancientDB. I removed `ancient` and will use the exact path same to user specified.

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/main.go
Mercybudda: I think it should be necessary? Because it will initialize the config, which will be used for GlobalFinding  the value of ctx

cmd/geth/snapshot.go
Mercybudda: Done, introduced in the user CLI command.

cmd/geth/dbcmd.go
```diff
@@ -195,6 +196,16 @@ WARNING: This is a low-level operation which may cause database corruption!`,
...
 	}
+	ancientInspectCmd = cli.Command{
+		Action: utils.MigrateFlags(ancientInspect),
+		Name:   "inspect-reserved-oldest-blocks",
+		Flags: []cli.Flag{
+			utils.DataDirFlag,
+		},
+		Usage: "Inspect the ancientStore information",
+		Description: `This commands will read current offset from kvdb, which is the current offset and starting BlockNumber
+		of ancientStore, will also displays the reserved number of blocks in ancientStore `,
```
guagualvcha, line 207: No newline is allowed here.  I am getting:

core/rawdb/database.go
```diff
@@ -148,15 +151,74 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	// Create the idle freezer instance, this operation should be atomic to avoid mismatch between offset and acientDB.
+	frdb, errors := newFreezer(frz, namespace, readonly)
+	if errors != nil {
+		return nil, errors
+	}
+	frdb.offset = newOffSet
+	offsetBatch := db.NewBatch()
+	WriteOffSetOfCurrentAncientFreezer(offsetBatch, newOffSet)
```
guagualvcha, line 192: It is not proper to do `WriteOffSetOfCurrentAncientFreezer` inside `NewFreezerDb`.

core/rawdb/database.go
```diff
@@ -148,15 +151,74 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+}
+func WriteOffSetOfLastAncientFreezer(db ethdb.KeyValueWriter, offset uint64) {
+	if err := db.Put(offSetOfLastAncientFreezer, new(big.Int).SetUint64(offset).Bytes()); err != nil {
+		log.Crit("Failed to store offSetOfAncientFreezer", "err", err)
+	}
+}
+
+//NewFreezerDb only create a freezer without statedb.
+func NewFreezerDb(db ethdb.KeyValueStore, frz, namespace string, readonly bool, lastOffSet, newOffSet uint64) (*freezer, error) {
+	wg := sync.WaitGroup{}
```
guagualvcha, line 183: There is no need for `waitgroup`

guagualvcha (approved):

core/rawdb/database.go
Mercybudda: Separating this function is to ensure that two kinds of offsets are written into kvdb by the time creating ancient freezer. Any suggestion for this out of place you mentioned?

guagualvcha: I know why to separate the function. `WriteOffSetOfCurrentAncientFreezer ` should be placed in the main logic of block prune.

core/rawdb/database.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Done

cmd/geth/dbcmd.go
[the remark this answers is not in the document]
Mercybudda: Done

guagualvcha (changes requested):
frezzer need change

Mercybudda:
> frezzer need change

Done and QA team has done the finalized tests, all passed.
As a plus, introduced the bool flag `check-snapshot-with-mpt` to enable/disable checking between MPT and snapshot. Because there's no need for that check in fast node scenario.

Mercybudda:
@guagualvcha Updated the frozen filed logic in freezer, the frozen number will not only be the number of items in current ancientDB but also will count all blocks pruned in previous ancientDB if any.

core/rawdb/freezer.go
```diff
@@ -164,22 +166,22 @@ func (f *freezer) Close() error {
...
 func (f *freezer) Ancient(kind string, number uint64) ([]byte, error) {
 	if table := f.tables[kind]; table != nil {
-		return table.Retrieve(number)
+		return table.Retrieve(number - f.offset)
 	}
 	return nil, errUnknownTable
 }
 
 // Ancients returns the length of the frozen items.
 func (f *freezer) Ancients() (uint64, error) {
```
guagualvcha, line 183: Please check all the usage of `Ancients`, make sure they are all right.

Mercybudda: Done, for the usage of `Ancients`, introduced a function for return actual number of items in current ancientDB, and in the usage area of `Ancients`, some means the actual number, some means all the frozen blocks including pruned in this feature, all handled case by case, please check.

guagualvcha (approved):

yutianwu (approved):

cmd/geth/snapshot.go
```diff
@@ -149,11 +180,168 @@ It's also usable without snapshot enabled.
...
+
+	// Use the HEAD-(n-1) as the target root. The reason for picking it is:
+	// - in most of the normal cases, the related state is available
+	// - the probability of this layer being reorg is very low
+
+	// Retrieve all snapshot layers from the current HEAD.
+	// In theory there are n difflayers + 1 disk layer present,
+	// so n diff layers are expected to be returned.
+	layers := snaptree.Snapshots(headHeader.Root, TriesInMemory, true)
+	if len(layers) != TriesInMemory {
```
j75689, line 212: `if len(layers) >= TriesInMemory {` ? In my experience, the number of difflayers will be larger than `TriesMemory` because `stateDB` does not merge difflayers every time, it depends on `AggregatorMemoryLimit`

cmd/geth/snapshot.go
```diff
@@ -149,11 +180,168 @@ It's also usable without snapshot enabled.
...
+		if !found {
+			if len(layers) > 0 {
+				log.Error("no snapshot paired state")
+				return nil, errors.New("no snapshot paired state")
+			}
+			return nil, fmt.Errorf("associated state[%x] is not present", targetRoot)
+		}
+	} else {
+		if len(layers) > 0 {
+			log.Info("Selecting bottom-most difflayer as the pruning target", "root", targetRoot, "height", headHeader.Number.Uint64()-127)
```
j75689, line 262: `headHeader.Number.Uint64()-127` ==> `headHeader.Number.Uint64()-len(layers)`?

cmd/geth/snapshot.go
```diff
@@ -149,11 +180,168 @@ It's also usable without snapshot enabled.
...
+		return err
+	}
+	var newAncientPath string
+	oldAncientPath := ctx.GlobalString(utils.AncientFlag.Name)
+	if !filepath.IsAbs(oldAncientPath) {
+		oldAncientPath = stack.ResolvePath(oldAncientPath)
+	}
+
+	path, _ := filepath.Split(oldAncientPath)
+	if path != "" {
```
j75689, line 285: suggestion, keep the code style and handle the unexpected first
```go
if path == "" {
	return errors.New("prune failed, did not specify the AncientPath")
}
newAncientPath = filepath.Join(path, "ancient_back")
```

cmd/geth/snapshot.go
```diff
@@ -149,11 +180,168 @@ It's also usable without snapshot enabled.
...
+		log.Error("file lock error", "err", err)
+		return err
+	}
+	if exist {
+		log.Info("file lock existed, waiting for prune recovery and continue", "err", err)
+		if err := blockpruner.RecoverInterruption("chaindata", config.Eth.DatabaseCache, utils.MakeDatabaseHandles(), "", false); err != nil {
+			log.Error("Pruning failed", "err", err)
+			return err
+		}
+		lock.Release()
```
j75689, line 307: why not use `defer lock.Release()` on line:301, It's better to release `lock` before function returns

core/rawdb/database.go
```diff
@@ -148,15 +159,69 @@ func NewDatabase(db ethdb.KeyValueStore) ethdb.Database {
...
+func WriteOffSetOfLastAncientFreezer(db ethdb.KeyValueWriter, offset uint64) {
+	if err := db.Put(offSetOfLastAncientFreezer, new(big.Int).SetUint64(offset).Bytes()); err != nil {
+		log.Crit("Failed to store offSetOfAncientFreezer", "err", err)
+	}
+}
+
+// NewFreezerDb only create a freezer without statedb.
+func NewFreezerDb(db ethdb.KeyValueStore, frz, namespace string, readonly bool, newOffSet uint64) (*freezer, error) {
+	// Create the idle freezer instance, this operation should be atomic to avoid mismatch between offset and acientDB.
+	frdb, errors := newFreezer(frz, namespace, readonly)
```
j75689, line 192: `frdb, errors := newFreezer(frz, namespace, readonly)` ==> `frdb, err := newFreezer(frz, namespace, readonly)`

core/rawdb/database.go
```diff
@@ -298,6 +367,32 @@ func (s *stat) Size() string {
 func (s *stat) Count() string {
 	return s.count.String()
 }
+func AncientInspect(db ethdb.Database) error {
```
j75689, line 370: seems this function never return an error

core/state/pruner/pruner.go
```diff
@@ -115,6 +129,16 @@ func NewPruner(db ethdb.Database, datadir, trieCachePath string, bloomSize, trie
 	}, nil
 }
 
+func NewBlockPruner(db ethdb.Database, n *node.Node, oldAncientPath, newAncientPath string, BlockAmountReserved uint64) (*BlockPruner, error) {
```
j75689, line 132: no need to return error

core/state/pruner/pruner.go
```diff
@@ -233,6 +257,197 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+
+	offsetBatch := chainDb.NewBatch()
+	rawdb.WriteOffSetOfCurrentAncientFreezer(offsetBatch, startBlockNumber)
+	rawdb.WriteOffSetOfLastAncientFreezer(offsetBatch, oldOffSet)
+	if err := offsetBatch.Write(); err != nil {
+		log.Crit("Failed to write offset into disk", "err", err)
+	}
+
+	// It's guaranteed that the old/new offsets are updated as well as the new ancientDB are created if this flock exist.
+	lock, _, err := fileutil.Flock(filepath.Join(p.newAncientPath, "PRUNEFLOCKBACK"))
```
j75689, line 319: it's better to use `defer lock.Release()`?

core/state/pruner/pruner.go
```diff
@@ -233,6 +257,197 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+		return err
+	}
+
+	if newExist {
+		log.Info("New ancientDB_backup existed in interruption scenario")
+		var flockOfAncientBack bool
+		if exist, err := CheckFileExist(filepath.Join(p.newAncientPath, "PRUNEFLOCKBACK")); err != nil {
+			log.Error("Failed to check flock of ancientDB_Back %v", err)
+			return err
+		} else {
```
j75689, line 381: this `else` is unnecessary

core/state/pruner/pruner.go
```diff
@@ -233,6 +257,197 @@ func prune(snaptree *snapshot.Tree, root common.Hash, maindb ethdb.Database, sta
...
+		}
+	}
+
+	return nil
+}
+
+func CheckFileExist(path string) (bool, error) {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			//Indicating the file didn't exist.
```
j75689, line 428: `//Indicating the file didn't exist.` ==> `// Indicating the file didn't exist.`

ethdb/database.go
```diff
@@ -81,6 +81,12 @@ type AncientReader interface {
 
 	// AncientSize returns the ancient size of the specified category.
 	AncientSize(kind string) (uint64, error)
+
+	// ItemAmountInAncient returns the actual length of current ancientDB.
+	ItemAmountInAncient() (uint64, error)
+
+	// AncientOffSet returns the offset of current ancientDB.
+	AncientOffSet() (uint64, error)
```
j75689, line 89: this function never returns an error.

cmd/geth/snapshot.go
Mercybudda: I think in `p.snaptree.Snapshots()`,we have specified the  `TriesInMemory` as layer amount limit, so the case for `len(layers) >= TriesInMemory` will not exist?

cmd/geth/snapshot.go
Mercybudda: Done, updated to `headHeader.Number.Uint64()-(len(layers)-1)`

cmd/geth/snapshot.go
Mercybudda: Done

cmd/geth/snapshot.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Done

core/rawdb/database.go
Mercybudda: Done

core/state/pruner/pruner.go
Mercybudda: Done

core/state/pruner/pruner.go
Mercybudda: For handle the case of interruption by users, there's a kind of logic for reading old or current offset based on the exist of this filelock. I think the file lock will be removed by deferred if the function returned in advance for any reason, which is not expected and impact on the logic for reading offset.

core/state/pruner/pruner.go
Mercybudda: This  `else` will assign exist to flockOfAncientBack, right?

ethdb/database.go
Mercybudda: Yes, but actually I found that there're many other places used the same way as well, like `Ancients()`, will be a problem?

core/state/pruner/pruner.go
Mercybudda: Done

core/state/pruner/pruner.go
j75689: If `if exists, error: = CheckFileExist(filepath.Join(p.newAncientPath, "PRUNEFLOCKBACK")); err != nil {` will return this function and will not continue, so you can ignore else

ethdb/database.go
Mercybudda: Done

core/state/pruner/pruner.go
Mercybudda: Done

j75689 (approved):

PR #588, opened 2021-11-24
Optimized db storage

Mercybudda:
This PR re-opened and refined the change from PR:
https://github.com/binance-chain/bsc/pull/384
This is the optimization for database storage

Rationale
Block data is written to leveldb for two times. It occupied too much disk bandwidth. And as block data is sequential and less read operations, compaction to block data also takes disk bandwidth/ cpu/ memory.
State data in KV Store needs to do compaction in order to get better read performance. If block data and state data in the same KV Store, compaction will take more time(more data/ cpu/ memory/ disk bandwidth). Indeed we just need current world state(or latest world state) to execute. In future we needs other solution to get better performance.

I/O test in AWS test machine:
From test, the I/O performance improved roughly 1/3 comparing with the old way

Pending for the stress test from QA team

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,5 +46,5 @@ profile.cov
 /dashboard/assets/bundle.js
 /dashboard/assets/bundle.js.map
 /dashboard/assets/package-lock.json
-
+.history/
 **/yarn-error.log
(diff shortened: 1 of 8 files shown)


Mercybudda:
suspended

PR #668, opened 2021-12-16
[R4R] implement State Verification && Snapshot Commit pipeline

guagualvcha:
## Motivation 

The state verification && Snapshot/MPT commit takes a lot of time.  While in most cases, the verification can pass, there is no need to wait for the verification result serially. For snapshot/MPT commit, we only persist the bottom diff layer to disk, there is no need to wait for it to finish.  We propose an asynchronous state verification and asynchronous snapshot/MPT commit solution.

## Spec

The insert blockchain can be split into three-phase:
1. Process. Apply the transactions within the blocks.
2. Apply the state change into MPT and calculate the state root, verify it against the block header.
3. Commit MPT and Snapshot.

The three-phase are executed serially. 
Phase2 occupies about 15-20% of the time, and in most case, it should succeed.
Phase3 occupies about 10-15% of the time. 

### State Verification Pipeline
A pipeline workflow is introduced in this PR:

image

Once **Block Process** and **Snapshot Commit** are done, there is no need to wait for **State Verification**, the blockchain can process the next block. While for the next block, before it starts **State Verification**, it has to wait until the **State Verification** of its parent finish.

### Persistent Storage Pipeline
Both MPT and Snapshot need to be persisted into storage,  the storage can be Memory DB and Disk DB, this PR will introduce the asynchronous persistence of Disk DB without breaking the processing of the following blocks.

#### How to store the Verification Result

Since Snapshot already has a very good layout for the execution result of the block, we try to reuse it by adding a verifiedCh channel as the verification result.

#### How to Rollback if Verification Failed

There is an auto rollback mechanism when verification is failed.

#### How to avoid race condition
Introduce a few locks, but ensure it won't bring too much overhead.

### Performance
Following is the 24hours comparison of a Pipeline node(10.211.32.131) and a normal node(10.211.31.49)。

Hardware: 24core, 96G, SSD, storage occupation 2.5T

#### Execution Time
Before: 

image

After:
image

The execution time is mainly improved by this PR: https://github.com/binance-chain/bsc/pull/704, while in this PR, we further improve it by avoiding the CPU usage burst.

#### Validation Time
Before:
image

After:
image

#### Commit Time
Before:
image

After:
image

#### CommitTxs Time

Before: 

image

After:
image

#### Merge Multi Difflayer and Do Verification Once
1. How to solve fork issue?
2. How to avoid breaking the trie prefetcher?

#### Improve `AccountsIntermediateRoot`
1. Do prehash?
2. Calculation concurrently?

#### Improve Miner Module
1. Improve L3 cache hit rate during mining.
2. Reduce L2 cache overhead.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -127,6 +127,10 @@ var (
 		Usage: "Enable diffy sync, Please note that enable diffsync will improve the syncing speed, " +
 			"but will degrade the security to light client level",
 	}
+	PipeCommitFlag = cli.BoolFlag{
+		Name:  "pipecommit",
+		Usage: "Enable pipeline commit feature",
+	}
 	RangeLimitFlag = cli.BoolFlag{
 		Name:  "rangelimit",
 		Usage: "Enable 5000 blocks limit for range query",
@@ -1631,6 +1635,9 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	}
 	if ctx.GlobalIsSet(DiffSyncFlag.Name) {
 		cfg.DiffSync = ctx.GlobalBool(DiffSyncFlag.Name)
+	}
+	if ctx.GlobalIsSet(PipeCommitFlag.Name) {
+		cfg.PipeCommit = ctx.GlobalBool(PipeCommitFlag.Name)
 	}
 	if ctx.GlobalIsSet(RangeLimitFlag.Name) {
 		cfg.RangeLimit = ctx.GlobalBool(RangeLimitFlag.Name)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -129,7 +129,7 @@ var (
 	}
 	PipeCommitFlag = cli.BoolFlag{
 		Name:  "pipecommit",
-		Usage: "Enable MPT pipeline commit, it will improve syncing performance. It is an experimental feature",
+		Usage: "Enable MPT pipeline commit, it will improve syncing performance. It is an experimental feature(default is false)",
 	}
 	RangeLimitFlag = cli.BoolFlag{
 		Name:  "rangelimit",
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -129,7 +129,7 @@ var (
 	}
 	PipeCommitFlag = cli.BoolFlag{
 		Name:  "pipecommit",
-		Usage: "Enable pipeline commit feature",
+		Usage: "Enable MPT pipeline commit, it will improve syncing performance. It is an experimental feature",
 	}
 	RangeLimitFlag = cli.BoolFlag{
 		Name:  "rangelimit",
--- a/consensus/clique/clique.go
+++ b/consensus/clique/clique.go
@@ -550,7 +550,7 @@ func (c *Clique) Prepare(chain consensus.ChainHeaderReader, header *types.Header
 func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs *[]*types.Transaction, uncles []*types.Header,
 	receipts *[]*types.Receipt, _ *[]*types.Transaction, _ *uint64) (err error) {
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
-	header.Root, err = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
 	header.UncleHash = types.CalcUncleHash(nil)
 	return
 }
@@ -561,7 +561,7 @@ func (c *Clique) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *
 	txs []*types.Transaction, uncles []*types.Header, receipts []*types.Receipt) (*types.Block, []*types.Receipt, error) {
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
 	var err error
-	header.Root, err = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
 	if err != nil {
 		return nil, nil, err
 	}
--- a/consensus/clique/clique.go
+++ b/consensus/clique/clique.go
@@ -550,7 +550,7 @@ func (c *Clique) Prepare(chain consensus.ChainHeaderReader, header *types.Header
 func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs *[]*types.Transaction, uncles []*types.Header,
 	receipts *[]*types.Receipt, _ *[]*types.Transaction, _ *uint64) (err error) {
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
-	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	header.Root, err = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
 	header.UncleHash = types.CalcUncleHash(nil)
 	return
 }
@@ -560,7 +560,11 @@ func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 func (c *Clique) FinalizeAndAssemble(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB,
 	txs []*types.Transaction, uncles []*types.Header, receipts []*types.Receipt) (*types.Block, []*types.Receipt, error) {
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
-	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	var err error
+	header.Root, err = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	if err != nil {
+		return nil, nil, err
+	}
 	header.UncleHash = types.CalcUncleHash(nil)
 
 	// Assemble and return the final block for sealing
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -58,7 +58,7 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 		return ErrKnownBlock
 	}
 	if v.bc.isCachedBadBlock(block) {
-		return errStateRootVerificationFailed
+		return ErrKnownBadBlock
 	}
 	// Header validity is known at this point, check the uncles and transactions
 	header := block.Header()
@@ -138,8 +138,8 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
 	}
 	if !skipHeavyVerify {
 		validateFuns = append(validateFuns, func() error {
-			if root := statedb.IntermediateRoot(v.config.IsEIP158(header.Number)); header.Root != root {
-				return fmt.Errorf("invalid merkle root (remote: %x local: %x)", header.Root, root)
+			if root, err := statedb.IntermediateRoot(v.config.IsEIP158(header.Number)); header.Root != root || err != nil {
+				return fmt.Errorf("invalid merkle root (remote: %x local: %x), err %v", header.Root, root, err)
 			} else {
 				return nil
 			}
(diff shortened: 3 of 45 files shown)


KeefeL:
What is the behavior if S4 finished earlier than V3, should block5 P5 waits until V3 finished ?

consensus/clique/clique.go
```diff
@@ -550,7 +550,7 @@ func (c *Clique) Prepare(chain consensus.ChainHeaderReader, header *types.Header
 func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs *[]*types.Transaction, uncles []*types.Header,
 	receipts *[]*types.Receipt, _ *[]*types.Transaction, _ *uint64) (err error) {
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
-	header.Root = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
+	header.Root, err = state.IntermediateRoot(chain.Config().IsEIP158(header.Number))
```
KeefeL, line 553: It seems this err can be ignored.

core/blockchain.go
```diff
@@ -573,6 +586,23 @@ func (bc *BlockChain) SetHead(head uint64) error {
...
+	snaps := bc.snaps
+	// Verified and Result is false
+	if snaps != nil && snaps.Snapshot(block.Root()) != nil &&
+		snaps.Snapshot(block.Root()).Verified() && !snaps.Snapshot(block.Root()).WaitVerified() {
+		// Rewind by one block
+		log.Warn("current block verified failed, rewind to its parent", "height", block.NumberU64(), "hash", block.Hash())
+		bc.futureBlocks.Remove(block.Hash())
+		bc.badBlockCache.Add(block.Hash(), time.Now())
+		bc.reportBlock(block, nil, errStateRootVerificationFailed)
+		bc.setHeadBeyondRoot(block.NumberU64()-1, common.Hash{})
```
KeefeL, line 602: Shall we rewind cached difflayer as well?

core/state/statedb.go
```diff
@@ -154,25 +160,33 @@ func newStateDB(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB,
...
 	}
+
+	snapVerified := sdb.snap != nil && sdb.snap.Verified()
+	tr, err := db.OpenTrie(root)
+	// return error when 1. failed to open trie and 2. the snap is not nil and done verification
+	if err != nil && (sdb.snap == nil || snapVerified) {
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("no available state")
```
KeefeL, line 178: Won't reach here at anytime, I didn't get the point here?

core/state/statedb.go
```diff
@@ -956,29 +989,21 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
...
-	//
-	// This is weird pre-byzantium since the first tx runs with a prefetcher and
-	// the remainder without, but pre-byzantium even the initial prefetcher is
-	// useless, so no sleep lost.
-	prefetcher := s.prefetcher
-	if s.prefetcher != nil {
-		defer func() {
-			s.prefetcher.close()
-			s.prefetcher = nil
-		}()
```
KeefeL, line 979: Should s.StopPrefetcher() here.

KeefeL (commented):
LGTM

consensus/clique/clique.go
guagualvcha: The error will be return, because the function signature is 
```func (c *Clique) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state *state.StateDB, txs *[]*types.Transaction, uncles []*types.Header,
	receipts *[]*types.Receipt, _ *[]*types.Transaction, _ *uint64) (err error) 
```
The return value get a name `err`

core/blockchain.go
guagualvcha: You are right, fixed.

core/state/statedb.go
guagualvcha: you are right, fixed

consensus/clique/clique.go
KeefeL: ok

KeefeL (approved):

eth/tracers/api.go
```diff
@@ -556,7 +556,11 @@ func (api *API) IntermediateRoots(ctx context.Context, hash common.Hash, config
 		}
 		// calling IntermediateRoot will internally call Finalize on the state
 		// so any modifications are written to the trie
-		roots = append(roots, statedb.IntermediateRoot(deleteEmptyObjects))
+		root, err := statedb.IntermediateRoot(deleteEmptyObjects)
+		if err != nil {
+			return roots, err
```
yutianwu, line 561: why not return `nil, err` here

core/state/statedb.go
```diff
@@ -154,25 +160,30 @@ func newStateDB(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB,
...
 			sdb.snapDestructs = make(map[common.Address]struct{})
 			sdb.snapAccounts = make(map[common.Address][]byte)
 			sdb.snapStorage = make(map[common.Address]map[string][]byte)
 		}
 	}
+
+	snapVerified := sdb.snap != nil && sdb.snap.Verified()
+	tr, err := db.OpenTrie(root)
+	// return error when 1. failed to open trie and 2. the snap is not nil and done verification
+	if err != nil && (sdb.snap == nil || snapVerified) {
```
yutianwu, line 174: according to the comment, should the condition here be `sdb.snap != nil && snapVerified`

core/chain_makers.go
```diff
@@ -254,9 +254,9 @@ func makeHeader(chain consensus.ChainReader, parent *types.Block, state *state.S
 	} else {
 		time = parent.Time() + 10 // block time is fixed at 10 seconds
 	}
-
+	root, _ := state.IntermediateRoot(chain.Config().IsEIP158(parent.Number()))
```
yutianwu, line 257: do we need to check error here

core/blockchain.go
```diff
@@ -573,6 +586,25 @@ func (bc *BlockChain) SetHead(head uint64) error {
...
 }
 
+func (bc *BlockChain) tryRewindBadBlocks() {
+	bc.chainmu.Lock()
+	defer bc.chainmu.Unlock()
+	block := bc.CurrentBlock()
+	snaps := bc.snaps
+	// Verified and Result is false
+	if snaps != nil && snaps.Snapshot(block.Root()) != nil &&
+		snaps.Snapshot(block.Root()).Verified() && !snaps.Snapshot(block.Root()).WaitVerified() {
```
yutianwu, line 596: It costs me a while to figure out the difference between `Verified` and `WaitVerified` at the first glance. Since `Verified` returns the status of the verified process and `WaitVerified` returns the result of the verified process. But they share the same word `Verified`, can we have better names for these functions.

eth/tracers/api.go
guagualvcha: fixed

core/state/statedb.go
guagualvcha: if `sdb.snap == nil`,  we won't allow open trie error.

core/chain_makers.go
guagualvcha: I think no need. `chain_makers.go` is usually for testing.

core/blockchain.go
guagualvcha: sure, changed. Please check latest commit.

yutianwu (approved):

j75689 (approved):

core/state/statedb.go
brilliant-lx: "err != nil" means OpenTrie failed, newStateDB should fail as well?
newStateDB will success now, if OpenTrie fail when snap!=nil and snap verify failed.
It is make sense?

core/state/statedb.go
```diff
@@ -154,25 +160,30 @@ func newStateDB(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB,
...
+	sdb.trie = tr
 	return sdb, nil
 }
 
 // StartPrefetcher initializes a new trie prefetcher to pull in nodes from the
 // state trie concurrently while the state is mutated so that when we reach the
 // commit phase, most of the needed data is already hot.
 func (s *StateDB) StartPrefetcher(namespace string) {
+	s.prefetcherLock.Lock()
+	defer s.prefetcherLock.Unlock()
```
brilliant-lx, line 186: This PR adds a new goroutine to for pipeline commit?
Prefetcher will be called both on commit and verify routine now?
It is a bit complicated.

core/state/trie_prefetcher.go
```diff
@@ -257,9 +256,7 @@ func newSubfetcher(db Database, root common.Hash, accountHash common.Hash) *subf
 		seen:        make(map[string]struct{}),
 		accountHash: accountHash,
 	}
-	gopool.Submit(func() {
-		sf.loop()
```
brilliant-lx, line 261: why this change? irrelevant with pipe commit?

core/state/statedb.go
```diff
@@ -956,29 +986,21 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 // IntermediateRoot computes the current root hash of the state trie.
 // It is called in between transactions to get the root hash that
 // goes into transaction receipts.
-func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) common.Hash {
+func (s *StateDB) IntermediateRoot(deleteEmptyObjects bool) (common.Hash, error) {
```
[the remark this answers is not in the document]
guagualvcha: When we do `IntermediateRoot`, I am not sure wether the parent MPT root verification is correct, we need return error if the parent verification failed.

core/state/statedb.go
guagualvcha:
```
"err != nil" means OpenTrie failed, newStateDB should fail as well?
newStateDB will success now, if OpenTrie fail when snap!=nil and snap verify failed.
It is make sense?
```
We allow `OpenTrie` failed for `newStateDB`, since Trie is not the necessary condition for block process.

guagualvcha:
```
This PR adds a new goroutine to for pipeline commit?
Prefetcher will be called both on commit and verify routine now?
It is a bit complicated.
```
Prefetcher is only called when processing the block(apply transactions)

core/state/trie_prefetcher.go
guagualvcha: Yes, a bit overhead with `gopool.Submit`

core/block_validator.go
```diff
@@ -129,13 +135,15 @@ func (v *BlockValidator) ValidateState(block *types.Block, statedb *state.StateD
...
 		},
-		func() error {
-			if root := statedb.IntermediateRoot(v.config.IsEIP158(header.Number)); header.Root != root {
-				return fmt.Errorf("invalid merkle root (remote: %x local: %x)", header.Root, root)
+	}
+	if !skipHeavyVerify {
+		validateFuns = append(validateFuns, func() error {
+			if root, err := statedb.IntermediateRoot(v.config.IsEIP158(header.Number)); header.Root != root || err != nil {
+				return fmt.Errorf("invalid merkle root (remote: %x local: %x), err %v", header.Root, root, err)
 			} else {
```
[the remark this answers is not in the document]
guagualvcha: fixed

core/state/statedb.go
brilliant-lx: > "err != nil" means OpenTrie failed, newStateDB should fail as well? newStateDB will success now, if OpenTrie fail when snap!=nil and snap verify failed. It is make sense?
Got it, the behavior changed.
With pipeline commit, trie tree may not ready since IntermediateRoot() will be executed asynchronously.
OpenTrie is not necessary and a failed OpenTrie is probably not matter, since state verify is probably to success.

cmd/utils/flags.go
```diff
@@ -127,6 +127,10 @@ var (
 		Usage: "Enable diffy sync, Please note that enable diffsync will improve the syncing speed, " +
 			"but will degrade the security to light client level",
 	}
+	PipeCommitFlag = cli.BoolFlag{
+		Name:  "pipecommit",
+		Usage: "Enable MPT pipeline commit, it will improve syncing performance. It is an experimental feature",
```
guagualvcha, line 132: Add default option.

motobrowning (commented):
K

Jemerald83 (approved):

PR #673, opened 2021-12-18
fixed direct db operation in trie database

nikandfor:
### Description

fixed direct diskdb read in node method

### Rationale

made it similar to Node method and others.

### Example

...

### Changes

Notable changes: 
* changed db.Get(hash[:]) to rawdb.ReadTrieNode(db, hash)

### Preflight checks

### Already reviewed by

...

### Related issues

...

### DIFF
--- a/trie/database.go
+++ b/trie/database.go
@@ -396,8 +396,8 @@ func (db *Database) node(hash common.Hash) node {
 	memcacheDirtyMissMeter.Mark(1)
 
 	// Content unavailable in memory, attempt to retrieve from disk
-	enc, err := db.diskdb.Get(hash[:])
-	if err != nil || enc == nil {
+	enc := rawdb.ReadTrieNode(db.diskdb, hash)
+	if len(enc) == 0 {
 		return nil
 	}
 	if db.cleans != nil {


guagualvcha:
Is there any difference between current implementation?

nikandfor:
> Is there any difference between current implementation?

It equivalent for now but it's a time bomb if something changed.

guagualvcha:
> > Is there any difference between current implementation?
> 
> It is equivalent for now but it's a time bomb if something changed.
> 
> EDIT: I mean I made it more similar and following the same rules and style.

sorry , I can not quite get this, so what is the motivation for this PR?

nikandfor:
> sorry , I can not quite get this, so what is the motivation for this PR?

There is unified interface to work with database. It's defined in [rawdb package](https://github.com/binance-chain/bsc/blob/master/core/rawdb/schema.go). This package incapsulates how data types are stored in database and especially how byte keys are built. Most of the code in the repo uses this interface including the following method in [40 lines below](https://github.com/binance-chain/bsc/pull/673/files#diff-8348a172e9fd3d3eb93c445d4ca58b8753b0f6d626c7af7db3b30b820d0788daR439).

My motivation is to prevent possible bugs in case something about database will change. My way to do that is to make db usage more uniform. This is a kind of tiny refactoring.

laizy (approved):

PR #704, opened 2022-01-07
[R4R]prefetch state by applying the transactions within one block

guagualvcha:
### Description
By improving the snapshot clean cache hit rate, we can improve the capacity of BSC further.

### Rationale
According to the test, the cache hitting rate on BSC is about 94%, while the miss 6% IO will cost 32% of execution time.

We try to preload the state from DB by applying the transactions in advance. It turns out the execution  time was reduced by 30%-40%  

image

### Example
Verified on testnet and mainnet.

### Changes

No impact to users

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2065,10 +2065,10 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 		//Process block using the parent state as reference point
 		substart := time.Now()
 		statedb, receipts, logs, usedGas, err := bc.processor.Process(block, statedb, bc.vmConfig)
+		atomic.StoreUint32(&followupInterrupt, 1)
 		activeState = statedb
 		if err != nil {
 			bc.reportBlock(block, receipts, err)
-			atomic.StoreUint32(&followupInterrupt, 1)
 			return it.index, err
 		}
 		// Update the metrics touched during block processing
@@ -2087,7 +2087,6 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 			if err := bc.validator.ValidateState(block, statedb, receipts, usedGas); err != nil {
 				log.Error("validate state failed", "error", err)
 				bc.reportBlock(block, receipts, err)
-				atomic.StoreUint32(&followupInterrupt, 1)
 				return it.index, err
 			}
 		}
@@ -2105,7 +2104,6 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 		substart = time.Now()
 		status, err := bc.writeBlockWithState(block, receipts, logs, statedb, false)
 		if err != nil {
-			atomic.StoreUint32(&followupInterrupt, 1)
 			return it.index, err
 		}
 		// Update the metrics touched during block commit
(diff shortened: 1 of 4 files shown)


core/state_prefetcher.go
```diff
@@ -35,42 +36,58 @@ type statePrefetcher struct {
...
-		statedb.Prepare(tx.Hash(), block.Hash(), i)
-		if err := precacheTransaction(msg, p.config, gaspool, statedb, header, evm); err != nil {
-			return // Ugh, something went horribly wrong, bail out
-		}
-		// If we're pre-byzantium, pre-load trie nodes for the intermediate root
-		if !byzantium {
-			statedb.IntermediateRoot(true)
+	transactions := block.Transactions()
+	threads := runtime.NumCPU()
+	batch := len(transactions) / (threads + 1)
```
yutianwu, line 58: what if `len(transactions)` < `(threads + 1)`, do we need to check it

core/blockchain.go
```diff
@@ -2051,13 +2054,21 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 
 		// Enable prefetching to pull in trie node paths while processing transactions
 		statedb.StartPrefetcher("chain")
-
+		var followupInterrupt uint32
+		// For diff sync, it may fallback to full sync, so we still do prefetch
+		if len(block.Transactions()) >= prefetchTxNumber {
+			throwaway, _ := state.New(parent.Root, bc.stateCache, bc.snaps)
```
yutianwu, line 2060: do we need to swallow the error here

guagualvcha: actually, it can must be no error, because `statedb` has been initilized in the same way. But I will protect here now.

core/state_prefetcher.go
guagualvcha: emm, because if the len(transactions) <=100, I will not start prefetcher if you check the code. But since you mention this, I will do double-protect here. Thanks

yutianwu (approved):

core/blockchain.go
```diff
@@ -2051,13 +2054,21 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 
 		// Enable prefetching to pull in trie node paths while processing transactions
 		statedb.StartPrefetcher("chain")
-
+		var followupInterrupt uint32
```
j75689, line 2057: Should we set `followupInterrupt` to 1 before the next loop starts? To avoid doing anything tricky to cause prefetching to be slower than block execution.

guagualvcha: good idea.

guagualvcha: please check the latest commit

j75689 (approved):

core/state_prefetcher.go
```diff
@@ -35,42 +36,60 @@ type statePrefetcher struct {
 	engine consensus.Engine    // Consensus engine used for block rewards
 }
 
+// newStatePrefetcher initialises a new statePrefetcher.
```
KeefeL, line 39: Should be NewStatePrefetcher

KeefeL (approved):

core/state_prefetcher.go
guagualvcha: fixed

PR #728, opened 2022-01-21
[R4R]Separate Processing and State Verification on BSC: implement trust protocol and verify node

KeefeL:
## Motivation

Currently a BSC node has two kinds of state world: MPT and snapshot. MPT(Merkle Patricia Tries) is a tree-structured state world. The key function of MPT is to generate the state root to ensure state consistency, while the query/commit on MPT is quite slow. Snapshot is a flattened key-value-based state world. Snapshot provides fast queries and commits. The storage size of the snapshot increases slowly even with a large transaction volume. Snapshot is usually used for block processing, while MPT is used for state verification.

In order to lower the hardware requirement and keep security, we introduce two type nodes to make full use of different storages, one is named **fast node**, and the other is named **verify node**. The **fast node** will do block processing with snapshot, it will do all verification against blocks except state root. The **verify node** receives  diffhash from the fast node then responds MPT root to the **fast node**. 

The **fast node** doesn’t need to store MPT, so the storage and computation requirement will be much lower.

### Architecture
The network topology of the **fast node** and **verify node**:
topology

**Fast node** is a bsc client that does fullsync using only Snapshot and generates difflayer. It needs the confirm message from the **verify node** before freezing the blocks, it has to wait until it receives a confirm message from the **verify node** for the ancestor block before inserting new blocks.

**Verify node** is a normal bsc full node that does fullsync using Snapshot and MPT, generates difflayer. It receives diffhash from **fast nodes**, finds corresponding difflayer whose diffhash is matched, then it responds MPT root message to the **fast node**. 

All the messages exchanged between **fast node** and **verify node** are based on the **trust p2p protocol**.

The relationship between the **fast node** and the **verify node**:
architecture

On the **fast node** side, for each inserted block, verify manager creates a verify task, the verify task submits the block’s diffhash to multiple trusted **verify nodes**, and receives response from them, once received the confirm message, the block has been verified.

On the **verify node** side, it receives requests from **fast nodes**, does corresponding processing and then responds to **fast nodes** immediately. The processing of diffhash finds difflayer from cache first, if not found then search from diff store, if still no difflayer found, then find from blockchain headers, this processing would be very quick. 

### Authentication
**Fast node** can only rely on trusted **verify node**, either deployed by the same developer or deployed by a trusted organization. We know the different peers will verify the peer id during the handshake, we will borrow this mechanism to do the authentication. We introduce **VerifyNodes** settings, which is a list of encoded addresses, the **fast nodes** only build connections(based on trust protocol) with **VerifyNodes**.

### Chain Tools

### Prototype verification

#### Fast Node
Introduce a new **tries-verify-mode** setting, there are four modes:
- **local**: a normal full node;
- **full**: state verification by verify node, full secure mode.
- **light**: state verification by verify node, light secure mode.
- **insecure**: no state verification.

If the fast node runs in not local mode, the node will disable diff protocol by default, If the fast node runs in full or light mode, the node will enable trust protocol by default.

` ./geth --config ./config.toml --datadir ./node --syncmode full --cache 5000 --tries-verify-mode light`

#### Verify node
When a full node has enabled the trust protocol, it can serve as a verify node, at the same time, we will recommend you to enable persist diff, disable snap protocol and diff protocol when running a verify node.

` ./geth --config ./config.toml --datadir ./node --syncmode full --cache 5000 --persistdiff --enabletrustprotocol --disablesnapprotocol --disablediffprotocol`

#### Prune tries node


core/blockchain.go
```diff
@@ -1679,6 +1689,12 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		diffLayer.Receipts = receipts
 		diffLayer.BlockHash = block.Hash()
 		diffLayer.Number = block.NumberU64()
+
+		sort.Sort(types.DiffCodeSlice(diffLayer.Codes))
+		sort.Sort(common.AddressSlice(diffLayer.Destructs))
+		sort.Sort(types.DiffAccountSlice(diffLayer.Accounts))
+		sort.Sort(types.DiffStorageSlice(diffLayer.Storages))
+
```
guagualvcha, line 1697: Since `sort` is a heavy operation, can we move this to another routine, so that it doesn't have to block the processing of blockchain?

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
 		return chain
 	}
 }
+
+func (bc *BlockChain) GetRootByDiffHash(blockNumber uint64, blockHash common.Hash, diffHash common.Hash) (*types.VerifyResult, error) {
```
guagualvcha, line 3038: It is not proper to return a `*types.VerifyResult` while the function name is `GetRootByDiffHash`. Try better name?

guagualvcha: `VerifyResult` is a too common name in `types` package, need better name as well.

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+	var res types.VerifyResult
+	res.BlockNumber = blockNumber
+	res.BlockHash = blockHash
+
+	if blockNumber > bc.CurrentHeader().Number.Uint64()+11 {
```
guagualvcha, line 3043: `11` is a magic number here, lets name it.

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+		return &res, nil
+	} else if blockNumber > bc.CurrentHeader().Number.Uint64() {
+		res.Status = types.StatusBlockNewer
+		return &res, nil
+	}
+
+	diff := bc.GetTrustedDiffLayer(blockHash)
+	if diff != nil {
+		if diff.DiffHash == (common.Hash{}) {
+			hash, err := GetTrustedDiffHash(diff)
```
guagualvcha, line 3054: Rename `GetTrustedDiffHash` to `CalDiffHash`?

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+			if err != nil {
+				res.Status = types.StatusUnexpectedError
+				return &res, err
+			}
+
+			diff.DiffHash = hash
```
guagualvcha, line 3060: there is a race condition on `diff.DiffHash`,  concurrently read and write on `DiffHash`. Try using `atomic` to avoid the race condition.

core/blockchain.go
guagualvcha, line 3038: I am not sure wether this func should return `error`, because if return error, the connection with peers will close, is it what we expected?

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+		}
+
+		if diffHash != diff.DiffHash {
+			res.Status = types.StatusDiffHashMismatch
+			return &res, nil
+		}
+
+		header := bc.GetHeaderByHash(blockHash)
```
guagualvcha, line 3068: `GetHeaderByHash` shows up twice within this function twice, it is a bit duplicated, and I think check whether the header exist or not should happen before `GetTrustedDiffLayer`(who is heavy). We can simplify it.

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+		if blockNumber > bc.CurrentHeader().Number.Uint64()-11 {
+			res.Status = types.StatusPossibleFork
+			return &res, nil
+		}
+
+		res.Status = types.StatusImpossibleFork
+		return &res, nil
+	}
+
+	res.Status = types.StatusUntrustedVerified
```
guagualvcha, line 3089: `StatusUntrustedVerified` may be a bit misleading, what about`StatusPartiallyVerified`?

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+			}
+		}
+
+		if _, err := ApplyMessage(vmenv, msg, new(GasPool).AddGas(tx.Gas())); err != nil {
+			return nil, fmt.Errorf("transaction %#x failed: %v", tx.Hash(), err)
+		}
+		statedb.Finalise(vmenv.ChainConfig().IsEIP158(block.Number()))
+	}
+
+	return statedb.GenerateDiffLayer(), nil
```
guagualvcha, line 3151: Is this been tested? Since `IntermediateRoot` is required before  `GenerateDiffLayer`.

core/blockchain.go
```diff
@@ -3015,3 +3034,155 @@ func EnablePersistDiff(limit uint64) BlockChainOption {
...
+
+	hasher := sha3.NewLegacyKeccak256()
+	_, err = hasher.Write(rawData)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("hasher write error: %v", err)
+	}
+
+	var hash common.Hash
+	hasher.Sum(hash[:0])
+	return hash, nil
```
guagualvcha, line 3187: We can assign the `DiffHash` of `DiffLayer` within this function.

core/types/block.go
```diff
@@ -40,6 +40,40 @@ var (
 	EmptyUncleHash = rlpHash([]*Header(nil))
 )
 
+type VerifyStatus struct {
```
guagualvcha, line 43: I think it is better to create a new golang file than reuse `block.go`.

core/types/block.go
```diff
@@ -40,6 +40,40 @@ var (
...
+	StatusBlockTooNew    = VerifyStatus{Code: 0x301, Msg: "can’t verify because of block number larger than current height more than 11"}
+	StatusBlockNewer     = VerifyStatus{Code: 0x302, Msg: "can’t verify because of block number larger than current height"}
+	StatusPossibleFork   = VerifyStatus{Code: 0x303, Msg: "can’t verify because of possible fork detected"}
+	StatusRequestTooBusy = VerifyStatus{Code: 0x304, Msg: "can’t verify because of request too busy"}
+
+	// StatusUnexpectedError is unexpected internal error.
+	StatusUnexpectedError = VerifyStatus{Code: 0x400, Msg: "can’t verify because of unexpected internal error"}
+)
+
+type VerifyResult struct {
```
guagualvcha, line 70: Try better name of `VerifyResult` and `VerifyStatus`

eth/backend.go
```diff
@@ -550,6 +553,7 @@ func (s *Ethereum) Protocols() []p2p.Protocol {
 	}
 	// diff protocol can still open without snap protocol
 	protos = append(protos, diff.MakeProtocols((*diffHandler)(s.handler), s.snapDialCandidates)...)
+	protos = append(protos, trust.MakeProtocols((*trustHandler)(s.handler), s.snapDialCandidates)...)
```
guagualvcha, line 556: I made a mistake when implement diff sync, I think need a  `tustDialCandidates` here

eth/handler_trust.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// Handle is invoked from a peer's message handler when it receives a new remote
+// message that the handler couldn't consume and serve itself.
+func (h *trustHandler) Handle(peer *trust.Peer, packet trust.Packet) error {
+	switch packet := packet.(type) {
+	case *trust.RootResponsePacket:
+		// TODO: h.bc.VerifyManager().HandleRootResponse(peer.ID(), *packet)
+		return nil
+
```
guagualvcha, line 39: I think this PR needs to implement the handle of `RootRequestPacket`?

eth/peerset.go
```diff
@@ -148,6 +159,40 @@ func (ps *peerSet) registerDiffExtension(peer *diff.Peer) error {
...
+// being until the `eth` main protocol starts looking for it.
+func (ps *peerSet) registerTrustExtension(peer *trust.Peer) error {
+	// Reject the peer if it advertises `trust` without `eth` as `trust` is only a
+	// satellite protocol meaningful with the chain selection of `eth`
+	if !peer.RunningCap(eth.ProtocolName, eth.ProtocolVersions) {
+		return errTrustWithoutEth
+	}
+	// If the peer isn't verify node, don't register trust extension into eth protocol.
+	if !peer.VerifyNode() {
+		return nil
```
guagualvcha, line 173: Do we need to:
```
delete(ps.trustWait, id)
wait <- peer
```

So that the eth protocol won't wait any more.

eth/peerset.go
```diff
@@ -234,16 +279,74 @@ func (ps *peerSet) waitDiffExtension(peer *eth.Peer) (*diff.Peer, error) {
...
 func (ps *peerSet) GetDiffPeer(pid string) downloader.IDiffPeer {
 	if p := ps.peer(pid); p != nil && p.diffExt != nil {
 		return p.diffExt
 	}
 	return nil
 }
 
+// GetVerifyPeers returns an array of verify nodes.
+func (ps *peerSet) GetVerifyPeers() []*trustPeer {
+	res := make([]*trustPeer, 0)
```
guagualvcha, line 338: Race condition on `peerSet.peers`, try avoid that.

ethclient/ethclient.go
```diff
@@ -200,6 +200,12 @@ func (ec *Client) GetDiffAccountsWithScope(ctx context.Context, number *big.Int,
 	return &result, err
 }
 
+func (ec *Client) GetRootByDiffHash(ctx context.Context, blockNr *big.Int, blockHash common.Hash, diffHash common.Hash) (*types.VerifyResult, error) {
```
guagualvcha, line 203: Try come up with better name.

core/blockchain.go
KeefeL: Yes, you are right, I have missed this, I think we just need intermediate account root, state root is not needed.

KeefeL: Updated

core/blockchain.go
KeefeL: Sure, I will move these sort functions into cacheDiffLayer function, and add a sorted flag, if need sort, create a go routine to call cacheDiffLayer, otherwise call directly.

core/blockchain.go
KeefeL: sure, replace by maxDiffForkDist

core/blockchain.go
KeefeL: Ok

core/blockchain.go
KeefeL: I think better not, as this function is not a method of DiffLayer, and the DiffHash have multiple use cases, I will recommend this function only compute the hash, the caller should decide whether to record it or not.

core/blockchain.go
KeefeL: replace by atomic.value

core/blockchain.go
KeefeL: Yes, since the fast node receive error msg, it will add the verify node into bad peer, this connection is expected to close.

core/blockchain.go
KeefeL: You are right, I will move GetHeaderByHash before GetTrustedDiffLayer.

core/blockchain.go
KeefeL: Good idea.

eth/handler_trust.go
KeefeL: Yes, we have implemented the HandleRootRequest in the trust/handler.go

eth/peerset.go
KeefeL: Look the waitTrustExtension function, if the peer isn't a verify node, eth protocol won't wait, so I think it is unnecessary to do what you mentioned here.

eth/peerset.go
KeefeL: You are right, I will add rlock here.

eth/backend.go
KeefeL: You are right, I will modify it.

core/blockchain.go
KeefeL: Actually, I plan move the verify related structure into the verify_manager.go after fast node has completed.

core/types/block.go
KeefeL: You are right, we plan to move these definition into verify manager related file, here is temporary solution.

core/blockchain.go
KeefeL: updated

core/blockchain.go
guagualvcha: Maybe `maxForkDist` is better, so other place can use it too.

core/blockchain.go
guagualvcha: As we discussed, bad peer is within task, so no need to close the connection.  Please do not return error here.

core/blockchain.go
guagualvcha: ok

core/blockchain.go
KeefeL: maxDiffForkDist has been defined in the core package, it seems use maxDiffForkDist is ok.

eth/peerset.go
guagualvcha: cool

guagualvcha (approved):

core/blockchain.go
KeefeL: Updated.

KeefeL:
We are going to merge this PR first, and move on fast node, then submit PR to develop branch.

PR #768, opened 2022-02-15
[WIP] Collect performance metrics from different aspects

forcodedancing:
### Description

Get different metrics for performance analysis. These metrics includes:

* Main Process - mining & block importing
* Top Addresses - top external owned addresses and contracts
* Cache & Prefetcher
(1) add hit & miss count on cache layers on L1, the L2-L3 already have this metrics
(2) delays on each cache layers, include disk layer
(3) accumulated account of hit number on each layers , include disk layer
(4) accumulated io cost of one block when importing block or mining
* Config for collecting mutex 
* Config for collecting gotrace

### Rationale

### Example

### Changes

Notable changes: 
* ...


perf/top_address.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func UpdateTopContractStats(address common.Address) {
+	if !addrStatsEnabled {
+		return
+	}
+	inc(contractMap, address.Hex())
+}
+
+func inc(m map[string]int64, address string) {
```
guagualvcha, line 40: I think we should limit the size of the map, if we want top1K, the map should limit at 10K size. Otherwise, OOM will happen.

perf/top_address.go
```diff
@@ -0,0 +1,113 @@
...
+func getTopContract(topN int) AddrTxStats {
+	topM := top(contractMap, topN)
+	return AddrTxStats{
+		Map:   topM,
+		Start: start,
+		End:   time.Now(),
+	}
+}
+
+func top(m map[string]int64, topN int) (topM map[string]int64) {
```
guagualvcha, line 66: I think there is a race condition on map `m`, since another routine may add item into map `m`, while this func is doing iterate over it.

perf/top_address.go
```diff
@@ -0,0 +1,113 @@
...
+		addr, count := q.Pop()
+		if addr != nil {
+			topM[addr.(string)] = count
+			counter++
+		}
+	}
+	return topM
+}
+
+type getTop func(int) AddrTxStats
```
guagualvcha, line 83: can we sort the result when doing `getTop`, I mean I can quickly find exactly know which contract is the mostly called.

perf/go_trace.go
```diff
@@ -0,0 +1,82 @@
...
+
+func StartGoTraceMetrics() {
+	if !traceMetricsEnabled {
+		return
+	}
+	startTrace()
+	go func() {
+		router := gin.Default()
+		router.GET("/trace/start", func(c *gin.Context) {
+			startTrace()
```
guagualvcha, line 28: Shall we pass a trace duration? After that the node will stop trace automatically, so we dont have to call the `stop` again,

miner/worker.go
```diff
@@ -980,8 +995,15 @@ func (w *worker) commitNewWork(interrupt *int32, noempty bool, timestamp int64)
...
 		if len(remoteTxs) > 0 {
+			startOrder := time.Now()
 			txs := types.NewTransactionsByPriceAndNonce(w.current.signer, remoteTxs)
-			if w.commitTransactions(txs, w.coinbase, interrupt) {
+			//only tracking remote txs, for our validators has no local txs in general
+			perf.RecordMPMetrics(perf.MpMiningOrder, startOrder)
+
+			startCommit := time.Now()
+			succeed := w.commitTransactions(txs, w.coinbase, interrupt)
+			perf.RecordMPMetrics(perf.MpMiningCommit, startCommit)
```
guagualvcha, line 1005: Can we rename to `MpMiningCommitTx`?

eth/protocols/eth/peer.go
```diff
@@ -348,10 +350,15 @@ func (p *Peer) SendNewBlock(block *types.Block, td *big.Int) error {
...
-	return p2p.Send(p.rw, NewBlockMsg, &NewBlockPacket{
+
+	start := time.Now()
+	err := p2p.Send(p.rw, NewBlockMsg, &NewBlockPacket{
 		Block: block,
 		TD:    td,
 	})
+	perf.RecordMPMetrics(perf.MpPropagationSend, start)
+
+	return err
```
guagualvcha, line 361: It is a bit tricky to get the propagation time, since we need record the sending time in one peer, and record another receiving time in other peers.

core/state_transition.go
```diff
@@ -276,6 +277,10 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	}
 	st.refundGas()
 
+	//track top EOA/contract addresses
+	perf.UpdateTopEOAStats(sender.Address())
+	perf.UpdateTopContractStats(st.to())
```
guagualvcha, line 282: Only do when `st.to` is not nil.

perf/top_address.go
forcodedancing: updated, use a ticker to clean the maps

perf/top_address.go
forcodedancing: added lock

perf/go_trace.go
forcodedancing: added duration

miner/worker.go
forcodedancing: renamed

core/state_transition.go
forcodedancing: done

eth/protocols/eth/peer.go
forcodedancing: added logs and fetchHeader/fetchBodies, not sure whether it is correct or not

brilliant-lx (commented):
LGTM.

brilliant-lx (approved):

PR #772, opened 2022-02-16
Cache metrics

flywukong:
### Description

Get different metrics for cache performance analysis.
(1)  hit & miss count on each cache layers (L1-L3)
(2) delays on each cache layers(include disk layer)
(3) accumulated account of hit number on each layers(include disk layer)
(4)  accumulated io cost of one block when importing block or mining 

### Rationale

### Example

### Changes

Notable changes: 
* ...


core/state/state_object.go
```diff
@@ -247,7 +273,7 @@ func (s *StateObject) GetCommittedState(db Database, key common.Hash) common.Has
 			*meter += time.Since(readStart)
 			readStart = time.Now()
 		}
-		if metrics.EnabledExpensive {
+		if metrics.EnableIORecord {
```
forcodedancing, line 276: Whey we change the flag? It will be difficult for others to use our metrics. They need to resolve such conficits.

core/state/statedb.go
```diff
@@ -669,16 +678,25 @@ func (s *StateDB) preloadStateObject(address []common.Address) []*StateObject {
...
+
+	cacheL1AccountMissMeter.Mark(1)
 	// If no live objects are available, attempt to use snapshots
 	var (
 		data *Account
 		err  error
 	)
 	if s.snap != nil {
-		if metrics.EnabledExpensive {
+		if metrics.EnableIORecord {
```
forcodedancing, line 699: same as previous comment

miner/worker.go
```diff
@@ -81,7 +81,11 @@ const (
 )
 
 var (
-	commitTxsTimer = metrics.NewRegisteredTimer("worker/committxs", nil)
+	commitTxsTimer        = metrics.NewRegisteredTimer("worker/committxs", nil)
```
forcodedancing, line 84: Is it possible to move these to a separated file, as well as several other files? It will be easier to use when others want to apply our code.

core/state/state_object.go
flywukong: it is unuseful in lastest code

flywukong: I will delete this change later

cachemetrics/get_gid.go
```diff
@@ -0,0 +1,42 @@
...
+func IsSyncMainRoutineID(id int64) bool {
+	if id == SyncingRoutineId {
+		return true
+	}
+	return false
+}
+
+// judge if it is main process of mining
+func IsMinerMainRoutineID(id int64) bool {
+	if id == MiningRoutineId {
```
guagualvcha, line 38: If it do not cost too much, I think a read write lock is needed here, otherwise, there is race condition on `MiningRoutineId`

cmd/geth/config.go
```diff
@@ -203,6 +203,12 @@ func applyMetricConfig(ctx *cli.Context, cfg *gethConfig) {
 	if ctx.GlobalIsSet(utils.MetricsEnabledExpensiveFlag.Name) {
 		cfg.Metrics.EnabledExpensive = ctx.GlobalBool(utils.MetricsEnabledExpensiveFlag.Name)
 	}
+	if ctx.GlobalIsSet(utils.MetricsEnabledRecordIOFlag.Name) {
+		cfg.Metrics.EnabledExpensive = ctx.GlobalBool(utils.MetricsEnabledRecordIOFlag.Name)
```
guagualvcha, line 207: why reuse  `cfg.Metrics.EnabledExpensive`, since it will introduce other expensive metrics other than IO metrics?

core/state/snapshot/disklayer.go
```diff
@@ -107,17 +109,63 @@ func (dl *diskLayer) AccountRLP(hash common.Hash) ([]byte, error) {
...
+			if hitInL3 {
+				syncL3AccountHitMeter.Mark(1)
+				cachemetrics.RecordCacheDepth("CACHE_L3_ACCOUNT")
+				cachemetrics.RecordCacheMetrics("CACHE_L3_ACCOUNT", start)
+				cachemetrics.RecordTotalCosts("CACHE_L3_ACCOUNT", start)
+			}
+			if hitInDisk {
+				syncL3AccountMissMeter.Mark(1)
+				cachemetrics.RecordCacheDepth("DISK_L4_ACCOUNT")
+				cachemetrics.RecordCacheMetrics("DISK_L4_ACCOUNT", startGetInDisk)
```
guagualvcha, line 133: Little difference between`startGetInDisk` and `start`? Why use different ones here>

core/state/state_object.go
```diff
@@ -197,24 +208,48 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
...
+			cachemetrics.RecordCacheDepth("CACHE_L1_STORAGE")
+			cachemetrics.RecordCacheMetrics("CACHE_L1_STORAGE", start)
+			cachemetrics.RecordTotalCosts("CACHE_L1_STORAGE", start)
+		}
+
+		if isMinerMainProcess && hitInCache {
+			cachemetrics.RecordMinerCacheDepth("MINER_L1_STORAGE")
+			cachemetrics.RecordMinerCacheMetrics("MINER_L1_STORAGE", start)
+			cachemetrics.RecordMinerTotalCosts("MINER_L1_STORAGE", start)
+		}
```
guagualvcha, line 228: Can we record a total IO cost here? no matter hit or miss.

guagualvcha: And I think the `L1 cache` metrics should measure from:
```
// GetState retrieves a value from the account storage trie.
func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
	// If the fake storage is set, only lookup the state here(in the debugging mode)
	if s.fakeStorage != nil {
		return s.fakeStorage[key]
	}
	// If we have a dirty value for this state entry, return it
	value, dirty := s.dirtyStorage[key]
	if dirty {
		return value
	}
	// Otherwise return the entry's original value
	return s.GetCommittedState(db, key)
}
```

guagualvcha: I think `dirtyStorage` belongs to L1 cache too.

core/state/state_object.go
```diff
@@ -314,12 +349,30 @@ func (s *StateObject) setState(key, value common.Hash) {
...
 	for key, value := range s.dirtyStorage {
 		s.pendingStorage[key] = value
 		if value != s.originStorage[key] {
 			slotsToPrefetch = append(slotsToPrefetch, common.CopyBytes(key[:])) // Copy needed for closure
 		}
 	}
+	overheadCost = time.Since(start)
 	if s.db.prefetcher != nil && prefetch && len(slotsToPrefetch) > 0 && s.data.Root != emptyRoot {
 		s.db.prefetcher.prefetch(s.data.Root, slotsToPrefetch, s.addrHash)
 	}
```
guagualvcha, line 378: `overheadCost = time.Since(start)` should move after `s.db.prefetcher.prefetch`

core/state/statedb.go
```diff
@@ -604,9 +644,30 @@ func (s *StateDB) TryPreload(block *types.Block, signer types.Signer) {
...
+			minerPreloadCost.Update(overheadCost)
+			minerPreloadCounter.Inc(overheadCost.Nanoseconds())
+		}
+	}()
+
+	start := time.Now()
 	for account := range accounts {
 		accountsSlice = append(accountsSlice, account)
 	}
+	overheadCost = time.Since(start)
```
guagualvcha, line 669: Do you think we should move `	overheadCost = time.Since(start)` down?

cmd/geth/config.go
flywukong: it should be EnableIORecord，but it is not useful now , I has tried to record I/O cost of one block using EnableIORecord, but  then I found out that this is not necessary ， I will delete the relevant code

core/state/statedb.go
flywukong: updated

cmd/geth/config.go
flywukong: updated

core/state/state_object.go
flywukong: both Updated

core/state/statedb.go
flywukong: yes ,  it  should move down

flywukong: updated

core/state/state_object.go
flywukong: yes , it belongs to L1,  Updated

cachemetrics/get_gid.go
guagualvcha: `SyncingRoutineId` need to be atomic too.

cmd/geth/config.go
```diff
@@ -203,6 +203,7 @@ func applyMetricConfig(ctx *cli.Context, cfg *gethConfig) {
 	if ctx.GlobalIsSet(utils.MetricsEnabledExpensiveFlag.Name) {
 		cfg.Metrics.EnabledExpensive = ctx.GlobalBool(utils.MetricsEnabledExpensiveFlag.Name)
 	}
+
```
forcodedancing, line 206: revert this file

cmd/geth/main.go
```diff
@@ -204,6 +204,8 @@ var (
 	metricsFlags = []cli.Flag{
 		utils.MetricsEnabledFlag,
 		utils.MetricsEnabledExpensiveFlag,
+		utils.MetricsEnabledRecordIOFlag,
+		utils.MetricsDisablePrefetchFlag,
```
forcodedancing, line 208: `utils.MetricsDisablePrefetchFlag` useless now, please remove it

cmd/utils/flags.go
```diff
@@ -744,6 +744,14 @@ var (
 		Name:  "metrics.expensive",
 		Usage: "Enable expensive metrics collection and reporting",
 	}
+	MetricsDisablePrefetchFlag = cli.BoolFlag{
```
forcodedancing, line 747: remove `MetricsDisablePrefetchFlag`

core/state/snapshot/disklayer.go
```diff
@@ -133,6 +181,50 @@ func (dl *diskLayer) AccountRLP(hash common.Hash) ([]byte, error) {
...
+	hitInL3 := false
+	hitInDisk := false
+	var startGetInDisk time.Time
+	defer func() {
+		isSyncMainProcess := cachemetrics.IsSyncMainRoutineID(routeid)
+		isMinerMainProcess := cachemetrics.IsMinerMainRoutineID(routeid)
+		if isSyncMainProcess {
+			// layer 2 miss
+			syncL2StorageMissMeter.Mark(1)
+			if hitInL3 {
```
forcodedancing, line 196: move this `syncL2StorageMissMeter.Mark(1)` into `if hitInL3`

core/state/snapshot/disklayer.go
```diff
@@ -133,6 +181,50 @@ func (dl *diskLayer) AccountRLP(hash common.Hash) ([]byte, error) {
...
+				// layer 3 miss
+				syncL3StorageMissMeter.Mark(1)
+				cachemetrics.RecordCacheDepth("DISK_L4_STORAGE")
+				cachemetrics.RecordCacheMetrics("DISK_L4_STORAGE", startGetInDisk)
+				cachemetrics.RecordTotalCosts("DISK_L4_STORAGE", start)
+			}
+		}
+		if isMinerMainProcess {
+			// layer 2 miss
+			minerL2StorageMissMeter.Mark(1)
```
forcodedancing, line 212: move `minerL2StorageMissMeter.Mark(1)` into  `if hitInL3 {`

core/state/trie_prefetcher.go
```diff
@@ -308,7 +346,10 @@ func (sf *subfetcher) abort() {
 func (sf *subfetcher) loop() {
```
forcodedancing, line 346: not sure about this loop? do we need to record it?

miner/worker.go
```diff
@@ -81,7 +83,8 @@ const (
 )
 
 var (
-	commitTxsTimer = metrics.NewRegisteredTimer("worker/committxs", nil)
+	commitTxsTimer   = metrics.NewRegisteredTimer("worker/committxs", nil)
```
forcodedancing, line 86: no use of this one `commitTxsTimer`

cmd/geth/main.go
forcodedancing: `MetricsEnabledRecordIOFlag` useless now, please remove it

core/state/snapshot/disklayer.go
flywukong: if we reach this function, it can be concluded that  L2 layer already missed , so it does not matter whether layer3 hit or miss.

core/state/snapshot/disklayer.go
flywukong: no need move it , same reason  as above

core/state/trie_prefetcher.go
flywukong: no , it does not affect main process cost

miner/worker.go
flywukong: yes , it is not added by this pr.
This is the metrics that bsc already has.

cmd/utils/flags.go
flywukong: updated

cmd/geth/main.go
flywukong: updated

core/state/statedb.go
```diff
@@ -391,42 +403,69 @@ func (s *StateDB) GetCodeSize(addr common.Address) int {
...
 		}
 	}()
+
 	stateObject := s.getStateObject(addr)
 	if stateObject != nil {
 		if isSyncMainProcess {
 			l1StorageMeter.Mark(1)
 		}
-		if isMinerMainProcess {
+		if isSyncMainProcess {
```
forcodedancing, line 468: Why change this?

flywukong: This variable does not need to be changed, it has been changed back

PR #786, opened 2022-02-26
fix bug: logs cannot be written in widnows and no tips

gavin2014:
### Description

fix bug: logs cannot be written in widnows and no tips

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck:
reopen if the issue still exist.

PR #792, opened 2022-03-11
[R4R]add sharedStorage for prefetching to L1

flywukong:
### Description

 In the previous prefetch behavior of importing blocks, the prefetch and the main process use completely different state dbs, the prefetch process can only put the retrieved data into the L3 layer, if the main process wants to access, it must go through the L1-L2-L3 layers,  the new design let  the main process and the prefetch process share a shared_pool with the stateDB. This pool stores the originStorage data (each stateobjects has the corresponding
 entry ).  By this way , the prefetch data that can be directly fished to the L1 layer
This shared_pool can be represented as the following secondary map structure, which is a private  member variable in stateDB , each stateObjects has pointer to access the  Corresponding entry of shared_pool

image

This PR also remove the overhead of TryPreload  which cost 2%-3% in main process. 

### Rationale

we can make a lot of data in the main process that originally need to be accessed in L3 directly accessed in L1 layer, which reduces a lot of L2 and L3 layer overhead in the number of layers
 

### Performance
The  two test  machine  is on AWS ,  the two nodes has worked in fullsync mode and has caught up to the latest block

This is the delay of process block , we can see the machine which running master branch data  has nearly same performance data 
image

Then update a machine with this PR code, it can be clearly seen that the delay data is reduced (green performance data), and the average process delay is reduced by about 7%

image

The proportion of I/O overhead in the total block cost of the main process dropped from 33% to 30.5%
image

The overhead of total block cost has reduce from to 3.67% to 1.1% with PR code
image

### Example

### Changes

Notable changes: 
* ...


core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"sync"
+)
+
+// sharedStorage is used to store maps of originStorage of stateObjects
+type SharedStorage struct {
+	poolLock   *sync.RWMutex
+	shared_map map[common.Address]*sync.Map
```
RenRick, line 11: "shared_map" needs to keep the same naming style with poolLock

RenRick (commented):
LGTM

core/blockchain.go
```diff
@@ -2112,10 +2112,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 		var followupInterrupt uint32
 		// For diff sync, it may fallback to full sync, so we still do prefetch
 		if len(block.Transactions()) >= prefetchTxNumber {
-			throwaway := statedb.Copy()
 			go func(start time.Time, followup *types.Block, throwaway *state.StateDB, interrupt *uint32) {
 				bc.prefetcher.Prefetch(followup, throwaway, bc.vmConfig, &followupInterrupt)
-			}(time.Now(), block, throwaway, &followupInterrupt)
+			}(time.Now(), block, statedb, &followupInterrupt)
```
setunapo, line 2117: Would not suggest to share StateDB between Prefetcher and main StateDB.
Use prefetch DB instead, main StateDB can access prefetch DB to get the originStorage

setunapo, line 2117: Would not suggest to share StateDB between Prefetcher and main StateDB.
Use prefetch DB instead, main StateDB can access prefetch DB to get the originStorage

setunapo, line 2117: Would not suggest to share StateDB between Prefetcher and main StateDB.
Use prefetch DB instead, main StateDB can access prefetch DB to get the originStorage

setunapo: Prefetch DB could be used as:  L1.5 Cache

core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Check whether the storage exist in pool,
+// new one if not exist, it will be fetched in stateObjects.GetCommittedState()
+func (storage *SharedStorage) getOrInertStorage(address common.Address) *sync.Map {
+	storage.RLock()
+	storageMap, ok := storage.sharedMap[address]
+	storage.RUnlock()
+	if !ok {
+		m := new(sync.Map)
+		storage.Lock()
```
setunapo, line 30: concurrent race condition?

core/blockchain.go
flywukong: The stateDB of main process an prefetcher are different, they only share the sharedpool which store the originStroage maps , I  had adjust the sharedPool to be L1.5 Cache by the new commit
image

core/state/shared_pool.go
flywukong: updated

core/state/shared_pool.go
flywukong: yes, newObjects called by main preoces or prefetcher will both call this function to check or update the sharedpool

core/state/state_object.go
```diff
@@ -194,6 +204,25 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
...
 }
 
+func (s *StateObject) getStorageKey(key common.Hash) (common.Hash, bool) {
+	if value, cached := s.originStorage[key]; cached {
+		return value, true
+	}
+	// if L1 cache miss, try to get it from shared pool
+	val, ok := s.sharedOriginMap.Load(key)
+	if !ok {
+		return common.HexToHash(""), false
```
j75689, line 214: `return common.Hash{}, false`

flywukong: updated

core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
+package state
+
+import (
+	"github.com/ethereum/go-ethereum/common"
```
guagualvcha, line 4: Please pay attention to package import order

core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
...
+	"sync"
+)
+
+// sharedStorage is used to store maps of originStorage of stateObjects
+type SharedStorage struct {
+	*sync.RWMutex
```
guagualvcha, line 10: No need to be pointer here for `sync.RWMutex`

core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
...
+	sharedMap map[common.Address]*sync.Map
+}
+
+func NewSharedStorage() *SharedStorage {
+	sharedMap := make(map[common.Address]*sync.Map, 1500)
```
guagualvcha, line 15: Why `1500` is used here?  `1500` is not small size,  it means for testnet which has less transactions, it will always allocate much more memory for each block.

core/state/shared_pool.go
```diff
@@ -0,0 +1,36 @@
...
+	return &SharedStorage{
+		&sync.RWMutex{},
+		sharedMap,
+	}
+}
+
+// Check whether the storage exist in pool,
+// new one if not exist, it will be fetched in stateObjects.GetCommittedState()
+func (storage *SharedStorage) getOrInertStorage(address common.Address) *sync.Map {
```
guagualvcha, line 24: `storage` ==> `s`

guagualvcha: I thin we can rename `getOrInertStorage` to `getStorage`, `insert` is an internal action that can be hide from caller.

guagualvcha: It is a bit weird  `SharedStorage` get a function called `getOrInertStorage`, how can we get storage from a storage, I will suggest rename `SharedStorage` to `StoragePool`

core/state/shared_pool.go
flywukong: yes, I has not considered the waste of memory allocation.   I set the value beacause  most of the blocks does not have more than 1200 stateObjects in mainnet

core/state/state_object.go
```diff
@@ -79,7 +80,9 @@ type StateObject struct {
 	trie Trie // storage trie, which becomes non-nil on first access
 	code Code // contract bytecode, which gets set when code is loaded
 
-	originStorage  Storage // Storage cache of original entries to dedup rewrites, reset for every transaction
+	sharedOriginMap *sync.Map // Storage cache of original entries to dedup rewrites, reset for every transaction
```
guagualvcha, line 83: `sharedOriginMap`, the `Map` is a bit duplicated for me since it is already defined as `sync.Map`.  `sharedOriginStorage` seems better.

core/state/state_object.go
```diff
@@ -194,6 +204,28 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
 	return s.GetCommittedState(db, key)
 }
 
+func (s *StateObject) getStorageKey(key common.Hash) (common.Hash, bool) {
```
guagualvcha, line 207: This is quite confusing for me, a function name `getStorageKey`, while it is parameter is a `key`, the return is a `value`?

core/state/state_object.go
```diff
@@ -194,6 +204,28 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
...
+		val, ok := s.sharedOriginMap.Load(key)
+		if !ok {
+			return common.Hash{}, false
+		}
+		return val.(common.Hash), ok
+	}
+	return common.Hash{}, false
+}
+
+func (s *StateObject) setStorgeKey(key common.Hash, value common.Hash) {
```
guagualvcha, line 222: `setStorgeKey` is confusing too.

core/state/state_object.go
```diff
@@ -194,6 +204,28 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
...
+	if value, cached := s.originStorage[key]; cached {
+		return value, true
+	}
+	// if L1 cache miss, try to get it from shared pool
+	if s.sharedOriginMap != nil {
+		val, ok := s.sharedOriginMap.Load(key)
+		if !ok {
+			return common.Hash{}, false
+		}
+		return val.(common.Hash), ok
```
guagualvcha, line 217: `return val.(common.Hash), true` directly

core/state/state_object.go
```diff
@@ -194,6 +204,28 @@ func (s *StateObject) GetState(db Database, key common.Hash) common.Hash {
...
+	}
+	return common.Hash{}, false
+}
+
+func (s *StateObject) setStorgeKey(key common.Hash, value common.Hash) {
+	if s.db.isPrefetchDb {
```
guagualvcha, line 223: I think we dont need `isPrefetchDb` here, but  `if s.sharedOriginMap  != nil`

core/state/state_object.go
```diff
@@ -316,10 +349,12 @@ func (s *StateObject) finalise(prefetch bool) {
 	slotsToPrefetch := make([][]byte, 0, len(s.dirtyStorage))
 	for key, value := range s.dirtyStorage {
 		s.pendingStorage[key] = value
-		if value != s.originStorage[key] {
+		originValue, cached := s.getStorageKey(key)
+		if cached && value != originValue {
```
guagualvcha, line 353: The logic is changed from previous one, please check

core/state/state_object.go
```diff
@@ -352,10 +387,11 @@ func (s *StateObject) updateTrie(db Database) Trie {
 	usedStorage := make([][]byte, 0, len(s.pendingStorage))
 	for key, value := range s.pendingStorage {
 		// Skip noop changes, persist actual changes
-		if value == s.originStorage[key] {
+		originValue, cached := s.getStorageKey(key)
+		if cached && value == originValue {
```
guagualvcha, line 391: Logic changed when it is not cached.

core/state/state_object.go
```diff
@@ -478,7 +514,6 @@ func (s *StateObject) deepCopy(db *StateDB) *StateObject {
 	}
 	stateObject.code = s.code
 	stateObject.dirtyStorage = s.dirtyStorage.Copy()
-	stateObject.originStorage = s.originStorage.Copy()
```
guagualvcha, line 481: Why delete this?

core/state/statedb.go
```diff
@@ -147,6 +150,16 @@ func New(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error)
 	return newStateDB(root, db, snaps)
 }
 
+// NewWithSharedPool creates a new state with sharedStorge on layer 1.5
+func NewWithSharedPool(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB, error) {
+	statedb, err := newStateDB(root, db, snaps)
+	if err != nil {
+		return nil, err
+	}
+	statedb.sharedStorage = NewSharedStorage()
```
guagualvcha, line 159: Please change `sharedStorage` to `storagePool`

core/state/statedb.go
```diff
@@ -818,6 +761,14 @@ func (db *StateDB) ForEachStorage(addr common.Address, cb func(key, value common
 	return nil
 }
 
+// Used by prefetcher
+func (s *StateDB) CopyWithSharedStorage() *StateDB {
```
guagualvcha, line 765: I think we dont need this function

core/state/statedb.go
```diff
@@ -828,6 +779,7 @@ func (s *StateDB) Copy() *StateDB {
 		stateObjects:        make(map[common.Address]*StateObject, len(s.journal.dirties)),
 		stateObjectsPending: make(map[common.Address]struct{}, len(s.stateObjectsPending)),
 		stateObjectsDirty:   make(map[common.Address]struct{}, len(s.journal.dirties)),
+		sharedStorage:       NewSharedStorage(),
```
guagualvcha, line 782: Why `NewSharedStorage`, can't we share it?

core/state/state_object.go
flywukong: I deleted it before using Layer 1.5 ,it should not deleted when using Layer 1.5

core/state/shared_pool.go
flywukong: sure

core/state/statedb.go
```diff
@@ -101,6 +101,8 @@ type StateDB struct {
 	stateObjectsPending map[common.Address]struct{} // State objects finalized but not yet written to the trie
 	stateObjectsDirty   map[common.Address]struct{} // State objects modified in the current execution
 
+	storagePool  *StoragePool // shared_pool to store L1 originStorage of stateObjects
```
guagualvcha, line 104: `shared_pool` ==> `sharedPool`

core/state/statedb.go
```diff
@@ -101,6 +101,8 @@ type StateDB struct {
...
+	dbForSpeedup bool
```
guagualvcha, line 105: What about  `writeOnSharedStorage bool`, with the comment: `Write to the shared origin storage of a stateObject while reading from the underlying storage layer.`

core/state/statedb.go
```diff
@@ -155,6 +167,8 @@ func newStateDB(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB,
 		stateObjects:        make(map[common.Address]*StateObject, defaultNumOfSlots),
 		stateObjectsPending: make(map[common.Address]struct{}, defaultNumOfSlots),
 		stateObjectsDirty:   make(map[common.Address]struct{}, defaultNumOfSlots),
+		storagePool:         nil,
+		dbForSpeedup:        false,
```
guagualvcha, line 171: no need to initialise default value for `dbForSpeedup`.

core/state/statedb.go
```diff
@@ -178,6 +192,10 @@ func newStateDB(root common.Hash, db Database, snaps *snapshot.Tree) (*StateDB,
 	return sdb, nil
 }
 
+func (s *StateDB) SetPrefetchFlag() {
```
guagualvcha, line 195: This can be `EnableWriteOnSharedStorage`

guagualvcha (approved):

guagualvcha (changes requested):
Please fix lint.

core/state/statedb.go
flywukong: updated

core/state/shared_pool.go
flywukong: updated

guagualvcha (approved):

core/state/shared_pool.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Check whether the storage exist in pool,
+// new one if not exist, it will be fetched in stateObjects.GetCommittedState()
+func (s *StoragePool) getStorage(address common.Address) *sync.Map {
+	s.RLock()
+	storageMap, ok := s.sharedMap[address]
+	s.RUnlock()
+	if !ok {
+		m := new(sync.Map)
```
setunapo, line 30: Has race condition.

flywukong: updated

core/state/shared_pool.go
```diff
@@ -0,0 +1,39 @@
...
+
+func NewStoragePool() *StoragePool {
+	sharedMap := make(map[common.Address]*sync.Map)
+	return &StoragePool{
+		sync.RWMutex{},
+		sharedMap,
+	}
+}
+
+// Check whether the storage exist in pool,
```
forcodedancing, line 23: Would be better to follow the commenting style used in go-ethereum, like, `// Copy creates a deep, independent copy of the state.
// Snapshots of the copied state cannot be applied to the copy.
func (s *StateDB) Copy() *StateDB {`

guagualvcha:
LGTM

forcodedancing (approved):

richardrich975 (approved):
LGTM

core/state/shared_pool.go
flywukong: updated

j75689 (approved):

PR #803, opened 2022-03-16
[R4R]Prefetch state data on mining process

qinglin89:
### Description

When miner process mining work, there would be some hard disk access wich are quite time consuming processes.

### Rationale

- Add some prefetch routines to warm up touched data slots by apply transactions concurrently to the given individual state database copy.
mining-prefetcher-design

### Changes

Notable changes:

- Modify `Prefetch` inteface so that it would be compatible with mining `prefetcher`
- Add `PrefetchMining` method for state_prefetcher just like `Prefetch` but a mining version

### DIFF
--- a/core/gaspool.go
+++ b/core/gaspool.go
@@ -27,9 +27,6 @@ type GasPool uint64
 
 // SetGas set an initial value for gaspool
 func (gp *GasPool) SetGas(amount uint64) *GasPool {
-	if amount > math.MaxUint64 {
-		panic("gas pool pushed above uint64")
-	}
 	*(*uint64)(gp) = amount
 	return gp
 }
--- a/core/gaspool.go
+++ b/core/gaspool.go
@@ -25,12 +25,6 @@ import (
 // in a block. The zero value is a pool with zero gas available.
 type GasPool uint64
 
-// SetGas set an initial value for gaspool
-func (gp *GasPool) SetGas(amount uint64) *GasPool {
-	*(*uint64)(gp) = amount
-	return gp
-}
-
 // AddGas makes gas available for execution.
 func (gp *GasPool) AddGas(amount uint64) *GasPool {
 	if uint64(*gp) > math.MaxUint64-amount {
--- a/core/gaspool.go
+++ b/core/gaspool.go
@@ -25,6 +25,15 @@ import (
 // in a block. The zero value is a pool with zero gas available.
 type GasPool uint64
 
+// SetGas set an initial value for gaspool
+func (gp *GasPool) SetGas(amount uint64) *GasPool {
+	if amount > math.MaxUint64 {
+		panic("gas pool pushed above uint64")
+	}
+	*(*uint64)(gp) = amount
+	return gp
+}
+
 // AddGas makes gas available for execution.
 func (gp *GasPool) AddGas(amount uint64) *GasPool {
 	if uint64(*gp) > math.MaxUint64-amount {
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -27,6 +27,7 @@ import (
 )
 
 const prefetchThread = 2
+const checkInterval = 10
 
 // statePrefetcher is a basic Prefetcher, which blindly executes a block on top
 // of an arbitrary state with the goal of prefetching potentially useful state
@@ -133,7 +134,7 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 				return
 			default:
 			}
-			if count++; count%10 == 0 {
+			if count++; count%checkInterval == 0 {
 				if *txCurr == nil {
 					return
 				}
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -108,7 +108,7 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 				select {
 				case tx := <-startCh:
 					// Convert the transaction into an executable message and pre-cache its sender
-					msg, err := tx.AsMessagePrefetch(signer)
+					msg, err := tx.AsMessageNoNonceCheck(signer)
 					if err != nil {
 						return // Also invalid block, bail out
 					}
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -108,14 +108,14 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 				select {
 				case tx := <-txCh:
 					// Convert the transaction into an executable message and pre-cache its sender
-					msg, err := tx.AsMessage(signer)
+					msg, err := tx.AsMessagePrefetch(signer)
 					if err != nil {
 						return // Also invalid block, bail out
 					}
 					idx++
 					newStatedb.Prepare(tx.Hash(), header.Hash(), idx)
 					precacheTransaction(msg, p.config, gaspool, newStatedb, header, evm)
-					gaspool.SetGas(gasLimit)
+					gaspool = new(GasPool).AddGas(gasLimit)
 				case <-stopCh:
 					return
 				}
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
 	}
 }
 
+// PrefetchMining processes the state changes according to the Ethereum rules by running
+// the transaction messages using the statedb, but any changes are discarded. The
+// only goal is to pre-cache transaction signatures and snapshot clean state. Only used for mining stage
+func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce, header *types.Header, gasLimit uint64, statedb *state.StateDB, cfg vm.Config, interruptCh <-chan struct{}, txCurr **types.Transaction) {
+	var signer = types.MakeSigner(p.config, header.Number)
+
+	txCh := make(chan *types.Transaction, 2*prefetchThread)
+	for i := 0; i < prefetchThread; i++ {
+		go func(txCh <-chan *types.Transaction, stopCh <-chan struct{}) {
+			idx := 0
+			newStatedb := statedb.Copy()
+			gaspool := new(GasPool).AddGas(gasLimit)
+			blockContext := NewEVMBlockContext(header, p.bc, nil)
+			evm := vm.NewEVM(blockContext, vm.TxContext{}, statedb, p.config, cfg)
+			// Iterate over and process the individual transactions
+			for {
+				select {
+				case tx := <-txCh:
+					// Convert the transaction into an executable message and pre-cache its sender
+					msg, err := tx.AsMessage(signer)
+					if err != nil {
+						return // Also invalid block, bail out
+					}
+					idx++
+					newStatedb.Prepare(tx.Hash(), header.Hash(), idx)
+					precacheTransaction(msg, p.config, gaspool, newStatedb, header, evm)
+					gaspool.SetGas(gasLimit)
+				case <-stopCh:
+					return
+				}
+			}
+		}(txCh, interruptCh)
+	}
+	go func(txs *types.TransactionsByPriceAndNonce) {
+		count := 0
+		for {
+			tx := txs.Peek()
+			if tx == nil {
+				return
+			}
+			select {
+			case <-interruptCh:
+				return
+			default:
+			}
+			if count++; count%10 == 0 {
+				if *txCurr == nil {
+					return
+				}
+				txs.Forward(*txCurr)
+			}
+			txCh <- tx
+			txs.Shift()
+		}
+	}(txs)
+}
+
 // precacheTransaction attempts to apply a transaction to the given state database
 // and uses the input parameters for its environment. The goal is not to execute
 // the transaction successfully, rather to warm up touched data slots.
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -97,7 +97,7 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 
 	txCh := make(chan *types.Transaction, 2*prefetchThread)
 	for i := 0; i < prefetchThread; i++ {
-		go func(txCh <-chan *types.Transaction, stopCh <-chan struct{}) {
+		go func(startCh <-chan *types.Transaction, stopCh <-chan struct{}) {
 			idx := 0
 			newStatedb := statedb.Copy()
 			gaspool := new(GasPool).AddGas(gasLimit)
@@ -106,7 +106,7 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 			// Iterate over and process the individual transactions
 			for {
 				select {
-				case tx := <-txCh:
+				case tx := <-startCh:
 					// Convert the transaction into an executable message and pre-cache its sender
 					msg, err := tx.AsMessagePrefetch(signer)
 					if err != nil {
@@ -122,10 +122,10 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 			}
 		}(txCh, interruptCh)
 	}
-	go func(txs *types.TransactionsByPriceAndNonce) {
+	go func(txset *types.TransactionsByPriceAndNonce) {
 		count := 0
 		for {
-			tx := txs.Peek()
+			tx := txset.Peek()
 			if tx == nil {
 				return
 			}
@@ -138,10 +138,10 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 				if *txCurr == nil {
 					return
 				}
-				txs.Forward(*txCurr)
+				txset.Forward(*txCurr)
 			}
 			txCh <- tx
-			txs.Shift()
+			txset.Shift()
 		}
 	}(txs)
 }
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -458,6 +458,21 @@ func NewTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]Transa
 	}
 }
 
+// Copy copy a new TransactionsPriceAndNonce with the same *transaction
+func (t *TransactionsByPriceAndNonce) Copy() *TransactionsByPriceAndNonce {
+	heads := make([]*Transaction, len(t.heads))
+	copy(heads, t.heads)
+	txs := make(map[common.Address]Transactions, len(t.txs))
+	for acc, txsTmp := range t.txs {
+		txs[acc] = txsTmp
+	}
+	return &TransactionsByPriceAndNonce{
+		heads:  heads,
+		txs:    txs,
+		signer: t.signer,
+	}
+}
+
 // Peek returns the next transaction by price.
 func (t *TransactionsByPriceAndNonce) Peek() *Transaction {
 	if len(t.heads) == 0 {
@@ -486,6 +501,47 @@ func (t *TransactionsByPriceAndNonce) Pop() {
 
 func (t *TransactionsByPriceAndNonce) CurrentSize() int {
 	return len(t.heads)
+}
+
+//Forward move t to be one index behind tx, tx cant be nil
+func (t *TransactionsByPriceAndNonce) Forward(tx *Transaction) {
+	if tx == nil {
+		txTmp := t.Peek()
+		for txTmp != nil {
+			t.Shift()
+			txTmp = t.Peek()
+		}
+		return
+	}
+
+	l := len(t.heads)
+	acc, _ := Sender(t.signer, tx)
+	for i := 0; i < l; i++ {
+		accTmp, _ := Sender(t.signer, t.heads[i])
+		if acc == accTmp {
+			if tx == t.heads[i] {
+				txTmp := t.Peek()
+				for txTmp != tx {
+					t.Shift()
+					txTmp = t.Peek()
+				}
+				t.Shift()
+				return
+			}
+			for _, txTmp := range t.txs[accTmp] {
+				if txTmp == tx {
+					txTmp = t.Peek()
+					for txTmp != tx {
+						t.Shift()
+						txTmp = t.Peek()
+					}
+					t.Shift()
+					return
+				}
+			}
+			return
+		}
+	}
 }
 
 // Message is a fully derived transaction and implements core.Message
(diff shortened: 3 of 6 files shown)


core/gaspool.go
```diff
@@ -25,6 +25,15 @@ import (
 // in a block. The zero value is a pool with zero gas available.
 type GasPool uint64
 
+// SetGas set an initial value for gaspool
+func (gp *GasPool) SetGas(amount uint64) *GasPool {
```
guagualvcha, line 29: This function seems useless, we can use `new(GasPool).AddGas(gasLimit)` to replace.

core/state_prefetcher.go
```diff
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+// PrefetchMining processes the state changes according to the Ethereum rules by running
+// the transaction messages using the statedb, but any changes are discarded. The
+// only goal is to pre-cache transaction signatures and snapshot clean state. Only used for mining stage
+func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce, header *types.Header, gasLimit uint64, statedb *state.StateDB, cfg vm.Config, interruptCh <-chan struct{}, txCurr **types.Transaction) {
+	var signer = types.MakeSigner(p.config, header.Number)
+
+	txCh := make(chan *types.Transaction, 2*prefetchThread)
+	for i := 0; i < prefetchThread; i++ {
+		go func(txCh <-chan *types.Transaction, stopCh <-chan struct{}) {
+			idx := 0
```
guagualvcha, line 100: `idx` is useless here, please delete it.

core/state_prefetcher.go
```diff
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+			select {
+			case <-interruptCh:
+				return
+			default:
+			}
+			if count++; count%10 == 0 {
+				if *txCurr == nil {
+					return
+				}
+				txs.Forward(*txCurr)
```
guagualvcha, line 140: `txCurr` is a good idea here, but I am worried about race conditions here, it can be a problem. Please well test to ensure it takes well function.

core/types/transaction.go
```diff
@@ -458,6 +458,21 @@ func NewTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]Transa
 	}
 }
 
+// Copy copy a new TransactionsPriceAndNonce with the same *transaction
+func (t *TransactionsByPriceAndNonce) Copy() *TransactionsByPriceAndNonce {
+	heads := make([]*Transaction, len(t.heads))
+	copy(heads, t.heads)
+	txs := make(map[common.Address]Transactions, len(t.txs))
+	for acc, txsTmp := range t.txs {
+		txs[acc] = txsTmp
```
guagualvcha, line 467: We do assign a slice to another directly, not a deep actually. Please double-check no function is changing the content of txsTmp. It looks good to me so far.

core/types/transaction.go
```diff
@@ -488,6 +503,47 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+		txTmp := t.Peek()
+		for txTmp != nil {
+			t.Shift()
+			txTmp = t.Peek()
+		}
+		return
+	}
+
+	l := len(t.heads)
+	acc, _ := Sender(t.signer, tx)
```
guagualvcha, line 518: tx can be nil here because the miner worker is changing the value concurrently.

core/types/transaction.go
```diff
@@ -488,6 +503,47 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
 	return len(t.heads)
 }
 
+//Forward move t to be one index behind tx, tx cant be nil
+func (t *TransactionsByPriceAndNonce) Forward(tx *Transaction) {
+	if tx == nil {
+		txTmp := t.Peek()
```
guagualvcha, line 509: This do not make too much sense. The tx can be nil only when 1. The `Loop` inside `commitTransactions` does not start yet. 2. The `Loop` ends.  Fo the \#1 case, you need to nothong, for the \#2 case, we should abort prefethcer.

core/types/transaction.go
```diff
@@ -488,6 +503,47 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+		for txTmp != nil {
+			t.Shift()
+			txTmp = t.Peek()
+		}
+		return
+	}
+
+	l := len(t.heads)
+	acc, _ := Sender(t.signer, tx)
+	for i := 0; i < l; i++ {
```
guagualvcha, line 519: why not range `t.heads` here?

core/types/transaction.go
```diff
@@ -488,6 +503,47 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+		accTmp, _ := Sender(t.signer, t.heads[i])
+		if acc == accTmp {
+			if tx == t.heads[i] {
+				txTmp := t.Peek()
```
guagualvcha, line 523: If prefethcer is faster than miner worker, the logic here will drain all txs.

core/types/transaction.go
```diff
@@ -488,6 +503,47 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+				for txTmp != tx {
+					t.Shift()
+					txTmp = t.Peek()
+				}
+				t.Shift()
+				return
+			}
+			for _, txTmp := range t.txs[accTmp] {
+				if txTmp == tx {
```
guagualvcha, line 532: same here.

miner/worker.go
```diff
@@ -798,6 +808,7 @@ LOOP:
 					inc:   true,
 				}
 			}
+			close(interruptCh)
```
guagualvcha, line 811: Prefer `defer close(interruptCh) ` after where we define the `interruptCh`

core/gaspool.go
qinglin89: If use `AddGas` instead, may lead to one of these results: 
- reallocate a new `big.Int` struct after each transaction be executed
- set a magic number(much bigger than `block.gaslimit`)
- do some extra calculation to increase up the gas been consumed.

core/state_prefetcher.go
qinglin89: Seems safe, cause there's only one point would write this variable in the same routine `miner.worker`, and one point only too in the `prefetcher` routine would read it which will dispatch to all the sub routines then.
And since for every single `*tx`, it would only occur at most in one left upper corner or the 2d buckets(except that the buckets have the same `*tx` value) when the buckets move on concurrently.

core/types/transaction.go
qinglin89: Have went through the code, `worker` would pick `*tx` out from `txs(types.TransactionsByPriceAndNonce)` which eventually be nil, but not found nowhere modify `*tx`.

core/types/transaction.go
qinglin89: 1. `prefetcher` start before the `woker` routine execute `*tx`, yet i've set some check process that would avoid the *start* nil case in the `prefetcher`, i think we shouldn't rely on the caller to skip the nil check in the method itself.
2. as for the *end* case, `worker` routine would break and then set the signal(close the channel) to abort `prefetcher`, there seems no guarantee that the concurrent `prefecher` routine would catch that signal immediately before touch the nil variable, yet we shouldn't rely on the caller as mentioned before.

core/types/transaction.go
qinglin89: Is that not the expected behavior as we discussed ? :
only one of two conditions will stop the `prefetcher` routine:
- all txs been executed
- `worker` routine finished its job

miner/worker.go
qinglin89: maybe we want to set abort signal as early as possible ?

core/types/transaction.go
guagualvcha: \#2, I think the `shift` operation is CPU intensive, break and wait for the nil variable is more CPU friendly.

miner/worker.go
guagualvcha: The delay is ignorable, prefer code maintainability.

core/types/transaction.go
guagualvcha: Within this loop, it will visit `t.heads[i]` and do `t.Shift()` which will affect the length of `t.heads`, it is a dangerous signal even it is safe currently.  Can we try to move the catch-up logic outsides the loop?

core/types/transaction.go
qinglin89: updated

core/types/transaction.go
qinglin89: updated

miner/worker.go
qinglin89: updated

core/state_prefetcher.go
```diff
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+			tx := txs.Peek()
+			if tx == nil {
+				return
+			}
+			select {
+			case <-interruptCh:
+				return
+			default:
+			}
+			if count++; count%10 == 0 {
```
guagualvcha, line 136: Please make 10 a constant.

guagualvcha (approved):

core/gaspool.go
```diff
@@ -25,6 +25,15 @@ import (
...
+	if amount > math.MaxUint64 {
+		panic("gas pool pushed above uint64")
```
setunapo, line 31: uint64 > math.MaxUint64, Is the check necessary?

core/state_prefetcher.go
```diff
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+				if *txCurr == nil {
+					return
+				}
+				txs.Forward(*txCurr)
+			}
+			txCh <- tx
```
setunapo, line 142: it is good to dispatch to the idle prefetch routine, not dispatch by 0,2,4,6,8... and 1,3,5,7,9

core/types/transaction.go
```diff
@@ -488,6 +503,37 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
 	return len(t.heads)
 }
 
+//Forward move t to be one index behind tx, param tx cant be nil
+func (t *TransactionsByPriceAndNonce) Forward(tx *Transaction) {
```
setunapo, line 507: the code is a bit hard to read...

core/types/transaction.go
```diff
@@ -458,6 +458,21 @@ func NewTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]Transa
 	}
 }
 
+// Copy copy a new TransactionsPriceAndNonce with the same *transaction
+func (t *TransactionsByPriceAndNonce) Copy() *TransactionsByPriceAndNonce {
+	heads := make([]*Transaction, len(t.heads))
+	copy(heads, t.heads)
+	txs := make(map[common.Address]Transactions, len(t.txs))
+	for acc, txsTmp := range t.txs {
```
setunapo, line 466: naming can be more clear.
`acc` stands for accountAddress?
`txs` -> `txsMap`?
`txsTmp` -> `txs`?

core/state_prefetcher.go
setunapo, line 140: `txCurr` is the current transaction the miner is processing.
Why do `Forward` every 10 prefetch tx?

miner/worker.go
```diff
@@ -778,6 +780,15 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
 	}
 	bloomProcessors := core.NewAsyncReceiptBloomGenerator(processorCapacity)
 
+	//	var interruptPrefetch uint32
```
Mercybudda, line 783: Typo  here?

core/state_prefetcher.go
qinglin89: updated

core/gaspool.go
qinglin89: removed

core/state_prefetcher.go
```diff
@@ -88,6 +88,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+				case tx := <-txCh:
+					// Convert the transaction into an executable message and pre-cache its sender
+					msg, err := tx.AsMessage(signer)
+					if err != nil {
+						return // Also invalid block, bail out
+					}
+					idx++
+					newStatedb.Prepare(tx.Hash(), header.Hash(), idx)
+					precacheTransaction(msg, p.config, gaspool, newStatedb, header, evm)
+					gaspool.SetGas(gasLimit)
```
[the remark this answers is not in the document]
qinglin89: Maybe better to reuse the `big.Int`struct instead of reallocating a new one each time ? 

We somehow want `prefetcher` routines run a liittle faster/more than mining routine since it was hard to figure out the accurate transaction set which would finally be in the block.

core/types/transaction.go
qinglin89: annotation added

core/types/transaction.go
qinglin89: as for `acc`, `txs`, just follow the naming convention say like in method `NewTransactionsByPriceAndNonce`

miner/worker.go
```diff
@@ -778,6 +780,15 @@ func (w *worker) commitTransactions(txs *types.TransactionsByPriceAndNonce, coin
...
+	interruptCh := make(chan struct{})
+	defer close(interruptCh)
+	tx := &types.Transaction{}
+	txCurr := &tx
+	//prefetch txs from all pending txs
+	txsPrefetch := txs.Copy()
```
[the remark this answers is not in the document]
qinglin89: `Copy` is not designed as a concurrent access safe method

core/state_prefetcher.go
qinglin89: It's reasonable to consider `prefetcher` routine moving faster than `worker`, as a result, the left up corner in `worker` routine was expected to be in the former position than in prefetcher routine which means `Forward` to the 'former' element would cost a full range check of `TransactionPriceAndNonce.heads`. We check the `txcurr` every once a while(say 10 blocks, and yes there may be some better interval) in case that `worker`routine jumped some `txs` in the pending `tx` list because of gas limit.

miner/worker.go
qinglin89: removed

core/gaspool.go
```diff
@@ -25,6 +25,12 @@ import (
 // in a block. The zero value is a pool with zero gas available.
 type GasPool uint64
 
+// SetGas set an initial value for gaspool
```
yutianwu, line 28: `set` -> `sets`

core/types/transaction.go
```diff
@@ -458,6 +458,21 @@ func NewTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]Transa
 	}
 }
 
+// Copy copy a new TransactionsPriceAndNonce with the same *transaction
```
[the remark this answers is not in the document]
yutianwu: `copy` -> `copys`

core/gaspool.go
```diff
@@ -25,6 +25,12 @@ import (
...
+func (gp *GasPool) SetGas(amount uint64) *GasPool {
```
setunapo, line 29: as we have discussed, can remove this API.

core/gaspool.go
qinglin89: removed

core/gaspool.go
qinglin89: updated

core/types/transaction.go
```diff
@@ -488,6 +503,41 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
 
+//Forward move t to be one index behind tx, param tx cant be nil
+func (t *TransactionsByPriceAndNonce) Forward(tx *Transaction) {
+	//get the sender address of tx
+	acc, _ := Sender(t.signer, tx)
+	for _, head := range t.heads {
+		accTmp, _ := Sender(t.signer, head)
+		if acc == accTmp {
+			//found element in t.headers euqals to tx which means they point to the same transaction
+			if tx == head {
```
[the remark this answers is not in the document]
qinglin89: updated

core/types/transaction.go
```diff
@@ -488,6 +503,41 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
 	return len(t.heads)
 }
 
+//Forward move t to be one index behind tx, param tx cant be nil
+func (t *TransactionsByPriceAndNonce) Forward(tx *Transaction) {
+	//get the sender address of tx
+	acc, _ := Sender(t.signer, tx)
+	for _, head := range t.heads {
+		accTmp, _ := Sender(t.signer, head)
```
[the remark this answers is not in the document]
qinglin89: updated

core/state_prefetcher.go
```diff
@@ -88,6 +89,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+			if tx == nil {
+				return
+			}
+			select {
+			case <-interruptCh:
+				return
+			default:
+			}
+			if count++; count%checkInterval == 0 {
+				if *txCurr == nil {
```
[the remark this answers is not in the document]
qinglin89: updated

core/state_prefetcher.go
```diff
@@ -88,6 +89,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
+			newStatedb := statedb.Copy()
+			gaspool := new(GasPool).AddGas(gasLimit)
+			blockContext := NewEVMBlockContext(header, p.bc, nil)
+			evm := vm.NewEVM(blockContext, vm.TxContext{}, statedb, p.config, cfg)
+			// Iterate over and process the individual transactions
+			for {
+				select {
+				case tx := <-txCh:
+					// Convert the transaction into an executable message and pre-cache its sender
+					msg, err := tx.AsMessage(signer)
```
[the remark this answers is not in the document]
qinglin89: updated

core/types/transaction.go
```diff
@@ -535,6 +588,24 @@ func (tx *Transaction) AsMessage(s Signer) (Message, error) {
...
+		amount:     tx.Value(),
+		data:       tx.Data(),
+		accessList: tx.AccessList(),
+		checkNonce: false,
+	}
+
+	var err error
+	msg.from, err = Sender(s, tx)
+	return msg, err
+
```
yutianwu, line 606: remove blank line

core/types/transaction.go
```diff
@@ -535,6 +588,24 @@ func (tx *Transaction) AsMessage(s Signer) (Message, error) {
 	return msg, err
 }
 
+func (tx *Transaction) AsMessagePrefetch(s Signer) (Message, error) {
```
setunapo, line 591: `AsMessageNoNonceCheck`

guagualvcha (changes requested):
Will do review again after changes

core/state_prefetcher.go
```diff
@@ -88,6 +89,63 @@ func (p *statePrefetcher) Prefetch(block *types.Block, statedb *state.StateDB, c
...
 
+// PrefetchMining processes the state changes according to the Ethereum rules by running
+// the transaction messages using the statedb, but any changes are discarded. The
+// only goal is to pre-cache transaction signatures and snapshot clean state. Only used for mining stage
+func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce, header *types.Header, gasLimit uint64, statedb *state.StateDB, cfg vm.Config, interruptCh <-chan struct{}, txCurr **types.Transaction) {
+	var signer = types.MakeSigner(p.config, header.Number)
+
+	txCh := make(chan *types.Transaction, 2*prefetchThread)
+	for i := 0; i < prefetchThread; i++ {
+		go func(txCh <-chan *types.Transaction, stopCh <-chan struct{}) {
```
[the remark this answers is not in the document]
qinglin89: updated

core/types/transaction.go
qinglin89: updated

core/types/transaction.go
qinglin89: updated

core/types/transaction.go
```diff
@@ -458,6 +458,21 @@ func NewTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]Transa
...
+	heads := make([]*Transaction, len(t.heads))
+	copy(heads, t.heads)
+	txs := make(map[common.Address]Transactions, len(t.txs))
+	for acc, txsTmp := range t.txs {
+		txs[acc] = txsTmp
+	}
+	return &TransactionsByPriceAndNonce{
+		heads:  heads,
+		txs:    txs,
+		signer: t.signer,
```
setunapo, line 472: This is a copy by reference, and `Signer` is not concurrent safe, could have potential risk...
Maybe it is ok right now, since `signer` will not be updated since created.

core/types/transaction.go
```diff
@@ -488,6 +503,44 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+			for txTmp != tx {
+				t.Shift()
+				txTmp = t.Peek()
+			}
+			t.Shift()
+			return
+		}
+	}
+	//get the sender address of tx
+	acc, _ := Sender(t.signer, tx)
```
setunapo, line 526: As noticed above, `t.signer` concurrent race condition;

core/types/transaction.go
```diff
@@ -488,6 +503,44 @@ func (t *TransactionsByPriceAndNonce) CurrentSize() int {
...
+	//check whether target tx exists in t.txs
+	if txs, ok := t.txs[acc]; ok {
+		for _, txTmp := range txs {
+			//found the same pointer in t.txs as tx and then shift t to the position one after tx
+			if txTmp == tx {
+				txTmp = t.Peek()
```
setunapo, line 532: `t.Peek()`?

core/types/transaction.go
qinglin89: wont happen

core/types/transaction.go
qinglin89: yes

core/types/transaction.go
qinglin89: value copy

core/types/transaction.go
setunapo: got it, `Shift` will do tx append.

setunapo (approved):
LGTM

core/types/transaction.go
```diff
@@ -535,6 +588,24 @@ func (tx *Transaction) AsMessage(s Signer) (Message, error) {
 	return msg, err
 }
 
+// AsMessageNoNonceCheck returns the transaction with checkNonce field set to be false.
+func (tx *Transaction) AsMessageNoNonceCheck(s Signer) (Message, error) {
```
guagualvcha, line 592: Try to simplify the code:
```
// AsMessageNoNonceCheck returns the transaction with checkNonce field set to be false.
func (tx *Transaction) AsMessageNoNonceCheck(s Signer) (Message, error) {
	m, err := tx.AsMessage(s)
	if err == nil {
		m.checkNonce = false
	}
	return m, err
}
```

qinglin89: updated

guagualvcha (approved):

forcodedancing (approved):

PR #808, opened 2022-03-17
[R4R]fix code of difflayer not assgin before return

guagualvcha:
### Description
Bugfix for diff layer may have nil code even there is a new smart contract created in the block.

### Rationale
The bug is introduced in pipecommit PR. 
When doing `Commit` of `StateDB`,  the difflayer is returned while the `commmitTrie` routine may assign the code of difflayer in the future.

This bug will impact the diff sync: difflayer may not contains the all changed codes.

### Example
No

### Changes
After this patch, the node can always get the same difflayers.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1333,19 +1333,6 @@ func (s *StateDB) Commit(failPostCommitFunc func(), postCommitFuncs ...func() er
 				}()
 			}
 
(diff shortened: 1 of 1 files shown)


yutianwu (approved):

setunapo (approved):
LGTM

j75689 (approved):

PR #817, opened 2022-03-25
[R4R]fix pruner block tool bug, add some check logic

joeylichang:
### Description

fix prune-block-tool bugs

### Rationale

fix bugs:
1. When the prune-block-tool  is executed, "Failed to decode block body" error may be reported due to txloopuplimit setting.
2. Added some verification logic, because some data related to the block may be deleted after the prune-block-tool is executed.
3. Force the input of ‘--datadir.ancient’ to be an absolute path. Many users always fail to execute because they are caught in the splicing of relative paths of ‘--datadir’ and ‘--datadir.ancient’

### DIFF
--- a/cmd/geth/snapshot.go
+++ b/cmd/geth/snapshot.go
@@ -278,7 +278,8 @@ func pruneBlock(ctx *cli.Context) error {
 	var newAncientPath string
 	oldAncientPath := ctx.GlobalString(utils.AncientFlag.Name)
 	if !filepath.IsAbs(oldAncientPath) {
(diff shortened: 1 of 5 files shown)


guagualvcha (approved):

flywukong:
LGTM

flywukong (approved):

calmbeing (approved):

Mercybudda (approved):

richardrich975 (commented):
LGTM

PR #838, opened 2022-04-04
Fixed disconnect reason check

gvitali:
### Description

To avoid a runtime error when the reason ID is equal to a length of discReasonToString (`runtime error: index out of range [17] with length 17`)

### Rationale

It crashes when it comes with reason ID = 17

### Example

`runtime error: index out of range [17] with length 17`

### Changes

Changed a comparison operator from `<` to `<=`


zzzckck:
already included.

PR #844, opened 2022-04-06
Revert "[R4R] Fix nil statedb in diffsync"

Lippertt:
Reverts bnb-chain/bsc#842


Lippertt:
> Reverts bnb-chain/bsc#842

Issue #821, opened 2022-03-30
Snapshot extension registration failed


lpf763827726:
#### System information

Geth version: `1.1.8-859186f2`
OS & Version: Linux
Commit hash : (if `develop`)

#### Actual behaviour
```
t=2022-03-30T05:00:57+0000 lvl=warn msg="Synchronisation failed, dropping peer"  peer=20719e52103b613a67d6a35f31c8e6359d8efd241fbcd21bf9dbc9fa344b9ba4 err=timeout
t=2022-03-30T05:00:57+0000 lvl=eror msg="Ethereum peer removal failed"           peer=20719e52 err="peer not registered"
t=2022-03-30T05:01:03+0000 lvl=info msg="Looking for peers"                      peercount=1 tried=34  static=6
t=2022-03-30T05:01:22+0000 lvl=eror msg="Snapshot extension registration failed" peer=fda76e15 err="peer connected on snap without compatible eth support"
t=2022-03-30T05:01:23+0000 lvl=eror msg="Snapshot extension registration failed" peer=fde637d5 err="peer connected on snap without compatible eth support"
t=2022-03-30T05:02:04+0000 lvl=info msg="Looking for peers"                      peercount=0 tried=30  static=6
t=2022-03-30T05:02:56+0000 lvl=info msg="Looking for peers"                      peercount=1 tried=40  static=6
t=2022-03-30T05:03:06+0000 lvl=info msg="Looking for peers"                      peercount=2 tried=82  static=6
t=2022-03-30T05:03:08+0000 lvl=info msg="Downloader queue stats"                 receiptTasks=0 blockTasks=372 itemSize="4.05 KiB" throttle=8192
t=2022-03-30T05:03:17+0000 lvl=info msg="Looking for peers"                      peercount=2 tried=74  static=6
t=2022-03-30T05:03:57+0000 lvl=eror msg="Snapshot extension registration failed" peer=dd2fa788 err="peer connected on snap without compatible eth support"
t=2022-03-30T05:03:57+0000 lvl=eror msg="Snapshot extension registration failed" peer=dd02a483 err="peer connected on snap without compatible eth support"
t=2022-03-30T05:04:19+0000 lvl=eror msg="Ethereum peer removal failed"           peer=20719e52 err="peer not registered"

```

#### Steps to reproduce the behaviour
```
./geth_linux --datadir ./bsc init ./genesis.json
[...]

Issue #670, opened 2021-12-16
Ethereum peer removal failed    peer=-------- err="peer not registered"


nilampatel-engineer:
bsc node has downloaded snapshot and now syncing from chain after a while suddenly started showing above error, can some one help pls

Undead8:
I don't know what the error mean, but it's normal. It shows up all the time and it doesn't mean anything is wrong.

nilampatel-engineer:
but the syncing has stopped from last few hours...

Undead8:
Probably not related to that error.

RumeelHussainbnb:
Can you please share the logs and please check its simillar to your issue it might be solve your problem https://github.com/ethereum/go-ethereum/issues/22077
if not please proceed to join our Discord channel for more discussion at https://discord.com/invite/binancesmartchain

tculig:
I'm having the same issue where it says "Ethereum peer removal failed" and the sync just stops dead.

forcodedancing:
@nilampatel-engineer @tculig  We think this could be a potential bug, and we will look into it. Are your nodes recovered? If not, please let us know if we can help. Thanks.

forcodedancing:
Linked to #220

tculig:
My node recovered, thank you @forcodedancing.

nilampatel-engineer:
> My node recovered, thank you @forcodedancing. Took about 20 restarts, but in the end it managed to pull the entire history. Still importing block entries though, can't get it to fully sync. But that's a topic for another thread..

how did you resolved?

tculig:
> > My node recovered, thank you @forcodedancing. Took about 20 restarts, but in the end it managed to pull the entire history. Still importing block entries though, can't get it to fully sync. But that's a topic for another thread..
> 
> how did you resolved?

Just kept restarting the node whenever it crashed :/

nilampatel-engineer:
I have snapshot downloaded and it is crashing on whem chain start syncing and giving me this error about peer removal failed

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

namlt3820:
Same problem here. I got it after pruning the snapshot. Now I have to manually check and restart the node everyday. My other node without any pruning is running just fine.

zarakiftw:
@forcodedancing  Any update on this? My node keeps dropping blocks because of the above issue

White-Rabbit-7:
I am having this exact same issue constantly.

[...]

Issue #220, opened 2021-05-14
Failing to unregister sync peer which is not registered


an0nym0us101:
Using the latest beta release - v1.1.0 beta

A custom node went out of sync so we checked the logs and it looks like it was trying to drop a bad peer but it couldn't stating that the peer is not registered. We had to manually restart geth to fix the issue. This seems like this problem mentioned here - https://github.com/ethereum/go-ethereum/issues/22077

Attaching the bsc.log file, please advise.

[bsc.log](https://github.com/binance-chain/bsc/files/6477826/bsc.log)

knglenn:
I am running a fullnode and have been seeing the same issue since upgrading to the beta release. The node will periodically stop syncing for several minutes at a time, and when checking the logs the last message logged before syncing stopped will be similar to:
```
t=2021-05-16T13:17:38-0700 lvl=eror msg="Ethereum peer removal failed"          peer=6f221a41 err="peer not registered"
```
Sometimes the issue resolves itself after a few minutes and the node will start syncing again, but on several occasions, I've had to stop the node and restart to fix it. When the issue does resolve without a restart, no log messages are emitted for the duration of the freeze. There are no warning/error messages (aside from the "peer removal failed" messages) when syncing starts again.

an0nym0us101:
For me it doesnt restart syncing automatically, it keeps throwing the error in the bsc.log file I have attached. I have to manually stop geth and restart it. Worst part is it happens randomly at anytime so you have to keep checking if your node is in sync. It has happened twice today already, once after a few hours and once within a few minutes after having restarted geth. On the third run now!

JohnsonCaii:
Same issues here, would like to know how did you finally resolve this?

The node kept logging the `removal failed`, it almost lasts 2 hours before getting back to normal.
```
WARN [06-01|01:30:02.751] Synchronisation failed, dropping peer    peer=83b0745fe580f90960348cad600aae35bbdec49114b43403f3000984fc61bc4f err=timeout
ERROR[06-01|01:30:05.214] Ethereum peer removal failed             peer=83b0745f err="peer not registered"
WARN [06-01|01:30:17.349] Synchronisation failed, dropping peer    peer=d87ebe2bcb14571d81facb84eb4c0b85409768676c69bb6e933eb83b146d2f3b err=timeout
ERROR[06-01|01:30:20.279] Ethereum peer removal failed             peer=d87ebe2b err="peer not registered"
[...]

PR #856, opened 2022-04-12
fix logic issue: handlers.removePeer() is called twice.

Jolly23:
### Description

There is a logic issue which causes "Ethereum peer removal failed, err=peer not registered" occur quite often.
Related issues #220 #670 #821

### Rationale

`handler.runEthPeer()` set up a `defer removePeer().` This is always called after a peer is disconnected.
However removePeer is also called by multiple functions like downloader/fetcher.  After those kind of functions call `removePeer()`, peer handler executes `defer removePeer()`. This makes `removePeer()` happened twice, and this is the reason we often see "Ethereum peer removal failed, err=peer not registered".

Just took a look, [go-ethereum code here](https://github.com/ethereum/go-ethereum/blob/59f0e8ae60c777bef384f045edf2a816c4a3ca9d/eth/handler.go#L484) is exactly using this new logic

### Changes

Notable changes: 
* `removePeer()` hard Disconnect peer from networking layer. 
* `defer unregisterPeer()` will called by  `handler.runEthPeer()`, do the cleanup task after then.
* add `NewPeerPipe()` under `p2p.Peer`, for RW close testing. The code is referenced from [go-ethereum-p2p-peer](https://github.com/ethereum/go-ethereum/blob/c40943a167b603382bbd0f45c92c9b9c2669a080/p2p/peer.go#L143)


guagualvcha:
Thanks for the contribution. We will start review ASAP.

guagualvcha (approved):

guagualvcha:
LGTM

Jolly23:
Related issues #220 #670 #821

j75689 (approved):

Loverush (approved):

KeefeL (approved):
LGTM

PR #862, opened 2022-04-18
add prune ancient feature

joeylichang:
### Description

Block Data no longer transfer to AncientDB when delete from KVStore.
Save more os resources, disk space, disk bandwidth, cpu.

### Rationale

Release more disk space, disk bandwidth resources.

### Example

"./geth" start command add "--puneancient" parameter

### Changes

Notable changes: 
* add punedfreezer struct inherit AncientStore interface, implement not save to AncientDB when delete from KVStore
* add 'PruneAncientFlag'(record abcinet type) and 'StableStateBlockNumber'(record lastest rootHash savetodisk)
* add check logic for block data is nil, because maybe delete when transfer to AncientDB

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1629,7 +1629,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		if cfg.SyncMode == downloader.FullSync {
 			cfg.PruneAncientData = ctx.GlobalBool(PruneAncientDataFlag.Name)
 		} else {
-			log.Crit("pruneancient parameter take effect in full syncmode")
+			log.Crit("pruneancient parameter didn't take effect for current syncmode")
 		}
 	}
 	if gcmode := ctx.GlobalString(GCModeFlag.Name); gcmode != "full" && gcmode != "archive" {
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -457,6 +457,10 @@ var (
 		Usage: "The number of blocks should be persisted in db (default = 86400)",
 		Value: uint64(86400),
 	}
+	PruneAncientDataFlag = cli.BoolFlag{
+		Name:  "pruneancient",
+		Usage: "Keep a fixed number of blocks(9w) in kvstore under full syncmode, enable release more os resources",
+	}
 	// Miner settings
 	MiningEnabledFlag = cli.BoolFlag{
 		Name:  "mine",
@@ -1621,6 +1625,13 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	if ctx.GlobalIsSet(DiffBlockFlag.Name) {
 		cfg.DiffBlock = ctx.GlobalUint64(DiffBlockFlag.Name)
 	}
+	if ctx.GlobalIsSet(PruneAncientDataFlag.Name) {
+		if cfg.SyncMode == downloader.FullSync {
+			cfg.PruneAncientData = ctx.GlobalBool(PruneAncientDataFlag.Name)
+		} else {
+			log.Crit("pruneancient parameter take effect in full syncmode")
+		}
+	}
 	if gcmode := ctx.GlobalString(GCModeFlag.Name); gcmode != "full" && gcmode != "archive" {
 		Fatalf("--%s must be either 'full' or 'archive'", GCModeFlag.Name)
 	}
@@ -1914,7 +1925,7 @@ func MakeChainDatabase(ctx *cli.Context, stack *node.Node, readonly, disableFree
 		chainDb, err = stack.OpenDatabase(name, cache, handles, "", readonly)
 	} else {
 		name := "chaindata"
-		chainDb, err = stack.OpenDatabaseWithFreezer(name, cache, handles, ctx.GlobalString(AncientFlag.Name), "", readonly, disableFreeze, false)
+		chainDb, err = stack.OpenDatabaseWithFreezer(name, cache, handles, ctx.GlobalString(AncientFlag.Name), "", readonly, disableFreeze, false, false)
 	}
 	if err != nil {
 		Fatalf("Could not open database: %v", err)
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -459,7 +459,7 @@ var (
 	}
 	PruneAncientDataFlag = cli.BoolFlag{
 		Name:  "pruneancient",
-		Usage: "Keep a fixed number of blocks(9w) in kvstore under full syncmode, enable release more os resources",
+		Usage: "Prune ancient data, recommends to the user who don't care about the ancient data. Note that once be turned on, the ancient data will not be recovered",
 	}
 	// Miner settings
 	MiningEnabledFlag = cli.BoolFlag{
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -459,7 +459,7 @@ var (
 	}
 	PruneAncientDataFlag = cli.BoolFlag{
 		Name:  "pruneancient",
-		Usage: "Prune ancient data, recommends to the user who don't care about the ancient data. Note that once be turned on, the ancient data will not be recovered",
+		Usage: "Prune ancient data, recommends to the user who don't care about the ancient data. Note that once be turned on, the ancient data will not be recovered again",
 	}
 	// Miner settings
 	MiningEnabledFlag = cli.BoolFlag{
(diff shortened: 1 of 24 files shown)


core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/prometheus/tsdb/fileutil"
```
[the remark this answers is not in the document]
joeylichang: make lint check package order, should placed "fileutil" pkg last

cmd/utils/flags.go
```diff
@@ -457,6 +457,10 @@ var (
 		Usage: "The number of blocks should be persisted in db (default = 86400)",
 		Value: uint64(86400),
 	}
+	PruneAncientDataFlag = cli.BoolFlag{
+		Name:  "pruneancient",
+		Usage: "Keep a fixed number of blocks(9w) in kvstore under full syncmode, enable release more os resources",
```
guagualvcha, line 462: Please emphasize who is suggested to enable this feature, like anyone who don't care about the history blocks.

guagualvcha: And warn user the risk to enable this feature, like they can not revert this setting, they can not recover history data by disable this feature.

core/blockchain.go
```diff
@@ -2355,6 +2360,9 @@ func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator) (i
 	for i := len(hashes) - 1; i >= 0; i-- {
 		// Append the next block to our batch
 		block := bc.GetBlock(hashes[i], numbers[i])
+		if block == nil {
+			continue
```
guagualvcha, line 2364: If it happens, continue will cause chaos. Is it possible to happen?

core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+)
+
+// nodatafreezer is an empty freezer, only record 'frozen' , the next recycle block number form kvstore.
+type prunedfreezer struct {
```
guagualvcha, line 19: What is the name of this freezer,  `prunedfreezer` or `nodatafreezer`? Prefer `emptyFreezer` personaly.

core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// HasAncient returns an indicator whether the specified ancient data exists, return nil.
+func (f *prunedfreezer) HasAncient(kind string, number uint64) (bool, error) {
+	return false, nil
+}
+
+// Ancient retrieves an ancient binary blob from prunedfreezer, return nil.
+func (f *prunedfreezer) Ancient(kind string, number uint64) ([]byte, error) {
+	if _, ok := FreezerNoSnappy[kind]; ok {
```
guagualvcha, line 110: It will always returnn nil, try simplify it?

core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+// AppendAncient update frozen.
+//
+// Notably, this function is lock free but kind of thread-safe. All out-of-order
+// injection will be rejected. But if two injections with same number happen at
+// the same time, we can get into the trouble.
+func (f *prunedfreezer) AppendAncient(number uint64, hash, header, body, receipts, td []byte) (err error) {
+	if atomic.LoadUint64(&f.frozen) != number {
+		return errOutOrderInsertion
+	}
+	atomic.AddUint64(&f.frozen, 1)
```
guagualvcha, line 151: Shall we just keep an empty implementation?

core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+			if hash != (common.Hash{}) {
+				ancients = append(ancients, hash)
+			}
+		}
+		// Batch of blocks have been frozen, flush them before wiping from leveldb
+		if err := f.Sync(); err != nil {
+			log.Crit("Failed to flush frozen tables", "err", err)
+		}
+		// Batch of blocks have been frozen, flush them before wiping from leveldb
+		backoff = f.frozen-first >= freezerBatchLimit
```
guagualvcha, line 283: In most cases, `backoff` is false, will this keep this routine too busy to occupy one core of CPU, which is not necessary?

core/rawdb/prunedfreezer.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+			if hash != (common.Hash{}) {
+				ancients = append(ancients, hash)
+			}
+		}
+		// Batch of blocks have been frozen, flush them before wiping from leveldb
+		if err := f.Sync(); err != nil {
+			log.Crit("Failed to flush frozen tables", "err", err)
+		}
+		// Batch of blocks have been frozen, flush them before wiping from leveldb
```
guagualvcha, line 282: Duplicated comment `Batch of blocks have been frozen, flush them before wiping from leveldb`

core/blockchain.go
joeylichang: It won't happen because delete block that block has state
above 'hashes' is that no state block hash, so will access ancient data
add 'if block == nil ... continue’ judgement , only consider code rigor.

core/rawdb/prunedfreezer.go
joeylichang: I think the type of error should be distinguished. For tables outside the anti-counterfeiting range or other non-existent tables, the corresponding error should be returned. The upper layer can respond according to different situations. Just returning nil may affect the judgment of the upper layer.
At the same time, it can also prevent simple bugs. For example, when the tool accesses ancient, the name of the table may be misspelled, and it is impossible to locate the problem by simply returning nil.

core/blockchain.go
```diff
@@ -1764,6 +1768,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 							}
 							// Flush an entire trie and restart the counters
 							triedb.Commit(header.Root, true, nil)
+							rawdb.WriteStableStateBlockNumber(bc.db, current)
```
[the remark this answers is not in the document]
joeylichang: 1. 'Stable' reflects that the block state is complete/reliable, and CheckPoint cannot better reflect the block state. So it is more inclined to the expression of 'Stable'.
2. Yes, I checked the stop logic, indeed there should be a compass here, I will add it.

core/blockchain.go
guagualvcha: I see, I suggest log fatal here, to avoid unexpected consequence.

core/blockchain.go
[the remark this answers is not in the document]
guagualvcha: It is difficult for me to relate `Stable` with a complete block state. `Stable` means immutable, usable. Better name is suggested.

guagualvcha:
Besides, can you evaluate how this feature impacts the tx index system?

joeylichang:
> Besides, can you evaluate how this feature impacts the tx index system?

It has no effect. It has been fixed for the online problem (that is, the tx index error reported from ancient data after executing the pruneblock tool). The fix is to check whether the block range that needs to be indexed is in ancient before creating the index. If it is not, it will Adjust the boundaries of the block range.

code：
https://github.com/bnb-chain/bsc/pull/862/files/174e3cf6b35fe18c56f99e850321b0d7e5aaa11d#diff-5d45a05258d00e090bc241a1b22c8eb8f4183ac91bd480c46835bc72c8a15c6a

calmbeing (approved):

Mercybudda (dismissed):

KeefeL (approved):

j75689 (dismissed):

Mercybudda (dismissed):

KeefeL (approved):

j75689 (approved):

Mercybudda (approved):

richardrich975 (approved):
LGTM

guagualvcha (approved):

core/rawdb/freezer.go
```diff
@@ -494,3 +419,81 @@ func (f *freezer) repair() error {
...
+			}
+		}
+	}
+	if err := batch.Write(); err != nil {
+		log.Crit("Failed to delete frozen side blocks", "err", err)
+	}
+	batch.Reset()
+
+	// Step into the future and delete and dangling side chains
+	if frozen > 0 {
```
zzzckck, line 457: this code is quite risky, it could even delete the whole chain, is the "dangling" was created with a bug, such as issue: https://github.com/bnb-chain/bsc/issues/2529
Maybe we don't need it, as we have already deleted the "dangling" block from first to frozen.

PR #864, opened 2022-04-19
Fast node refactor

j75689:
### Description
1. fix the diffhash issue
2. refactor block remote validation code
3. add unit-tests

## DiffHash Issue
When using noTrie mode on the test node, we found a problem, due to the change of pipecommit, the update of snapshots was changed to be parallelized with insertBlock. This will cause a problem. When insertBlock, the EVM execution process needs to read data from the snpashot, and the snapshots refresh the data in the background at the same time, and the difflayer that has been merged into the underlying snapshot will be updated to the stale state, and the reader will get it. an empty array.
In the node with Trie, the correct data can be retrieved from the Trie Tree, but because the noTrie mode cannot retrieve any data from the Trie tree, it will cause problems.

In order to solve this issue, we have introduced a fix.
Reading snapshot data in noTrie mode will retry if it gets a stale error.

Under normal case, the snapshot updated to be outdated is not read at the time of reading, but there is a time gap between the snpashot update and the read lock, and there is a chance  to get stale difflayer under parallel execution.


core/remote_state_verifier.go
```diff
@@ -69,14 +71,26 @@ func NewVerifyManager(blockchain *BlockChain, peers verifyPeers, allowInsecure b
 		if oldBlock == nil {
 			return nil, fmt.Errorf("block is nil, number: %d", number)
 		}
-		blockHash := oldBlock.Hash()
-		_, err := blockchain.GenerateDiffLayer(blockHash)
-		if err != nil {
-			return nil, err
+
+		// rewind to last non verified block
+		blockchain.SetHead(oldBlock.NumberU64())
```
NashBC, line 76: It seems we don't need a loop here, just check the number, if number > maxForkHeight, rewind to number-maxForkHeight is OK, if not rewind to a genesis block?

core/remote_state_verifier.go
```diff
@@ -69,14 +71,26 @@ func NewVerifyManager(blockchain *BlockChain, peers verifyPeers, allowInsecure b
...
+		block = oldBlock
+		break
+	}
+
+	number = block.Number()
+	for i := maxForkHeight; i >= 0; i-- {
+		if new(big.Int).Sub(number, big.NewInt(int64(i))).Cmp(common.Big0) <= 0 {
+			continue
```
NashBC, line 84: break is better than continue here

core/remote_state_verifier.go
```diff
@@ -69,14 +71,26 @@ func NewVerifyManager(blockchain *BlockChain, peers verifyPeers, allowInsecure b
...
 		}
-		diffLayerCh := make(chan struct{})
-		close(diffLayerCh)
-		blockchain.diffLayerChanCache.Add(blockHash, diffLayerCh)
+		oldBlock := blockchain.GetBlockByNumber(number.Uint64() - uint64(i))
```
NashBC, line 86: Use GetBlockByHash not GetBlockByNumber will be better, we should cache parent block

eth/downloader/downloader.go
```diff
@@ -1812,6 +1813,9 @@ func (d *Downloader) importBlockResults(results []*fetchResult) error {
 			// of the blocks delivered from the downloader, and the indexing will be off.
 			log.Debug("Downloaded item processing failed on sidechain import", "index", index, "err", err)
 		}
+		if errors.Is(err, core.ErrAncestorHasNotBeenVerified) {
+			return err
```
NashBC, line 1817: Return err will make the peer disconnected, shall we return err here?

core/remote_state_verifier.go
```diff
@@ -126,16 +143,21 @@ func (vm *remoteVerifyManager) mainLoop() {
...
 		case message := <-vm.messageCh:
+			vm.taskLock.RLock()
 			if vt, ok := vm.tasks[message.verifyResult.BlockHash]; ok {
 				vt.messageCh <- message
 			}
+			vm.taskLock.RUnlock()
 
 		// System stopped
 		case <-vm.bc.quit:
+			vm.taskLock.RLock()
```
NashBC, line 156: In this loop go routine, we only need write lock before operating vm.tasks, the read lock we can skip?

eth/downloader/downloader.go
j75689: No, the peer will be disconnected is caused by the line 1819, the error was wrapped with errInvalidChain, and line 408 will detect this err to drop the connection. so in this line, we don't return errors with wrapping errInvalidChain, and the connection wouldn't be disconnected.

NashBC: Great

core/remote_state_verifier.go
j75689: fixed

core/remote_state_verifier.go
j75689: fixed

core/remote_state_verifier.go
j75689: fixed

core/remote_state_verifier.go
```diff
@@ -60,23 +62,25 @@ func NewVerifyManager(blockchain *BlockChain, peers verifyPeers, allowInsecure b
...
+
+	// rewind to last non verified block
+	number := new(big.Int).Sub(block.Number(), big.NewInt(int64(maxForkHeight)))
+	if number.Cmp(common.Big0) < 0 {
+		blockchain.SetHead(0)
+	} else {
+		numberU64 := number.Uint64()
+		blockchain.SetHead(numberU64)
+		block := blockchain.GetBlockByNumber(numberU64)
+		for i := 0; i < maxForkHeight; i++ {
```
NashBC, line 74: There may not enough parent blocks,  i < maxForkHeight && block.Number().Uint64 > 0 ?

NashBC (approved):

core/remote_state_verifier.go
j75689: fixed

core/blockchain.go
```diff
@@ -227,7 +227,7 @@ type BlockChain struct {
 	// trusted diff layers
 	diffLayerCache             *lru.Cache   // Cache for the diffLayers
 	diffLayerRLPCache          *lru.Cache   // Cache for the rlp encoded diffLayers
-	diffLayerChanCache         *lru.Cache   // Cache for
+	diffLayerChanCache         *sync.Map    // Cache for
```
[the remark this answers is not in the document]
j75689: fixed

PR #865, opened 2022-04-22
x = append(y) is equivalent to x = y

freaker2k7:
I noticed this redundant operation.

So I removed the unnecessary append.

Less operation - faster node :)


freaker2k7:
Can anyone please confirm this fail comes form my change ?!

freaker2k7:
closing in favor of #866

PR #867, opened 2022-04-23
no value of type uint64 is less than 0

freaker2k7:
I removed this redundant code, because a check on a `uint64` type whether it's less than 0 is useless.
Also, remove the redundant `uint64` casting as both `bavail` and `stat.Bavail` are already `uint64`.

Is it possible to get negative number of blocks on FreeBSD !?
Because `stat.Bavail` is `uint64`...

Please correct me if I'm wrong.

A few nanoseconds faster is always better ^_^


freaker2k7:
closing in favor of #868

PR #870, opened 2022-04-24
[R4R]fix:Shift panic for zero length of heads

qinglin89:
### Description

panic when length of heads equal to zero when `shift`

### Rationale

- `TransactionsByPriceAndNonce.Shift` should only be called after `TransactionsByPriceAndNonce.Peek` directly and void invoke on `TransactionsByPriceAndNonce.Shift` when `Peek` show a nil result, otherwise might cause a panic for zero length `TransactionsByPriceAndNonce.heads` when doing `Shift`
- `func (p *statePrefetcher) PrefetchMining(...)` invoked `TransactionsByPriceAndNonce.Forward` before `TransactionsByPriceAndNonce.Shift` with no `TransactionsByPriceAndNoncePeek` check while `TransactionsByPriceAndNonce.Forward` might change `TransactionsByPriceAndNonce.heads` to zero length.
### Example

N/A

### Changes

Notable changes: 
* rearrange calling sequence in the dispatcher of `prefechmining`, make sure `Peek` be checked before `Shift`
    * `func (p *statePrefetcher) PrefetchMining(...)`

### DIFF
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -127,6 +127,13 @@ func (p *statePrefetcher) PrefetchMining(txs *types.TransactionsByPriceAndNonce,
 	go func(txset *types.TransactionsByPriceAndNonce) {
 		count := 0
 		for {
+			if count++; count%checkInterval == 0 {
+				txTmp := *txCurr
+				if txTmp == nil {
+					return
+				}
+				txset.Forward(txTmp)
+			}
 			tx := txset.Peek()
 			if tx == nil {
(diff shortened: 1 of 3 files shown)


guagualvcha (approved):

guagualvcha (changes requested):
Will do review again.

guagualvcha (approved):

Mercybudda (approved):

forcodedancing (approved):

PR #898, opened 2022-05-09
[R4R] implement mimc hash function

forcodedancing:
### Description

Implement native support for MiMC hash function.

### Rationale

MiMC hash function is widely used in blockchain/web3 now, with this feature it will benefit many smart contract developers who are using MiMC in their projects.

### Example

Call the pre-defined contract address for using it.

image

### Changes

Notable changes: 
* add precompiles

### Others

`TODO` Change fork name


core/vm/contracts.go
```diff
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"math/big"
 
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr/mimc"
```
yutianwu, line 25: fix the imports

forcodedancing: updated

LBruyne:
The input parameter of this MiMc function is a string, right?

guagualvcha:
close because seems useless now.

PR #923, opened 2022-05-23
[WIP] 200m txpool pre-mining v1

qinglin89:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

importing process time consuming comparison between two instances with two version of nodes (about 30%)
importing process on 2 instances
total_mgasps without premining
nopre-mgasps
total_mgasps with pre-mining
premining-mgasps

importing process time consuming on the same instance between two version of nodes
importing process


miner/worker.go
```diff
@@ -383,10 +410,23 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+			select {
+			case w.preMiningInterruptOnMining <- struct{}{}:
+			default:
+			}
 			commit(true, commitInterruptNewHead)
 
 		case head := <-w.chainHeadCh:
 			if !w.isRunning() {
+				select {
+				case w.txpoolChainHeadCh <- struct{}{}:
```
setunapo, line 427: worker is not running, when downloader is started, it is for quick block catch up mode.
when downloader is done, the worker will be restarted, but downloader could be started again to downloader more header, block, receipt...
When downloader is running, it is better not to do TxPoolPrefetch.

Actually, TxPoolPrefetch should be executed only when the node has synced to the latest block

miner/worker.go
```diff
@@ -395,13 +435,32 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
 				signedRecent, err := p.SignRecently(w.chain, head.Block.Header())
 				if err != nil {
 					log.Info("Not allowed to propose block", "err", err)
+					//invoke pre-mining by notify txpoolsnapshotloop without blocking if not available
```
setunapo, line 438: Maybe TxPoolPrefetch is not necessary for nodes which are not allowed to propose block.
TxPoolPrefetch is for validators to reduce the time of block import, so it can have more time to mine when it is inturn.

miner/worker.go
```diff
@@ -1050,3 +1119,170 @@ func (w *worker) postSideBlock(event core.ChainSideEvent) {
...
+						tmp := w.calPendingTxs()
+						currentPoolTxsCh <- tmp
+						j++
+						time.Sleep(20 * time.Millisecond)
+					}
+					break loopIntern
+				case <-w.resetPoolSnapshot:
+					//signal from resetPoolSnapshot is used to interrupt snapshot process on txpool and reset the timer to wait for next round caused by locally mining
+					log.Debug("txpoolSnapshot resetPoolSnapshot on mining locally with 2700ms duration")
+					timer = time.NewTimer(2700 * time.Millisecond)
```
setunapo, line 1171: why 2700ms?

miner/worker.go
```diff
@@ -395,13 +435,32 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
 					log.Info("Signed recently, must wait")
+					//invoke pre-mining by notify txpoolsnapshotloop without blocking if not available
+					select {
+					case w.txpoolChainHeadCh <- struct{}{}:
+					default:
+					}
 					continue
 				}
 			}
+			//reset txpoolsnapshot and interrupt current pre-mining
```
setunapo, line 455: For 21 validators, 11 validators are not in recents, and they could reach here.
Then these 11 validators will not be allowed to do TxPoolPrefetch?

miner/worker.go
```diff
@@ -414,6 +473,15 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
 					timer.Reset(recommit)
 					continue
 				}
+				//reset txpoolsnapshot and interrupt current pre-mining
```
setunapo, line 476: why reset here? 
`timer` if for resubmit for new arrived high priced transactions, the default resubmit interval is 3 second?
I am not sure, how this timer is used?

PR #926, opened 2022-05-24
[R4R] Separate Processing and State Verification on BSC

dean65:
## Motivation

Currently a BSC node has two kinds of state world: MPT and snapshot. MPT(Merkle Patricia Tries) is a tree-structured state world. The key function of MPT is to generate the state root to ensure state consistency, while the query/commit on MPT is quite slow. Snapshot is a flattened key-value-based state world. Snapshot provides fast queries and commits. The storage size of the snapshot increases slowly even with a large transaction volume. Snapshot is usually used for block processing, while MPT is used for state verification.

In order to lower the hardware requirement and keep security, we introduce two type nodes to make full use of different storages, one is named **fast node**, and the other is named **verify node**. The **fast node** will do block processing with snapshot, it will do all verification against blocks except state root. The **verify node** receives  diffhash from the fast node then responds MPT root to the **fast node**. 

The **fast node** doesn’t need to store MPT, so the storage and computation requirement will be much lower.

### Architecture
The network topology of the **fast node** and **verify node**:
topology

**Fast node** is a bsc client that does fullsync using only Snapshot and generates difflayer. It needs the confirm message from the **verify node** before freezing the blocks, it has to wait until it receives a confirm message from the **verify node** for the ancestor block before inserting new blocks.

**Verify node** is a normal bsc full node that does fullsync using Snapshot and MPT, generates difflayer. It receives diffhash from **fast nodes**, finds corresponding difflayer whose diffhash is matched, then it responds MPT root message to the **fast node**. 

All the messages exchanged between **fast node** and **verify node** are based on the **trust p2p protocol**.

The relationship between the **fast node** and the **verify node**:
architecture

On the **fast node** side, for each inserted block, verify manager creates a verify task, the verify task submits the block’s diffhash to multiple trusted **verify nodes**, and receives response from them, once received the confirm message, the block has been verified.

On the **verify node** side, it receives requests from **fast nodes**, does corresponding processing and then responds to **fast nodes** immediately. The processing of diffhash finds difflayer from cache first, if not found then search from diff store, if still no difflayer found, then find from blockchain headers, this processing would be very quick. 

### Authentication
**Fast node** can only rely on trusted **verify node**, either deployed by the same developer or deployed by a trusted organization. We know the different peers will verify the peer id during the handshake, we will borrow this mechanism to do the authentication. We introduce **VerifyNodes** settings, which is a list of encoded addresses, the **fast nodes** only build connections(based on trust protocol) with **VerifyNodes**.

### Chain Tools

### Prototype verification

#### Fast Node
Introduce a new **tries-verify-mode** setting, there are four modes:
- **local**: a normal full node with complete state world(both MPT and snapshot), merkle state root will be verified against the block header.
- **full**: a fast node with only snapshot state world. Merkle state root is verified by the trustworthy remote verify node by comparing the diffhash(an identify of difflayer generated by the block) and state root.
- **insecure**: same as full mode, except that it can tolerate without verifying the diffhash when verify node does not have it.
- **none**: no merkle state root verification at all, there is no need to setup or connect remote verify node at all, it is more light comparing to full and insecure mode, but get a very little chance that the state is not consistent with other peers.

If the fast node runs in not local mode, the node will disable diff protocol by default, If the fast node runs in full or light mode, the node will enable trust protocol by default.

` ./geth --config ./config.toml --datadir ./node --syncmode full --cache 5000 --tries-verify-mode insecure`

#### Verify node
When a full node has enabled the trust protocol, it can serve as a verify node, at the same time, we will recommend you to enable persist diff, disable snap protocol and diff protocol when running a verify node.

` ./geth --config ./config.toml --datadir ./node --syncmode full --cache 5000 --persistdiff --enabletrustprotocol --disablesnapprotocol --disablediffprotocol`

#### Prune tries node

## Noteable
When using noTrie mode on the test node, we found a problem, due to the change of pipecommit, the update of snapshots was changed to be parallelized with insertBlock. This will cause a problem. When insertBlock, the EVM execution process needs to read data from the snpashot, and the snapshots refresh the data in the background at the same time, and the difflayer that has been merged into the underlying snapshot will be updated to the stale state, and the reader will get it. an empty array.
In the node with Trie, the correct data can be retrieved from the Trie Tree, but because the noTrie mode cannot retrieve any data from the Trie tree, it will cause problems.

In order to solve this issue, we have introduced a fix.
Reading snapshot data in noTrie mode will retry if it gets a stale error.

Under normal case, the snapshot updated to be outdated is not read at the time of reading, but there is a time gap between the snpashot update and the read lock, and there is a chance  to get stale difflayer under parallel execution.


eth/ethconfig/gen_config.go
```diff
@@ -42,28 +50,30 @@ func (c Config) MarshalTOML() (interface{}, error) {
...
 		DocRoot                 string `toml:"-"`
 		EWASMInterpreter        string
 		EVMInterpreter          string
-		RPCGasCap               uint64                         `toml:",omitempty"`
-		RPCTxFeeCap             float64                        `toml:",omitempty"`
+		RPCGasCap               uint64
+		RPCTxFeeCap             float64
 		Checkpoint              *params.TrustedCheckpoint      `toml:",omitempty"`
 		CheckpointOracle        *params.CheckpointOracleConfig `toml:",omitempty"`
+		OverrideBerlin          *big.Int                       `toml:",omitempty"`
```
yutianwu, line 76: why add `OverrideBerlin`

core/state/statedb.go
```diff
@@ -201,6 +204,9 @@ func (s *StateDB) EnableWriteOnSharedStorage() {
 func (s *StateDB) StartPrefetcher(namespace string) {
 	s.prefetcherLock.Lock()
 	defer s.prefetcherLock.Unlock()
+	if s.noTrie {
```
yutianwu, line 207: can we move it before  `s.prefetcherLock.Lock()`

core/state/statedb.go
```diff
@@ -215,6 +221,9 @@ func (s *StateDB) StartPrefetcher(namespace string) {
 func (s *StateDB) StopPrefetcher() {
 	s.prefetcherLock.Lock()
 	defer s.prefetcherLock.Unlock()
+	if s.noTrie {
```
yutianwu, line 224: aslo here

core/state/statedb.go
```diff
@@ -650,6 +669,20 @@ func (s *StateDB) getDeletedStateObject(addr common.Address) *StateObject {
 			}
 		}
 	}
+
+	// ErrSnapshotStale may occur due to diff layers in the update, so we should try again in noTrie mode.
+	if s.NoTrie() && err != nil && errors.Is(err, snapshot.ErrSnapshotStale) {
```
yutianwu, line 674: for other errors, do we need to handle too? since trie is no enabled, we should not go further if there is an error

yutianwu: since we can not regenerate snapshot if we don't have trie, should we add a doc for handling this issue

core/state/state_object.go
```diff
@@ -436,6 +452,13 @@ func (s *StateObject) updateTrie(db Database) Trie {
 
 // UpdateRoot sets the trie root to the current root hash of
 func (s *StateObject) updateRoot(db Database) {
+	// If node runs in no trie mode, set root to empty.
+	defer func() {
+		if db.NoTries() {
+			s.data.Root = common.Hash{}
+		}
+	}()
```
yutianwu, line 460: should we call `updateTrie` below if trie is not enabled

core/state/state_object.go
```diff
@@ -274,6 +278,18 @@ func (s *StateObject) GetCommittedState(db Database, key common.Hash) common.Has
...
+		// This error occurs when statedb.snaps.Cap changes the state of the merged difflayer
+		// to stale during the refresh of the difflayer, indicating that it is about to be discarded.
+		// Since the difflayer is refreshed in parallel,
+		// there is a small chance that the difflayer of the stale will be read while reading,
+		// resulting in an empty array being returned here.
+		// Therefore, noTrie mode must retry here,
+		// and add a time interval when retrying to avoid stacking too much and causing stack overflow.
+		time.Sleep(snapshotStaleRetryInterval)
+		return s.GetCommittedState(db, key)
+	}
```
yutianwu, line 292: also here, what if the err is not `ErrSnapshotStale`, should we panic

guagualvcha (approved):

core/state/statedb.go
dean65: fixed

eth/ethconfig/gen_config.go
dean65: This is a previously missing field

core/state/state_object.go
dean65: This is a deferred function that normally executes `updateTrie` below. If `NoTries` is enabled, root is set to `empty`

core/state/statedb.go
dean65: I think if get other errors, we should keep the original logic and let the block re-import

core/state/state_object.go
dean65: I think if you get other errors, we should keep the original logic and let the block re-import

guagualvcha (dismissed):

cmd/utils/flags.go
```diff
@@ -1822,6 +1858,7 @@ func SetDNSDiscoveryDefaults(cfg *ethconfig.Config, genesis common.Hash) {
 	if url := params.KnownDNSNetwork(genesis, protocol); url != "" {
 		cfg.EthDiscoveryURLs = []string{url}
 		cfg.SnapDiscoveryURLs = cfg.EthDiscoveryURLs
+		cfg.TrustDiscoveryURLs = cfg.EthDiscoveryURLs
```
brilliant-lx, line 1861: just wonder: `TrustDiscoveryURLs` use as same configure field as `SnapDiscoveryURLs`

core/block_validator.go
```diff
@@ -30,23 +30,38 @@ import (
 
 const badBlockCacheExpire = 30 * time.Second
 
+type BlockValidatorOption func(*BlockValidator) *BlockValidator
+
+func EnableRemoteVerifyManager(remoteValidator *remoteVerifyManager) BlockValidatorOption {
+	return func(bv *BlockValidator) *BlockValidator {
+		bv.remoteValidator = remoteValidator
+		return bv
```
brilliant-lx, line 38: Is it a bit weird to return `bv`, since it is passed in as self pointer?

core/blockchain.go
```diff
@@ -499,11 +510,31 @@ func (bc *BlockChain) cacheReceipts(hash common.Hash, receipts types.Receipts) {
 	bc.receiptsCache.Add(hash, receipts)
 }
 
-func (bc *BlockChain) cacheDiffLayer(diffLayer *types.DiffLayer) {
+func (bc *BlockChain) cacheDiffLayer(diffLayer *types.DiffLayer, diffLayerCh chan struct{}) {
+	sort.SliceStable(diffLayer.Codes, func(i, j int) bool {
```
brilliant-lx, line 514: dd some comments to explain why do sorting here?

core/blockchain.go
```diff
@@ -2119,6 +2163,9 @@ func (bc *BlockChain) insertChain(chain types.Blocks, verifySeals bool) (int, er
 		if err != nil {
 			return it.index, err
 		}
+		if statedb.NoTrie() {
+			statedb.SetExpectedStateRoot(block.Root())
+		}
```
brilliant-lx, line 2168: The bellow line#2185 set state root too, duplicated?
statedb.SetExpectedStateRoot(block.Root())

KeefeL (approved):

cmd/utils/flags.go
dean65: yes, they use the same configure

core/block_validator.go
dean65: I just want to keep the same style as the `BlockChainOption` function

core/blockchain.go
dean65: added

core/blockchain.go
dean65: yes, it was a duplicate, i fixed

core/blockchain.go
```diff
@@ -3100,19 +3158,138 @@ func (bc *BlockChain) SubscribeBlockProcessingEvent(ch chan<- bool) event.Subscr
...
+		Storages:  d.Storages,
+	}
+
+	for index, account := range diff.Accounts {
+		full, err := snapshot.FullAccount(account.Blob)
+		if err != nil {
+			return common.Hash{}, fmt.Errorf("decode full account error: %v", err)
+		}
+		// set account root to empty root
+		diff.Accounts[index].Blob = snapshot.SlimAccountRLP(full.Nonce, full.Balance, common.Hash{}, full.CodeHash)
```
brilliant-lx, line 3278: Should we use `emptyRoot` instead of `common.Hash{}` for `SlimAccountRLP`?

brilliant-lx: oh, it is ok anyway, the result is same.

brilliant-lx: oh, it is ok anyway, the result is same.

core/state/state_object.go
```diff
@@ -274,6 +278,18 @@ func (s *StateObject) GetCommittedState(db Database, key common.Hash) common.Has
...
+	if s.db.NoTrie() && err != nil && errors.Is(err, snapshot.ErrSnapshotStale) {
+		// This error occurs when statedb.snaps.Cap changes the state of the merged difflayer
+		// to stale during the refresh of the difflayer, indicating that it is about to be discarded.
+		// Since the difflayer is refreshed in parallel,
+		// there is a small chance that the difflayer of the stale will be read while reading,
+		// resulting in an empty array being returned here.
+		// Therefore, noTrie mode must retry here,
+		// and add a time interval when retrying to avoid stacking too much and causing stack overflow.
+		time.Sleep(snapshotStaleRetryInterval)
+		return s.GetCommittedState(db, key)
```
brilliant-lx, line 291: This is a recursively call, so if snapshot is stale, we will keep trying until we get it?

core/state/state_object.go
brilliant-lx: I guess you try to keep the behavior unchanged, by using the defer(){...}, right?
It is ok to me, since the current logic of `updateRoot` is a bit completed.

setunapo (approved):

brilliant-lx (approved):

guagualvcha (approved):

PR #932, opened 2022-05-29
[R4R] fix pipecommit issue

forcodedancing:
### Description

This pr will fix an issue, i.e., some account root is not correctly set when committing mpt ,during pipecommit.

### Rationale

During pipecommit, the account root is populated as `dummyRoot` and will be corrected at several places. However, the previous code will skip some accounts during correction stage.  This will lead to commit error when committing mpt.
This fix will correct the root for `stateOjbect`, which's account root length is zero.

### Example

NA

### Changes

Notable changes: 
* change CorrectAccountRoots function

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -900,9 +900,9 @@ func (s *StateDB) GetRefund() uint64 {
 	return s.refund
 }
 
-// GetRefund returns the current value of the refund counter.
+// WaitPipeVerification waits until the snapshot been verified
 func (s *StateDB) WaitPipeVerification() error {
-	// We need wait for the parent trie to commit
+	// Need to wait for the parent trie to commit
 	if s.snap != nil {
 		if valid := s.snap.WaitAndGetVerifyRes(); !valid {
 			return fmt.Errorf("verification on parent snap failed")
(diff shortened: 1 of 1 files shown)


guagualvcha (approved):

guagualvcha (changes requested):
It will keep as dummy root?

guagualvcha (dismissed):

yutianwu (approved):

core/state/statedb.go
```diff
@@ -989,7 +989,7 @@ func (s *StateDB) CorrectAccountsRoot(blockRoot common.Hash) {
 	if accounts, err := snapshot.Accounts(); err == nil && accounts != nil {
 		for _, obj := range s.stateObjects {
 			if !obj.deleted && !obj.rootCorrected && obj.data.Root == dummyRoot {
-				if account, exist := accounts[crypto.Keccak256Hash(obj.address[:])]; exist && len(account.Root) != 0 {
+				if account, exist := accounts[crypto.Keccak256Hash(obj.address[:])]; exist {
```
setunapo, line 992: This is the only place that changes the behavior, others changes in this PR is just code refine.
But how could `len(account.Root) == 0` ?

setunapo:
Another suggestion, commit description should add more detail to tell how it is fixed, better to avoid `fix a bug`

setunapo:
Another suggestion, commit description should add more detail to tell how it is fixed, better to avoid `fix a bug`

forcodedancing:
> Another suggestion, commit description should add more detail to tell how it is fixed, better to avoid `fix a bug`

sure, let me refine it.

forcodedancing:
> Another suggestion, commit description should add more detail to tell how it is fixed, better to avoid `fix a bug`

sure, let me refine it.

setunapo (dismissed):

yutianwu (approved):

setunapo (approved):

guagualvcha (approved):

PR #936, opened 2022-05-31
BEP-126: Introduce Fast Finality Mechanism

realuncle:
## Summary
This PR tries to implement fast finality mechanism introduced by [BEP-126](https://github.com/bnb-chain/BEPs/pull/126). 

## Abstract
BEP-126 Proposal describes a fast finality mechanism to finalize a block, once the block has been finalized, it won't be
reverted forever.

It takes several steps to finalize a block:

1. A block is proposed by a validator and propagated to other validators.
2. Validators use their BLS private key to sign for the block as a vote message.
3. Gather the votes from validators into a pool.
4. Aggregate the BLS signature if its direct parent block has gotten enough votes when proposing a new block.
5. Set the aggregated vote attestation into the extra field of the new block's header.
6. Validators and full nodes who received the new block with the direct parent block's attestation can justify the direct parent block.
7. If there are two continuous blocks have been justified, then the previous one is finalized.

The finality of a block can be achieved within two blocks in most cases, this is expected to reduce the chance
of chain re-organization and stabilize the block producing further.

## Motivation
Finality is critical for blockchain security, once the block is finalized, it would not be reverted anymore. The fast
finality feature is very useful, users can make sure they get the accurate information from the latest finalized block,
then they can decide what to do next instantly.

Currently, on BNB Smart chain, all the full nodes and validators need to wait until enough blocks have been produced
to ensure a probabilistic finality. According to the whitepaper, with 21 validators, full nodes and validators can
wait ⅔*21=14 blocks to ensure a relatively secure finality, it would be quite long time for some critical applications.

## Specification
The vote message is signed by validators’ private key through the BLS algorithm, however the BLS private and public keys are different from traditional BSC/Ethereum account’s private and public keys, so use a new key manager to manage BLS related keys.

In order to collect the vote information, we use another vote pool to gather the votes from validators, this vote pool is similar to txpool but independent.

Validators can vote for blocks if they think that block should be in the canonical chain, and once their votes have been wrapped into the header, they will get rewards. At the same time, they should obey the vote rules, the vote rules can be described as follows.

A validator must not publish two distinct votes for the same height. (Rule 1)
A validator must not vote within the span of its other votes . (Rule 2)
Validators always vote for their canonical chain’s latest block. (Rule 3)

Validators can vote for blocks they received under the above three rules. Once the parent block’s votes in the vote pool are more than ⅔+ validators, the validator  will aggregate the votes message and write them to the proposed block’s header when it proposes new blocks.

The block receivers, either full nodes or non-proposer validators, download the block from p2p node, verify the votes from valid validators and verify the BLS aggregated signature if exist, insert the block into the blockchain, write the new attestation info to the consensus snapshot, distribute reward in finalize function if reward requirements meet. 

The slash related function is implemented in a stand alone slash contract, anyone can get a vote message from the vote pool, once they find the malicious behavior, they can submit the assembled evidence to slash contract, and slash contract will do the detailed verification and slash actions. The evidence format, verification procedure and slash actions will be designed in slash component design.

### BLS and BLS Key Manager
BLS is used to sign the vote for fast finality, aggregate the votes from different validators, verify the signature from different validators and verify the aggregated signatures,  in order to format the vote data, in our design, we sign the hash of the vote data, not the data itself.

Since the BLS private key and public key are different from current Ethereum accounts’ key, we introduce another key manager to manage the BLS private key and public key. The BLS key manager is designed to store the BLS private key, since the private key shouldn’t be stored as plain text to avoid leaking, we introduce BLS wallet, BLS key manager and BLS accounts here. Each account is a pair of BLS private key and BLS public key, it will be stored as an encrypted keystore Json file. There will be a singleton wallet while running the Geth client, the wallet can manage several BLS accounts, actually, we just need one account in general. Users can create a new account or import existing Json keystore accounts.

The BLS wallet is bound with a singleton key manager, through the key manager, we can get all the public keys in the BLS wallet, and  once the validator needs to sign a vote, call the key manager’s sign method. 

### Vote Pool
The votepool is used to gather votes from different validators, propagating votes to other nodes. All active validators can vote for the fast finality of blocks under the above three rules, the vote data will be **{SourceNumber, SourceHash, TargetNumber, TargetHash}**.  Validators use their BLS private key to sign the hash of the vote data, so the vote message is **{BLSPubKey, {SourceNumber, SourceHash, TargetNumber, TargetHash}, Signature}**, once the receivers see the vote message, they can verify it easily.

As we know propagating messages is the duty of P2P protocols, in order to propagate the vote message, we will define new message types based on current ETH67 protocol.

In order to finalize blocks, the miner should assemble the vote messages for the direct parent block once its valid votes have more than ⅔+ validators, and write them to the block header as an attestation. The assembled signature should be **{ValidatorsIndexBitmap, {SourceNumber, SourceHash, TargetNumber, TargetHash}, AggregatedSignature}**, the bitmap indicates the validators that voted for the block, aggregate their vote signatures into aggregated signature, the aggregated signature is the same length with each common signature.

Once the validators or full nodes receive a block, they should verify the block header, for this fast finality feature, they should verify the assembled attestation information.
As the consensus engine can get validators’ information from the snapshot, the BLS public key should be recorded as a part of validator, through the ValidatorsIndexBitmap we can easily get their BLS public keys, then we can use the BLS FastAggregateVerify function to verify the aggregated signature.

### Reward
Distributing reward should be a part of the consensus engine, in order to make the consensus lighter, we would distribute the reward by batch, once an epoch. As the vote attestation has been recorded in the headers, so we can iterate the headers from the first block to the last block of the previous epoch, find out the attested blocks and which validators have voted for them.

Through the iteration, we can get two arrays {validators[], accumulatedWeights[]}, then call the distributeFinalityReward method of ValidatorSet contract and pass these two parameters.

The total reward for validators of this epoch would be a part balance of the SystemReward contract, the partial would be set to 10% initially and can be governed. Once we get the total reward, then distribute the total reward to each validator by the accumulatedWeights.

### Slash
Anyone can get validators’ vote message from vote pool, if they find some validators’ vote message violate one of the rules described in BEP-126, they can assemble the evidence and submit to the slash contract, slash contract will verify the submitted evidence, once the evidence has been verified, the malicious validator will be slashed and the submitter will be rewarded.


Mister-EA:
> Validators can vote for blocks they received under the above three rules. 

Which rules?

core/vm/contracts.go
```diff
@@ -1110,3 +1138,72 @@ func (c *bls12381MapG2) Run(input []byte) ([]byte, error) {
...
+
+	if pubKeyNumber > 1 {
+		if !sig.FastAggregateVerify(pubKeys, msg) {
+			return big0.Bytes(), nil
+		}
+	} else {
+		if !sig.Verify(pubKeys[0], msgBytes) {
+			return big0.Bytes(), nil
+		}
+	}
```
KeefeL, line 1206: if pubKeyNumber > 1 {
		if !sig.FastAggregateVerify(pubKeys, msg) {
			return nil, fmt.Errorf("signature verify failed")
		}
	} else {
		if !sig.Verify(pubKeys[0], msgBytes) {
			return nil, fmt.Errorf("signature verify failed")
		}
	}

	return big1.Bytes(), nil

buddh0: 1. first, fail to verify is a normal case,
    just return 'false' is enough, big0.Bytes() meaning 'false'
2. returning errors excluding `ErrExecutionReverted`
    will be ignored outside and swallow all the gas, not expected.
    
    
image

image

buddh0:
> > Validators can vote for blocks they received under the above three rules.
> 
> Which rules?

doc updated

buddh0 (dismissed):

common/math/integer.go
```diff
@@ -96,3 +96,7 @@ func SafeMul(x, y uint64) (uint64, bool) {
 	hi, lo := bits.Mul64(x, y)
 	return lo, hi != 0
 }
+
+func CeilDiv(x, y int) int {
+	return (x + y - 1) / y
```
guagualvcha, line 101: ensure `y` is not zero?

consensus/parlia/abi.go
```diff
@@ -1,1525 +1,3564 @@
 package parlia
 
+const validatorSetABIBeforeBoneh = `
+[
```
guagualvcha, line 4: The `abi` here needs to be regenerated according to the latest version.  Please confirm.

consensus/parlia/parlia_test.go
```diff
@@ -144,3 +153,449 @@ func randomAddress() common.Address {
...
+		validatorSet: validatorSet,
+		head:         GenesisBlock,
+		voteRecords:  make(map[uint64]*types.VoteData),
+	}
+	return v
+}
+
+func (v *MockValidator) SignRecently() bool {
+	parent := v.head
+	for i := 0; i < v.validatorSet*2/3; i++ {
```
guagualvcha, line 267: I believe the testcase here is out of date, can you refresh it?

consensus/parlia/parlia.go
```diff
@@ -53,13 +56,15 @@ const (
 	extraSeal        = 65 // Fixed number of extra-data suffix bytes reserved for signer seal
 	nextForkHashSize = 4  // Fixed number of extra-data suffix bytes reserved for nextForkHash.
 
-	validatorBytesLength = common.AddressLength
-	wiggleTime           = uint64(1) // second, Random delay (per signer) to allow concurrent signers
-	initialBackOffTime   = uint64(1) // second
-	processBackOffTime   = uint64(1) // second
+	validatorBytesLength           = common.AddressLength
```
guagualvcha, line 59: `validatorBytesLength` ==> `validatorBytesLengthBeforeBoneh`

`validatorBytesLengthAfterBoneh` ==> `validatorBytesLength`.

This fits out code style.

consensus/parlia/parlia.go
```diff
@@ -426,11 +592,22 @@ func (p *Parlia) verifyCascadingFields(chain consensus.ChainHeaderReader, header
 		return fmt.Errorf("invalid gas limit: have %d, want %d += %d", header.GasLimit, parent.GasLimit, limit)
 	}
 
+	// Verify vote attestation for fast finality.
+	if err := p.verifyVoteAttestation(chain, header, parents); err != nil {
+		if chain.Config().IsLynn(header.Number) {
+			return err
+		}
+		log.Warn("Verify vote attestation failed", "block", header.Number.Uint64(), "block hash", header.Hash(), "error", err)
```
guagualvcha, line 600: Can you log the whole header here? It will help to locate the issue.

consensus/parlia/parlia.go
```diff
@@ -595,6 +776,109 @@ func (p *Parlia) verifySeal(chain consensus.ChainHeaderReader, header *types.Hea
 	return nil
 }
 
+func (p *Parlia) prepareValidators(chain consensus.ChainHeaderReader, header *types.Header) error {
```
guagualvcha, line 779: `chain` here seems useless.

consensus/parlia/parlia.go
```diff
@@ -884,6 +1310,22 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 			return
 		case <-time.After(delay):
 		}
+
+		err := p.assembleVoteAttestation(chain, header)
```
guagualvcha, line 1314: There is a change in miner module recently that the miner will try mine blocks for multiple times during 3 seconds,  so not sure how this impact the logic here, please take care during testing.

common/math/integer.go
NathanBSC: accept

consensus/parlia/parlia_test.go
NathanBSC: ok

consensus/parlia/abi.go
NathanBSC: ok

consensus/parlia/parlia.go
NathanBSC: accept

consensus/parlia/parlia.go
NathanBSC: ok

consensus/parlia/parlia.go
NathanBSC: early broadcast not on master branch, have been revert
so need to take care during testing when importing early broadcast feature again

consensus/parlia/parlia.go
NathanBSC: ok

consensus/parlia/parlia_test.go
NathanBSC: these cases have been refreshed after changing new forkchoice
so only this code need to refresh caused by separating chain liveness from finality liveness

core/vote/vote_manager.go
```diff
@@ -0,0 +1,219 @@
...
+	// Create voteJournal
+	voteJournal, err := NewVoteJournal(journalPath)
+	if err != nil {
+		return nil, err
+	}
+	log.Info("Create voteJournal successfully")
+	voteManager.journal = voteJournal
+
+	// Subscribe to chain head event.
+	voteManager.chainHeadSub = voteManager.chain.SubscribeChainHeadEvent(voteManager.chainHeadCh)
```
guagualvcha, line 63: since we use a lot of `Subscribe` here, please take care of graceful shutdown, incase send object a closed channel or read from a closed channel, which may cause panic.

NathanBSC: I will review this  theme  related

core/blockchain.go
```diff
@@ -1203,7 +1241,7 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 
 		// Rewind may have occurred, skip in that case.
 		if bc.CurrentHeader().Number.Cmp(head.Number()) >= 0 {
-			reorg, err := bc.forker.ReorgNeeded(bc.CurrentFastBlock().Header(), head.Header())
+			reorg, err := bc.forker.ReorgNeededWithFastFinality(bc.CurrentFastBlock().Header(), head.Header())
```
Mister-EA, line 1244: Based on the `TestChainConfig` this code will be called with an underlying ethash engine. How will the new functionality actually be tested then?

core/blockchain_reader.go
```diff
@@ -42,6 +42,36 @@ func (bc *BlockChain) CurrentBlock() *types.Block {
 	return bc.currentBlock.Load().(*types.Block)
 }
 
+// CurrentFinalBlock retrieves the current finalized block of the canonical
+// chain. The block is retrieved from the blockchain's internal cache.
+func (bc *BlockChain) CurrentFinalBlock() *types.Header {
+	if p, ok := bc.engine.(consensus.PoSA); ok {
```
Mister-EA, line 48: Won't be covered by tests due to ethash engine in `TestChainConfig`.

consensus/parlia/parlia_test.go
```diff
@@ -144,3 +153,449 @@ func randomAddress() common.Address {
...
+		for _, val := range c.validators {
+			t.Logf("[Testcase %d] validator(%d) head block: %d",
+				index, val.index, val.head.blockNumber)
+			t.Logf("[Testcase %d] validator(%d) highest justified block: %d",
+				index, val.index, val.head.GetJustifiedNumber())
+			t.Logf("[Testcase %d] validator(%d) highest finalized block: %d",
+				index, val.index, val.head.GetFinalizedBlock().blockNumber)
+		}
+
+		if c.CheckChain() == false {
```
Mister-EA, line 597: What about tests where consensus is expected to fail?

core/forkchoice.go
```diff
@@ -106,3 +113,24 @@ func (f *ForkChoice) ReorgNeeded(current *types.Header, header *types.Header) (b
...
 }
+
+// ReorgNeededWithFastFinality compares justified block numbers firstly, backoff to compare tds when equal
+func (f *ForkChoice) ReorgNeededWithFastFinality(current *types.Header, header *types.Header) (bool, error) {
+	_, ok := f.chain.Engine().(consensus.PoSA)
+	if !ok {
+		return f.reorgNeeded(current, header)
+	}
+
+	justifiedNumber, curJustifiedNumber := uint64(0), uint64(0)
```
Mister-EA, line 124: This part won't be covered by  tests.

guagualvcha (dismissed):

guagualvcha (approved):

NathanBSC (approved):

PR #952, opened 2022-06-20
[R4R][Feature]: Improve trie prefetch

setunapo:
### Description
We found that the current trie prefetch could be not fast enough, especially trie prefetch of the outer big state trie tree. When processing a big block, thousands of accounts's root hash could be changed. In this PR, we try to optimize it in 2 approaches.

#### 1.Trie prefetch From/To address in advance
Instead of do trie prefetch until a transaction is finalized, we could do trie prefetch in advance. Try to prefetch the trie node of the From/To accounts, since their root hash are most likely to be changed.
The bellow tracing image shows, the trieprefetch has several idle state.
image
After trie prefetch From/To in advance, it will be busy.
image
#### 2.Parallel TriePrefetch
Currently, we create a subfetcher for each account address to do trie prefetch. But if the number of changed account or updated KV is too large, the subfetcher would not be able to fetch all the trie node  in time. In this case, we will create child subfetchers to do parallel trie prefetch for it.
Child subfetcher will be created on demand, when the pending task size exceed the threshold of `parallelTriePrefetchThreshold` and each has a capacity of `parallelTriePrefetchCapacity`. The created child subfetchers can be reused, we will try to dispatch tasks to these hungry child first before start new child subfecther. If the parent subfetcher is stopped, the child subfecthers will be stopped too.
The general workflow of parallel trie prefetch is shown by the following diagram.
image

### Performance Result
This PR would be more beneficial if the validation phase occupies more time. 
I tested based on BNB chain traffics of April 18th 2022, during that time, the peek validation could cost ~1.65 second on my test device. The hardware configuration of the device is: `4TB SSD, 16 core, 60GB memory`.
And I tested for different configuration of `parallelTriePrefetchThreshold` and `parallelTriePrefetchCapacity`, the results of peek insert and validation cost are:
**peek insert**
- v1.1.11:   ~2.30s
- this PR with threshold == 100 capacity == 200 : ~1.65s ==> improved: ~28%
- this PR with threshold == 100 capacity == 100 : ~1.48s ==> improved: ~36%
- this PR with threshold == 50  capacity == 100 : ~1.34s  ==> improved: ~42%
- this PR with threshold == 50  capacity == 50 : ~1.23s ==> improved: ~47%
- this PR with threshold == 25  capacity == 50 : ~1.13s  ==> improved: ~51%
- this PR with threshold == 25 capacity == 25 : ~1.05s ==> improved: ~54%
- this PR with threshold == 10 capacity == 20 : ~0.92s  ==> improved: ~60%
- this PR with threshold == 10 capacity == 10 : ~0.94s ==> improved: ~59%
- this PR with threshold == 5 capacity == 10 : ~0.87s  ==> improved: ~62%

**peek validation**
- v1.1.11:   ~1.65s
- this PR with threshold == 100 capacity == 200 : ~1.10s ==> improved: ~33%
- this PR with threshold == 100 capacity == 100 : ~0.85s ==> improved: ~48%
- this PR with threshold == 50 capacity == 100 : ~0.74s  ==> improved: ~55%
- this PR with threshold == 50 capacity == 50 : ~0.60s ==> improved: ~64%
- this PR with threshold == 25 capacity == 50 : ~0.52s ==> improved: ~68%
- this PR with threshold == 25 capacity == 25 : ~0.41s ==> improved: ~75%
- this PR with threshold == 10 capacity == 20 : ~0.33s ==> improved: ~80%
- this PR with threshold == 10 capacity == 10 : ~0.30s ==> improved: ~82%
- this PR with threshold == 5 capacity == 10 : ~0.26s ==> improved: ~84%

It seems `parallelTriePrefetchThreshold = 10` and `parallelTriePrefetchCapacity = 20` could be good enough

I also did a long run test for more the 14hours, the PR even have observable benefit for general traffic load, I saw ~60% reduce of the validation cost. 
image

### Changes
No impact to users.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -237,6 +237,9 @@ func (s *StateDB) StopPrefetcher() {
 }
 
 func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer) {
+	if s.prefetcher == nil {
+		return
+	}
 	accounts := make(map[common.Address]struct{}, block.Transactions().Len()<<1)
 	for _, tx := range block.Transactions() {
 		from, err := types.Sender(signer, tx)
@@ -254,7 +257,7 @@ func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer)
 		addressesToPrefetch = append(addressesToPrefetch, common.CopyBytes(addr[:])) // Copy needed for closure
 	}
 
-	if s.prefetcher != nil && len(addressesToPrefetch) > 0 {
+	if len(addressesToPrefetch) > 0 {
 		s.prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
 	}
 }
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -237,7 +237,11 @@ func (s *StateDB) StopPrefetcher() {
 }
 
 func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer) {
-	if s.prefetcher == nil {
+	s.prefetcherLock.Lock()
+	prefetcher := s.prefetcher // s.prefetcher could be resetted to nil
+	s.prefetcherLock.Unlock()
+
+	if prefetcher == nil {
 		return
 	}
 	accounts := make(map[common.Address]struct{}, block.Transactions().Len()<<1)
@@ -258,7 +262,7 @@ func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer)
 	}
 
 	if len(addressesToPrefetch) > 0 {
-		s.prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
+		prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
 	}
 }
 
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -236,6 +236,29 @@ func (s *StateDB) StopPrefetcher() {
 	}
 }
 
+func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer) {
+	accounts := make(map[common.Address]struct{}, block.Transactions().Len()<<1)
+	for _, tx := range block.Transactions() {
+		from, err := types.Sender(signer, tx)
+		if err != nil {
+			// invalid block, skip prefetch
+			return
+		}
+		accounts[from] = struct{}{}
+		if tx.To() != nil {
+			accounts[*tx.To()] = struct{}{}
+		}
+	}
+	addressesToPrefetch := make([][]byte, 0, len(accounts))
+	for addr := range accounts {
+		addressesToPrefetch = append(addressesToPrefetch, common.CopyBytes(addr[:])) // Copy needed for closure
+	}
+
+	if s.prefetcher != nil && len(addressesToPrefetch) > 0 {
+		s.prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
+	}
+}
+
 // Mark that the block is processed by diff layer
 func (s *StateDB) SetExpectedStateRoot(root common.Hash) {
 	s.expectedRoot = root
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -28,8 +28,8 @@ import (
 const (
 	abortChanSize                 = 64
 	concurrentChanSize            = 10
-	parallelTriePrefetchThreshold = 50
-	parallelTriePrefetchCapacity  = 100
+	parallelTriePrefetchThreshold = 100
+	parallelTriePrefetchCapacity  = 200
 )
 
 var (
@@ -115,13 +115,19 @@ func (p *triePrefetcher) mainLoop() {
 				p.fetchersMutex.Unlock()
 			}
 			fetcher.schedule(pMsg.keys)
+
+			// fecther could be interrupted by call to trie() or close()
+			if fetcher.interrupted {
+				continue
+			}
 			// no need to run parallel trie prefetch if threshold is not reached.
-			if fetcher.pendingSize() > parallelTriePrefetchThreshold {
+			if atomic.LoadUint32(&fetcher.pendingSize) > parallelTriePrefetchThreshold {
 				fetcher.scheduleParallel(pMsg.keys)
 			}
 
 		case <-p.closeMainChan:
 			for _, fetcher := range p.fetchers {
+				fetcher.interrupted = true
 				p.abortChan <- fetcher // safe to do multiple times
 				<-fetcher.term
 				if metrics.EnabledExpensive {
@@ -173,13 +179,10 @@ func (p *triePrefetcher) abortLoop() {
 		case fetcher := <-p.abortChan:
 			fetcher.abort()
 			// stop fetcher's parallel children
-			fetcher.childrenLock.Lock()
-			children := fetcher.children
-			fetcher.children = nil
-			fetcher.childrenLock.Unlock()
-			for _, child := range children {
+			for _, child := range fetcher.paraChildren {
 				child.abort()
 			}
+			fetcher.paraChildren = nil
 		case <-p.closeAbortChan:
 			return
 		}
@@ -276,6 +279,7 @@ func (p *triePrefetcher) trie(root common.Hash) Trie {
 		p.deliveryMissMeter.Mark(1)
 		return nil
 	}
+	fetcher.interrupted = true
 
 	// Interrupt the prefetcher if it's by any chance still running and return
 	// a copy of any pre-loaded trie.
@@ -324,10 +328,8 @@ type subfetcher struct {
 	root common.Hash // Root hash of the trie to prefetch
 	trie Trie        // Trie being populated with nodes
 
-	tasks          [][]byte   // Items queued up for retrieval
-	lock           sync.Mutex // Lock protecting the task queue
-	totalSize      uint32
-	processedIndex uint32
+	tasks [][]byte   // Items queued up for retrieval
+	lock  sync.Mutex // Lock protecting the task queue
 
 	wake chan struct{}  // Wake channel if a new task is scheduled
 	stop chan struct{}  // Channel to interrupt processing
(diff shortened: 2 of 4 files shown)


core/state_processor.go
```diff
@@ -401,10 +401,13 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	posa, isPoSA := p.engine.(consensus.PoSA)
 	commonTxs := make([]*types.Transaction, 0, txNum)
 
-	// initilise bloom processors
+	// initialise bloom processors
 	bloomProcessors := NewAsyncReceiptBloomGenerator(txNum)
 	statedb.MarkFullProcessed()
 
+	// do trie prefetch for the big state trie tree in advance based transaction's From/To address.
+	statedb.TriePrefetchInAdvance(block, signer)
```
guagualvcha, line 409: Can you try to make it a asynchronous go routine?

setunapo: triePrefetcher's API is not thread safe, should be called in a same routine.

core/state/statedb.go
```diff
@@ -222,6 +222,29 @@ func (s *StateDB) StopPrefetcher() {
...
+		if tx.To() != nil {
+			accounts[*tx.To()] = struct{}{}
+		}
+	}
+	addressesToPrefetch := make([][]byte, 0, len(accounts))
+	for addr := range accounts {
+		addressesToPrefetch = append(addressesToPrefetch, common.CopyBytes(addr[:])) // Copy needed for closure
+	}
+
+	if s.prefetcher != nil && len(addressesToPrefetch) > 0 {
```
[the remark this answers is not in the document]
setunapo: accept, will be updated

core/state_processor.go
guagualvcha: That is a problem then, any proof that it is worthy to introduce it in the critical path?

core/state/statedb.go
```diff
@@ -222,6 +222,32 @@ func (s *StateDB) StopPrefetcher() {
...
+			accounts[*tx.To()] = struct{}{}
+		}
+	}
+	addressesToPrefetch := make([][]byte, 0, len(accounts))
+	for addr := range accounts {
+		addressesToPrefetch = append(addressesToPrefetch, common.CopyBytes(addr[:])) // Copy needed for closure
+	}
+
+	if len(addressesToPrefetch) > 0 {
+		s.prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
```
guagualvcha, line 247: We can do asynchronous execution here. Since `addressesToPrefetch` are all unique, so it is thread safe.

core/state/trie_prefetcher.go
```diff
@@ -18,13 +18,18 @@ package state
...
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/metrics"
 )
 
-const abortChanSize = 64
+const (
+	abortChanSize                 = 64
+	parallelTriePrefetchThreshold = 100
+	parallelTriePrefetchCapacity  = 200
```
guagualvcha, line 31: Will `parallelTriePrefetchCapacity` be too big? How to measure whether it is a reasonable value?

core/state/trie_prefetcher.go
```diff
@@ -266,12 +298,52 @@ func (sf *subfetcher) schedule(keys [][]byte) {
...
+	var keyIndex uint32 = 0
+	childrenNum := len(sf.paraChildren)
+	if childrenNum > 0 {
+		// To feed the children first, if they are hungry.
+		// A child can handle keys with capacity of parallelTriePrefetchCapacity.
+		startIndex := len(keys) % childrenNum // randomly select the start child to avoid always feed the first one
+		for i := 0; i < childrenNum; i++ {
+			child := sf.paraChildren[startIndex]
+			startIndex = (startIndex + 1) % childrenNum
+			feedNum := parallelTriePrefetchCapacity - atomic.LoadUint32(&child.pendingSize)
```
guagualvcha, line 319: I am a bit worried about uint under-flow, let us protect it anyway.

guagualvcha: like  `if parallelTriePrefetchCapacity <=  atomic.LoadUint32(&child.pendingSize) { return }`

core/state/trie_prefetcher.go
```diff
@@ -241,6 +269,10 @@ type subfetcher struct {
 	used [][]byte            // Tracks the entries used in the end
 
 	accountHash common.Hash
+
+	interrupted  bool
```
guagualvcha, line 273: Do you think any need for this `interrupted` to be atomic?

guagualvcha: I am not sure, writing a bool var should be atomic, but in multi-core CPU framework, each core does have its own cache, so concurrent read-write may happen, but it won't panic, I am not sure what is the best practice.

core/state_processor.go
setunapo: Good point, we will consider to redesigned the current triePrefetch to make `prefetch` concurrent safe.
I will hold this PR until the new tirePrefetch is ready

core/state/statedb.go
setunapo: may not necessary, `addressesToPrefetch` is used as `tasks`, it will be dispatched to the loop as a unit.

core/state/trie_prefetcher.go
setunapo: I tried 50, 100 before, but smaller capacity could cause more child goroutine to be created.
Since there are already lots of trie prefetch routine right now, I would prefer to limit the child goroutine number.
The PR is mainly for handling big blocks, 200 could be enough. But it is ok, I will double confirm,

core/state/trie_prefetcher.go
setunapo: Accepted, will be updated.

core/state/trie_prefetcher.go
setunapo: Yes, boolean or integer value should be used with Atomic for concurrent access, otherwise the result will depend on platform.
But in this case, I deliberately not use atomic to protect, since `interrupted` will not be concurrently accessed, as currently `triePrefetch` API requests, `close()`, `trie()` and `prefetch()` have to be called within a same routine.

core/state_processor.go
setunapo: updated, pls check

core/state/trie_prefetcher.go
setunapo: `interrupted` is removed now, subfetcher's lifecycle is maintained in mainLoop now, no longer needed

core/state/trie_prefetcher.go
```diff
@@ -348,12 +366,51 @@ func (sf *subfetcher) schedule(keys [][]byte) {
...
+			if keyIndex+feedNum >= uint32(len(keys)) {
+				// the new arrived keys are all consumed by children.
+				child.schedule(keys[keyIndex:])
+				return
+			}
+			child.schedule(keys[keyIndex : keyIndex+feedNum])
+			keyIndex += feedNum
+		}
+	}
+	// Children did not comsume all the keys, to create new subfetch to handle left keys.
```
guagualvcha, line 401: `comsume` ==> `consume`

core/state_processor.go
```diff
@@ -403,9 +403,15 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 	posa, isPoSA := p.engine.(consensus.PoSA)
 	commonTxs := make([]*types.Transaction, 0, txNum)
 
-	// initilise bloom processors
+	// initialise bloom processors
 	bloomProcessors := NewAsyncReceiptBloomGenerator(txNum)
 	statedb.MarkFullProcessed()
+	signer := types.MakeSigner(p.config, header.Number)
+	// do trie prefetch for the big state trie tree in advance based transaction's From/To address.
+	go func() {
```
guagualvcha, line 411: I think it is not proper to do prefetch in `state_processor`,  it is better to follow the same style as snap prefetcher: do it on `insertChain`.

core/state/trie_prefetcher.go
setunapo: thx, updated.

guagualvcha (approved):

yutianwu (approved):

core/state/trie_prefetcher.go
```diff
@@ -348,12 +366,51 @@ func (sf *subfetcher) schedule(keys [][]byte) {
...
+			}
+			child.schedule(keys[keyIndex : keyIndex+feedNum])
+			keyIndex += feedNum
+		}
+	}
+	// Children did not consume all the keys, to create new subfetch to handle left keys.
+	keysLeft := keys[keyIndex:]
+	keysLeftSize := len(keysLeft)
+	for i := 0; i*parallelTriePrefetchCapacity < keysLeftSize; i++ {
+		child := newSubfetcher(sf.db, sf.root, sf.accountHash)
```
qinglin89, line 405: would lead `used` to incorrect statistics. also `dups`
- keys could be prefetched multi times among different children.
- might lift `seen` to parent's level ?

core/state/trie_prefetcher.go
```diff
@@ -348,12 +366,51 @@ func (sf *subfetcher) schedule(keys [][]byte) {
 	sf.lock.Lock()
 	sf.tasks = append(sf.tasks, keys...)
 	sf.lock.Unlock()
-
 	// Notify the prefetcher, it's fine if it's already terminated
 	select {
 	case sf.wake <- struct{}{}:
 	default:
 	}
+	atomic.AddUint32(&sf.pendingSize, uint32(len(keys)))
```
qinglin89, line 374: might do this before append `tasks`, otherwise decrease first and increase after would keep this children `fake full`

forcodedancing (approved):

core/state/statedb.go
```diff
@@ -236,6 +236,36 @@ func (s *StateDB) StopPrefetcher() {
 	}
 }
 
+func (s *StateDB) TriePrefetchInAdvance(block *types.Block, signer types.Signer) {
+	s.prefetcherLock.Lock()
```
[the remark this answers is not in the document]
setunapo: why? since `s.prefetcher` could be concurrently updated.

core/state/trie_prefetcher.go
setunapo: accept

core/state/statedb.go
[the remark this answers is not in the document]
setunapo: accepted

core/state/trie_prefetcher.go
```diff
@@ -161,17 +185,6 @@ func (p *triePrefetcher) mainLoop() {
 	}
```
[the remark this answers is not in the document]
setunapo: accept

core/state/trie_prefetcher.go
setunapo: accept

core/state/trie_prefetcher.go
setunapo: Actually it is ok, since parent subfetcher's tasks is unchanged, its metrics of `used`, `dups`, `seen` will be correct.

qinglin89 (approved):

qinglin89 (approved):

PR #972, opened 2022-06-30
[R4R]: Redesign triePrefetcher to make it thread safe

setunapo:
### Description
There are 2 types of triePrefetcher instances.
1.New created `triePrefetcher`: it is the one to do trie prefetch to speed up validation phase.
2.Copied `triePrefetcher`: it just do a copy of the prefetched trie,  it won't do prefetch at all. The copied tries are all kept in p.fetches.

In thie PR, we try to improve the new created one, to make it concurrent safe, while the copied one's behavior stay unchanged(its logic is very simple).

### Rationale
As commented in `triePrefetcher` struct, its APIs are not thread safe. So callers should make sure the created triePrefetcher should be used within a single routine. As we are trying to improve triePrefetcher, we would use it concurrently, so it is necessary to redesign it for concurrent access.

The redesigned workflow:
1.start a mainLoop to do the real trie prefetch work, caller could schedule prefetch request through channel message.
2.as the trie prefetch close operation is only allowed to do once and will be processed in the mainLoop too, since it will update the fetchers informantion.
3.The other operations will be done outside of the mainLoop for performance concern, since mainLoop could be too busy if all the works are done within the mainLoop. The action lists are: trie(),used(), copy(), these actions can be done in parallel with a simple read lock.

There is no obvious performance impact.
image

### Changes
No impact to users.

### DIFF
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -24,12 +24,26 @@ import (
 	"github.com/ethereum/go-ethereum/metrics"
 )
 
-const abortChanSize = 64
+const (
+	abortChanSize      = 64
+	concurrentChanSize = 10
+)
 
 var (
 	// triePrefetchMetricsPrefix is the prefix under which to publis the metrics.
 	triePrefetchMetricsPrefix = "trie/prefetch/"
 )
+
+type prefetchMsg struct {
+	root        common.Hash
+	accountHash common.Hash
+	keys        [][]byte
+}
+
+type usedMsg struct {
+	root common.Hash
+	used [][]byte
+}
 
 // triePrefetcher is an active prefetcher, which receives accounts or storage
 // items and does trie-loading of them. The goal is to get as much useful content
@@ -42,8 +56,17 @@ type triePrefetcher struct {
 	fetches  map[common.Hash]Trie        // Partially or fully fetcher tries
 	fetchers map[common.Hash]*subfetcher // Subfetchers for each trie
 
-	abortChan chan *subfetcher
-	closeChan chan struct{}
+	closeMainChan     chan struct{} // it is to inform the mainLoop
+	closeMainDoneChan chan struct{}
+	copyChan          chan struct{}
+	copyDoneChan      chan *triePrefetcher
+	prefetchChan      chan *prefetchMsg // no need to wait for return
+	trieChan          chan *common.Hash
+	trieDoneChan      chan *subfetcher
+	usedChan          chan *usedMsg // no need to wait for return
+
+	abortChan      chan *subfetcher
+	closeAbortChan chan struct{} // it is used to inform abortLoop
 
 	deliveryMissMeter metrics.Meter
 	accountLoadMeter  metrics.Meter
@@ -60,11 +83,21 @@ type triePrefetcher struct {
 func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePrefetcher {
 	prefix := triePrefetchMetricsPrefix + namespace
 	p := &triePrefetcher{
-		db:        db,
-		root:      root,
-		fetchers:  make(map[common.Hash]*subfetcher), // Active prefetchers use the fetchers map
-		abortChan: make(chan *subfetcher, abortChanSize),
-		closeChan: make(chan struct{}),
+		db:       db,
+		root:     root,
+		fetchers: make(map[common.Hash]*subfetcher), // Active prefetchers use the fetchers map
+
+		abortChan:      make(chan *subfetcher, abortChanSize),
+		closeAbortChan: make(chan struct{}),
+
+		closeMainChan:     make(chan struct{}),
+		closeMainDoneChan: make(chan struct{}),
+		prefetchChan:      make(chan *prefetchMsg, concurrentChanSize),
+		trieChan:          make(chan *common.Hash, concurrentChanSize),
+		trieDoneChan:      make(chan *subfetcher, concurrentChanSize),
+		usedChan:          make(chan *usedMsg, concurrentChanSize),
+		copyChan:          make(chan struct{}, concurrentChanSize),
+		copyDoneChan:      make(chan *triePrefetcher, concurrentChanSize),
 
 		deliveryMissMeter: metrics.GetOrRegisterMeter(prefix+"/deliverymiss", nil),
 		accountLoadMeter:  metrics.GetOrRegisterMeter(prefix+"/account/load", nil),
@@ -77,7 +110,101 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
 		storageWasteMeter: metrics.GetOrRegisterMeter(prefix+"/storage/waste", nil),
 	}
 	go p.abortLoop()
+	go p.mainLoop()
 	return p
+}
+
+func (p *triePrefetcher) mainLoop() {
+	// a series of anonymous functions which are concurrent unsafe,
+	// to avoid being accessed outside of the mainloop
+	copyFunc := func() *triePrefetcher {
+		copy := &triePrefetcher{
+			db:             p.db,
+			root:           p.root,
+			fetches:        make(map[common.Hash]Trie, len(p.fetches)),
+			abortChan:      make(chan *subfetcher),
+			closeAbortChan: make(chan struct{}),
+		}
+		// we're copying an active fetcher, retrieve the current states
+		for root, fetcher := range p.fetchers {
+			copy.fetches[root] = fetcher.peek()
+		}
+		return copy
+	}
+
+	closeFunc := func() {
+		for _, fetcher := range p.fetchers {
+			p.abortChan <- fetcher // safe to do multiple times
+			<-fetcher.term
+			if metrics.EnabledExpensive {
+				if fetcher.root == p.root {
+					p.accountLoadMeter.Mark(int64(len(fetcher.seen)))
+					p.accountDupMeter.Mark(int64(fetcher.dups))
+					p.accountSkipMeter.Mark(int64(len(fetcher.tasks)))
+
+					for _, key := range fetcher.used {
+						delete(fetcher.seen, string(key))
+					}
+					p.accountWasteMeter.Mark(int64(len(fetcher.seen)))
+				} else {
+					p.storageLoadMeter.Mark(int64(len(fetcher.seen)))
+					p.storageDupMeter.Mark(int64(fetcher.dups))
+					p.storageSkipMeter.Mark(int64(len(fetcher.tasks)))
+
+					for _, key := range fetcher.used {
+						delete(fetcher.seen, string(key))
+					}
+					p.storageWasteMeter.Mark(int64(len(fetcher.seen)))
+				}
+			}
+		}
+		close(p.closeAbortChan)
+		// Clear out all fetchers (will crash on a second call, deliberate)
+		p.fetchers = nil
+	}
+
+	prefetchFunc := func(root common.Hash, keys [][]byte, accountHash common.Hash) {
+		fetcher := p.fetchers[root]
+		if fetcher == nil {
+			fetcher = newSubfetcher(p.db, root, accountHash)
+			p.fetchers[root] = fetcher
+		}
+		fetcher.schedule(keys)
+	}
+
+	trieFunc := func(root common.Hash) *subfetcher {
+		// Otherwise the prefetcher is active, bail if no trie was prefetched for this root
+		fetcher := p.fetchers[root]
+		if fetcher == nil {
+			p.deliveryMissMeter.Mark(1)
+			return nil
+		}
+		return fetcher
+	}
+
+	usedFunc := func(root common.Hash, used [][]byte) {
+		if fetcher := p.fetchers[root]; fetcher != nil {
+			fetcher.used = used
+		}
+	}
+
+	// the main loop is executed from here
+	for {
+		select {
+		case <-p.copyChan:
+			p.copyDoneChan <- copyFunc()
+		case pMsg := <-p.prefetchChan:
+			prefetchFunc(pMsg.root, pMsg.keys, pMsg.accountHash)
+		case tireHash := <-p.trieChan:
+			p.trieDoneChan <- trieFunc(*tireHash)
+		case used := <-p.usedChan:
+			usedFunc(used.root, used.used)
+		case <-p.closeMainChan:
+			closeFunc()
+			close(p.closeMainDoneChan)
+			return
+		}
+	}
 }
 
 func (p *triePrefetcher) abortLoop() {
@@ -85,7 +212,7 @@ func (p *triePrefetcher) abortLoop() {
 		select {
 		case fetcher := <-p.abortChan:
 			fetcher.abort()
-		case <-p.closeChan:
+		case <-p.closeAbortChan:
 			// drain fetcher channel
 			for {
 				select {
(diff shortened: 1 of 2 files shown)


core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,86 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
 	go p.abortLoop()
+	go p.mainLoop()
 	return p
 }
 
+func (p *triePrefetcher) mainLoop() {
+	for {
+		select {
+		case <-p.copyChan:
+			copy := &triePrefetcher{
```
qinglin89, line 123: avoid 'copy'

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,86 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+				db:      p.db,
+				root:    p.root,
+				fetches: make(map[common.Hash]Trie, len(p.fetchers)),
```
qinglin89, line 126: lost `channels` would cause panic with nil pointer

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,86 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+			}
+			// we're copying an active fetcher, retrieve the current states
+			for root, fetcher := range p.fetchers {
+				copy.fetches[root] = fetcher.peek()
```
qinglin89, line 130: maybe the `peek` with currently `copy` is a better implementation?

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,86 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+			}
+			p.copyDoneChan <- copy
+
+		case pMsg := <-p.prefetchChan:
+			fetcher := p.fetchers[pMsg.root]
```
qinglin89, line 135: as above, prefetchChan should be copied, otherwise no concurrency safe could be assurance

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,86 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+			if fetcher == nil {
+				fetcher = newSubfetcher(p.db, pMsg.root, pMsg.accountHash)
+				p.fetchers[pMsg.root] = fetcher
+			}
+			fetcher.schedule(pMsg.keys)
+
+		case tireMsg := <-p.trieChan:
+			fetcher := p.fetchers[tireMsg.root]
```
qinglin89, line 143: as the original way, when trie be used, the `prefetcher` on this root should be stoped, as it would reach downwards to disk until get a trie in the `caller`

core/state/trie_prefetcher.go
```diff
@@ -101,69 +207,41 @@ func (p *triePrefetcher) abortLoop() {
...
-				for _, key := range fetcher.used {
-					delete(fetcher.seen, string(key))
-				}
-				p.storageWasteMeter.Mark(int64(len(fetcher.seen)))
-			}
-		}
+	select {
+	case <-p.closeMainChan: // already closed
+		return
+	default:
```
qinglin89, line 215: multi-close would cause panic

core/state/trie_prefetcher.go
```diff
@@ -188,15 +260,16 @@ func (p *triePrefetcher) trie(root common.Hash) Trie {
...
 			return nil
 		}
 		return p.db.CopyTrie(trie)
 	}
-	// Otherwise the prefetcher is active, bail if no trie was prefetched for this root
-	fetcher := p.fetchers[root]
+
+	trieChan := make(chan *subfetcher)
+	defer func() { close(trieChan) }()
+	p.trieChan <- &trieMsg{root, trieChan}
```
qinglin89, line 270: as above, should copy `triechan` if want to use this way for concurrency access

qinglin89:
too many channels and messages communication, actually only `fetchers` need to be taken care if want the `trie_prefetecher` be concurrent access.

core/state/trie_prefetcher.go
setunapo: accepted, rename `copy` to `fetcherCopied`

core/state/trie_prefetcher.go
setunapo: copied trie prefecther won't access channel, so, it is ok.

core/state/trie_prefetcher.go
setunapo: to be considered

core/state/trie_prefetcher.go
setunapo: as replied above, it is ok.

core/state/trie_prefetcher.go
setunapo: yes, it is ok, the sub fetcher of the root will be stopped, but not in mainLoop.
Deliberated moved outside for performance concern.

core/state/trie_prefetcher.go
setunapo: yes, would consider use lock or something else.

core/state/trie_prefetcher.go
setunapo: it is ok as noted above

setunapo (commented):
Replied

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,91 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+}
+
 func (p *triePrefetcher) abortLoop() {
 	for {
 		select {
 		case fetcher := <-p.abortChan:
 			fetcher.abort()
-		case <-p.closeChan:
+		case <-p.closeAbortChan:
 			// drain fetcher channel
```
qinglin89, line 200: why drain the channel ? before `closeAbortChan` has check `term` for all fetchers.

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,91 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+			fetcherCopied := &triePrefetcher{
+				db:      p.db,
+				root:    p.root,
+				fetches: make(map[common.Hash]Trie, len(p.fetchers)),
+			}
+			// we're copying an active fetcher, retrieve the current states
+			for root, fetcher := range p.fetchers {
+				fetcherCopied.fetches[root] = fetcher.peek()
+			}
+			p.copyDoneChan <- fetcherCopied
```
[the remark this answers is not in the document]
brilliant-lx: For channel of channel:
** pros: save channel buffer & return value in order
** cons: temporary channel variables

Accepted, since channel variable's cost is negligible,  can be seen as a normal variable.

core/state/trie_prefetcher.go
```diff
@@ -77,15 +112,91 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+
+						for _, key := range fetcher.used {
+							delete(fetcher.seen, string(key))
+						}
+						p.storageWasteMeter.Mark(int64(len(fetcher.seen)))
+					}
+				}
+			}
+			close(p.closeAbortChan)
+			close(p.closeMainDoneChan)
```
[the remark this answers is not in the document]
brilliant-lx: accept

core/state/trie_prefetcher.go
```diff
@@ -101,69 +212,58 @@ func (p *triePrefetcher) abortLoop() {
...
+	// If the prefetcher is an inactive one, bail out
+	if p.fetches != nil {
+		return
+	}
+	select {
+	case <-p.closeMainChan: // skip if already closed
+		return
+	default:
+		close(p.closeMainChan)
+		<-p.closeMainDoneChan // wait until all subfetcher are stopped
```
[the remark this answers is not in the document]
brilliant-lx: accept, may not need to use channel for close, think about atomic value check.

core/state/trie_prefetcher.go
brilliant-lx: accept, can be removed.

core/state/trie_prefetcher.go
```diff
@@ -101,69 +212,58 @@ func (p *triePrefetcher) abortLoop() {
...
+			fetches: make(map[common.Hash]Trie, len(p.fetches)),
+		}
+		// for closed trie prefetcher, retrieve the current states
+		for root, fetcher := range p.fetchers {
+			fetcherCopied.fetches[root] = fetcher.peek()
+		}
+		return fetcherCopied
+	default:
+		p.copyChan <- struct{}{}
+		return <-p.copyDoneChan
```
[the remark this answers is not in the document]
setunapo: not accept, same as above

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
setunapo: accept, will try to solve it.

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
setunapo: accept, may not need to use channel for close, think about atomic value check.

core/state/trie_prefetcher.go
setunapo: accept, can be removed.

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
setunapo: updated.
use multiple select for close()  & trie() now, so they won't deadlock any more.

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
setunapo: updated, use CompareAndSwapInt32 to make it concurrent safe.

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
brilliant-lx: copy is now moved out of the mainLoop, use RWMutex to protect for efficiency.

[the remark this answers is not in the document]
setunapo: copy is now moved out of the mainLoop, use RWMutex to protect for efficiency.

core/state/trie_prefetcher.go
```diff
@@ -77,20 +97,62 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
 		select {
-		case fetcher := <-p.abortChan:
-			fetcher.abort()
-		case <-p.closeChan:
-			// drain fetcher channel
+		case pMsg := <-p.prefetchChan:
+			fetcher := p.fetchers[pMsg.root]
+			if fetcher == nil {
+				fetcher = newSubfetcher(p.db, pMsg.root, pMsg.accountHash)
+				p.fetchersMutex.Lock()
```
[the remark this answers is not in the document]
setunapo: Yes, we can eliminate the `prefetchChan` and move the logic to the caller side, but it would make the code logic more complicated. 
Since there could several routine do schedule at the same time, mean while, close will be executed in other routine... We have to add the sync protect for subfetcher's create and delete.
It is much clean to keep subfetcher's lifecycle within a routine,

qinglin89 (dismissed):

core/state/trie_prefetcher.go
```diff
@@ -214,8 +281,23 @@ func (p *triePrefetcher) trie(root common.Hash) Trie {
 // used marks a batch of state items used to allow creating statistics as to
 // how useful or wasteful the prefetcher is.
 func (p *triePrefetcher) used(root common.Hash, used [][]byte) {
-	if fetcher := p.fetchers[root]; fetcher != nil {
-		fetcher.used = used
+	if !metrics.EnabledExpensive {
```
yutianwu, line 284: why does metrics config affect the logic

forcodedancing (approved):

core/state/trie_prefetcher.go
setunapo: `used` is an expensive metrics, not needed if EnabledExpensive is not enabled.

core/state/trie_prefetcher.go
```diff
@@ -172,13 +237,10 @@ func (p *triePrefetcher) prefetch(root common.Hash, keys [][]byte, accountHash c
...
 		return
 	}
-	// Active fetcher, schedule the retrievals
-	fetcher := p.fetchers[root]
-	if fetcher == nil {
-		fetcher = newSubfetcher(p.db, root, accountHash)
-		p.fetchers[root] = fetcher
+	select {
+	case <-p.closeMainChan: // skip closed trie prefetcher
+	case p.prefetchChan <- &prefetchMsg{root, accountHash, keys}:
```
guagualvcha, line 242: I will suggest add a `case default` here. If the `p.prefetchChan` if full, it will block the transaction execution.

core/state/trie_prefetcher.go
```diff
@@ -42,8 +52,14 @@ type triePrefetcher struct {
 	fetches  map[common.Hash]Trie        // Partially or fully fetcher tries
 	fetchers map[common.Hash]*subfetcher // Subfetchers for each trie
 
-	abortChan chan *subfetcher
-	closeChan chan struct{}
+	closed            int32
```
guagualvcha, line 55: I am confused about how `fetches` here been used, I did not see anywhere it is been assigned.

core/state/trie_prefetcher.go
qinglin89: will skip since no `for` loop here. `select` will keep try the channel till one is available, so no deadlock

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
qinglin89: avoid double close on `closeMainChan `

core/state/trie_prefetcher.go
```diff
@@ -188,20 +250,25 @@ func (p *triePrefetcher) trie(root common.Hash) Trie {
...
 		p.deliveryMissMeter.Mark(1)
 		return nil
 	}
+
 	// Interrupt the prefetcher if it's by any chance still running and return
 	// a copy of any pre-loaded trie.
-	p.abortChan <- fetcher // safe to do multiple times
+	select {
+	case <-p.closeAbortChan:
+	case p.abortChan <- fetcher: // safe to do multiple times
```
guagualvcha, line 270: I agree with @qinglin89 , if we continue to send message to `p.abortChan` when `abortLoop` already return , deadlock may happen.


I am afraid about this, a long run test  and a go routine profile to ensure no routine leakage happen.

core/state/trie_prefetcher.go
setunapo: 1.it should not be deadlock
2.if we add default, prefetchMsg could be dropped once the channel is full.

core/state/trie_prefetcher.go
setunapo: `fetches ` is for copied triePefetcher only, for newly created triePrefetcher, `fetches` is nil and useless

core/state/trie_prefetcher.go
setunapo: As mention above, should no deadlock. I can capture a go profile, I have test device running for more than 1day.

core/state/trie_prefetcher.go
```diff
@@ -77,20 +97,62 @@ func newTriePrefetcher(db Database, root common.Hash, namespace string) *triePre
...
+			p.fetchersMutex.Lock()
+			p.fetchers = nil
+			p.fetchersMutex.Unlock()
+
+			// drain all the channels before quit the loop
 			for {
 				select {
-				case fetcher := <-p.abortChan:
-					fetcher.abort()
+				case <-p.prefetchChan:
```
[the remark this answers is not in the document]
setunapo: Do we need to explicitly close a channel? Since it will be GC once it is no referred any more.

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
brilliant-lx: yes, prefetcherLock is to pretect `prefetcers`, `closed` flags is to make sure we do close only once.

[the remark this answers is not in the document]
setunapo: yes, prefetcherLock is to pretect prefetcers, closed flags is to make sure we do close only once.

core/state/trie_prefetcher.go
guagualvcha: make sense.

core/state/trie_prefetcher.go
guagualvcha: get.

guagualvcha (approved):

core/state/trie_prefetcher.go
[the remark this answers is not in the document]
j75689: Yes, it will be collected by GC. But personally, I still habitually close or recycle unused resources manually.

This won't cause serious code problems, maybe can follow your style

j75689 (approved):

yutianwu (approved):

guagualvcha (approved):

j75689 (dismissed):

guagualvcha (approved):

yutianwu (approved):

guagualvcha (dismissed):

j75689 (dismissed):

qinglin89 (dismissed):

guagualvcha (approved):

brilliant-lx (approved):
SGTM

qinglin89 (approved):

j75689 (approved):

yutianwu (approved):

KeefeL (approved):

PR #992, opened 2022-07-14
[R4R] Pipecommit enable trie prefetcher

qinglin89:
### Description
current `pipecommit` mode would use dummyRoot as account storageRoot before previous block had done commit. This would lead `trirePrefetcher` not able to work.

### Rationale
Remove dummyRoot to keep previous storage root while doing triePrefetch and recover stateRoot to enable effective triePrefetch in `pipecommit` mode
It's fine to do `triePrefetch` on previous `stateRoot/storageRoot` and would do a lot help for all the `accounts/storages` which are not modified during previous block.
And this would cause no harm since we use the corrected root when we do updateTrie and commit, which would ignore the stale node prefetched and the path node in cache would still help even though they were prefetched from a stale root.

#### metrics on `verifystate` stage under `pipecommit` mode
verifystate-metrics
metrics
green-line: before<br/> 
yellow-line: after

#### triePrefetch efficiency:
- on Account &emsp; **up from 0% to 100%**
  - before: 0%
   accountPrefetch_before

  - after: 100%
   accountPrefetch_after

 
- on Storage
  - before: about 70%
  storagePrefetch_before

  - after: about 84%
  storagePrefetch_after

### Changes

Notable changes: 
* remove `dummyroot`, keep root before previous block for `triePrefetch` on `storage`
* use `stateRoot` of block before previous one for `triePrefetch`on `account`

### DIFF
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -49,11 +49,11 @@ type prefetchMsg struct {
 //
 // Note, the prefetcher's API is not thread safe.
 type triePrefetcher struct {
-	db           Database    // Database to fetch trie nodes through
-	root         common.Hash // Root hash of theaccount trie for metrics
-	rootPrefetch common.Hash
-	fetches      map[common.Hash]Trie        // Partially or fully fetcher tries
-	fetchers     map[common.Hash]*subfetcher // Subfetchers for each trie
+	db         Database    // Database to fetch trie nodes through
+	root       common.Hash // Root hash of theaccount trie for metrics
+	rootParent common.Hash
+	fetches    map[common.Hash]Trie        // Partially or fully fetcher tries
+	fetchers   map[common.Hash]*subfetcher // Subfetchers for each trie
 
 	abortChan         chan *subfetcher // to abort a single subfetcher and its children
 	closed            int32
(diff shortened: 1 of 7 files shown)


tyntyn1:
@hackmod

brilliant-lx (commented):
To support the account stale metrics, this PR change some API interface, e.g. `Snapshot.Parent`, `NewTriePrefetch`... It may not worth it.
Pls consider remove metrics in this PR.

qinglin89:
> To support the account stale metrics, this PR change some API interface, e.g. `Snapshot.Parent`, `NewTriePrefetch`... It may not worth it. Pls consider remove metrics in this PR.

These modify was not meant to support the metrics, they just used for account root prefetch in `pipecommit` mode, and the new metrics was added for consistence since new account root was added for prefetch.

qinglin89:
> @hackmod

any comments ？

core/state/trie_prefetcher.go
```diff
@@ -49,10 +49,11 @@ type prefetchMsg struct {
 //
 // Note, the prefetcher's API is not thread safe.
 type triePrefetcher struct {
-	db       Database                    // Database to fetch trie nodes through
-	root     common.Hash                 // Root hash of theaccount trie for metrics
-	fetches  map[common.Hash]Trie        // Partially or fully fetcher tries
-	fetchers map[common.Hash]*subfetcher // Subfetchers for each trie
+	db           Database    // Database to fetch trie nodes through
+	root         common.Hash // Root hash of theaccount trie for metrics
+	rootPrefetch common.Hash
```
brilliant-lx, line 54: Maybe we can name it clearly? `rootPrefetch` is a bit confusing. Or you should add some comments to it.

brilliant-lx: How about rootParent? It is the root's parent's state root

qinglin89: updated

brilliant-lx (approved):
LGTM

forcodedancing (approved):

guagualvcha (approved):

Issue #991, opened 2022-07-13
[Feature] update dockerfile to support env variable


jasonyic:
# Rationale

Currently, `Dockerfile` does not support start with default configuration, e.g testnet configuration and docker run does not init genesis & start node automatically. same docker image should also be able to start as full node in either `mainnet` or `testnet`, based on a `env` variable, e.g. `BSC_NETWORK`.

e.g. 
`docker run -d ghcr.io/bnb-chain/bsc` should init with the genesis file and start a full node in `testnet` 
`docker run -d -e BSC_NETWORK=mainnet ghcr.io/bnb-chain/bsc` should start a full node in `mainnet`
`docker run -d -e BSC_NETWORK=mainnet ghcr.io/bnb-chain/bsc --datadir /node` should start a full node in `mainnet` with custom data directory.

# Implementation
refactor Dockerfile to include entrypoint script to use the network related config.toml, base on environment variable `BSC_NETWORK`. entrypoint script will also init with the genesis file and accept any additional command line options

Two predefined env variables `BSC_NETWORK` & `BSC_DATA_DIR` are used to provide the convenience to launch different node in different network and customise the data directory.

PR #993, opened 2022-07-14
feat: refactor dockerfile to add entrypoint script

jasonyic:
### Description
- refactor `dockerfile` to add default `config.toml` & `genesis` file from `mainnet` & `testnet` based on the binary release version.
- add entry point script to init genesis file and start node, with additional command options
- add support on 2 env variables `BSC_NETWORK` & `BSC_DATA_DIR` to provide the flexibility to boot node in different network. 

### Rationale
 - can easily start a full node in `testnet` or `mainnet` with default configuration.
 - also support the command line options to overwrite default configuration
 
### Example
`docker run -d ghcr.io/bnb-chain/bsc` should init with the genesis file and start a full node in `testnet`
`docker run -d -e BSC_NETWORK=mainnet ghcr.io/bnb-chain/bsc` should start a full node in `mainnet`
`docker run -d -e BSC_NETWORK=mainnet ghcr.io/bnb-chain/bsc --datadir /data2` should start a full node in `mainnet` with custom data directory.

### Changes

Notable changes: 
* refactor dockefile to add default configuration, 
* refactor dockerfile to use non-root user
* refactor dockerfile to use `tini` for better process handling
* add entry point script to init genesis file and boot node, also support command line options
* a docker-composer file provided for local testing

Feature Request: https://github.com/bnb-chain/bsc/issues/991


Dockerfile
```diff
@@ -7,10 +7,46 @@ ADD . /go-ethereum
...
+    && echo "Version: ${BSC_VERSION}" > /etc/motd
+
+WORKDIR ${BSC_HOME}
 
-RUN apk add --no-cache ca-certificates curl jq tini
 COPY --from=builder /go-ethereum/build/bin/geth /usr/local/bin/
 
+COPY docker-entrypoint.sh ./
+
+RUN curl -LO https://github.com/bnb-chain/bsc/releases/download/${BSC_VERSION}/mainnet.zip \
```
j75689, line 37: the URL can be replaced by `.github/release.env`. 
the mainnet and testnet config of the new release is obtained here

docker-compose.yaml
```diff
@@ -0,0 +1,23 @@
+version: '3.6'
+services:
+  bsc:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: ["--http.addr", "0.0.0.0", "--ws.addr", "0.0.0.0"]
+    environment:
+      - LOG_LEVEL=3
+      - DIFF_SYNC=true
```
j75689, line 10: Most of the flags do not yet support the reading of env...
Must use additional script to convert to flag

Dockerfile
jasonyic: `release.env` still pointing to `v1.1.8` and that is why I have a ARG to build on.

docker-compose.yaml
jasonyic: the reason which I suggest to only support 2 BSC unique env: `BSC_NETWORK` & `BSC_DATA_DIR` is that `geth` has quite a few command line options and is quite familiar by the developers. to keep the compatibility and minimise the maintenance for the script. `BSC_NETWORK` is used to boot the node to different network. default `/data` will be convenient to configure docker image, specially when hosting in k8s when mounting volume. `config.toml` & `genesis.json` can also be mounted by `configmap` in k8s into the `BSC_HOME/mainnet/` or`BSC_HOME/testnet/` before bootstrap script kicks in.

jasonyic: another thought on building network specific image is to use another `ARG: <mainnet|testnet>` to build image for `mainnet` & `testnet` separately, this will end up with two image tags, e.g. `ghcr.io/bnb-chain/bsc:v1.1.11-mainnet`, `ghcr.io/bnb-chain/bsc:v1.1.11-testnet`. the config in the docker image will only host the config from the network..this will give much clean image but extra image tag.

Dockerfile
j75689: Actually every time a new version is released the additional files come from here, so it will be the same if `releas.env` doesn't update

The reason I want you to use `release.env` is that developers will most likely forget to update the Dockerfile when they release a new version

docker-compose.yaml
j75689: 1. I agree with you, but I just want to say that the `DIFF_SYNC=true` environment variable has no effect here
2. I think building it as an image is enough, the program is the same, just the configuration is different

j75689:
Can you write a simple document on how to use your docker image? It can be placed in `docs/docker/readme.md`

j75689:
and please change the target branch to `develop`, thanks

Dockerfile
jasonyic: when each release, no need to update the `BSC_VERSION` in `Dockerfile`, the build process should pass in `--build-arg` for the new version to be used, similar to the new target `docker` in makefile. Additionally, this won't cause confusion for  developer between the regular release in the github`- repo and configuration version being used in Dockerfile.

docker-compose.yaml
jasonyic: sample docker-compose file just updated to reflect the usage with command line.
in k8s, it will look like
```
     containers:
      - name: client
        image: bnb-chain/bsc:v1.1.11
        args: ["--http.addr", "0.0.0.0", "--http.port", "8545", "--http.vhosts", "*", "--verbosity", "3"]
        env:
          - name: BSC_NETWORK
            value: mainnet
```

jasonyic:
close this PR and replace with PR https://github.com/bnb-chain/bsc/pull/998

PR #996, opened 2022-07-15
[R4R]: Trie prefetch on state pretch

setunapo:
### Description
This is another PR to enhance the trie prefetch, try to reuse the execution results of state prefetch.

### Rationale
Currently, state prefetch just pre-execute the transactions and discard the results. It is helpful to increase the snapshot cache hit rate. It would be more helpful, if it can do trie prefetch at the same time, since the it will preload the trie node and build the trie tree in advance.
This PR is to implement it, by reusing the main trie prefetch and doing finalize after transaction is executed.

### Test Result
With this PR, the peek validation of the BSC 2022.04.18 burst traffic reduced by ~39% even based on the previous trie prefetch optimizations.
image

### Changes
No impact to user.

### DIFF
--- a/core/state/trie_prefetcher.go
+++ b/core/state/trie_prefetcher.go
@@ -172,15 +172,7 @@ func (p *triePrefetcher) mainLoop() {
 			p.fetchersMutex.Lock()
 			p.fetchers = nil
 			p.fetchersMutex.Unlock()
-
-			// drain all the channels before quit the loop
-			for {
-				select {
-				case <-p.prefetchChan:
-				default:
-					return
-				}
-			}
+			return
 		}
 	}
 }
@@ -362,6 +354,7 @@ func newSubfetcher(db Database, root common.Hash, accountHash common.Hash) *subf
 
 // schedule adds a batch of trie keys to the queue to prefetch.
 func (sf *subfetcher) schedule(keys [][]byte) {
+	atomic.AddUint32(&sf.pendingSize, uint32(len(keys)))
 	// Append the tasks to the current queue
 	sf.lock.Lock()
 	sf.tasks = append(sf.tasks, keys...)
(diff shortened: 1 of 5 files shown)


qinglin89 (approved):

core/state/trie_prefetcher.go
```diff
@@ -362,6 +354,7 @@ func newSubfetcher(db Database, root common.Hash, accountHash common.Hash) *subf
 
 // schedule adds a batch of trie keys to the queue to prefetch.
 func (sf *subfetcher) schedule(keys [][]byte) {
+	atomic.AddUint32(&sf.pendingSize, uint32(len(keys)))
```
forcodedancing, line 357: Why need to change the location of this line?

core/state_prefetcher.go
```diff
@@ -153,5 +153,8 @@ func precacheTransaction(msg types.Message, config *params.ChainConfig, gaspool
 	// Update the evm with the new transaction context.
 	evm.Reset(NewEVMTxContext(msg), statedb)
 	// Add addresses to access list if applicable
-	ApplyMessage(evm, msg, gaspool)
+	if _, err := ApplyMessage(evm, msg, gaspool); err == nil {
+		statedb.Finalise(true)
```
forcodedancing, line 157: What's the difference if we call tire_prefetcher here? Not sure is there any other side effects.

core/state/trie_prefetcher.go
brilliant-lx: `pendingSize` is added on `schedule`, while decreased in `loop`, move it ahead to avoid decreasing before increasing.

setunapo: `pendingSize` is added on `schedule`, while decreased in `loop`, move it ahead to avoid decreasing before increasing.

core/state_prefetcher.go
setunapo: Actually, we should do Finalise() here, to flush DirtyObjects into PendingObjects and clear the journal.
It can help to make the concurrent StatePrefetch's behavior more similar to the main process.
It is ok, should have no side effect.

forcodedancing (approved):

guagualvcha (approved):

PR #1002, opened 2022-07-19
[R4R] Fix pipecommit active statedb

qinglin89:
### Description

`prefetcher` be accessed after non-nil condition as a field of `statedb` is not safe.
eg: as in `pipecommit` mode, `statedb` might be used as `activestatedb` to `StopPrefetcher`, which would `panic` during:
```
if prefetcher != nil {
    prefetcher.used(s.data.Root, usedStorage)
 }
```

### Rationale

get `prefetcher` from `statedb` before check nil condition

### Example

N/A

### Changes
-

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -857,8 +857,9 @@ func (s *StateDB) Copy() *StateDB {
 	// If there's a prefetcher running, make an inactive copy of it that can
 	// only access data but does not actively preload (since the user will not
 	// know that they need to explicitly terminate an active copy).
-	if s.prefetcher != nil {
-		state.prefetcher = s.prefetcher.copy()
+	prefetcher := s.prefetcher
+	if prefetcher != nil {
+		state.prefetcher = prefetcher.copy()
 	}
 	if s.snaps != nil {
 		// In order for the miner to be able to use and make additions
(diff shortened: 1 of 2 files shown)


core/state/statedb.go
```diff
@@ -1000,8 +1001,9 @@ func (s *StateDB) Finalise(deleteEmptyObjects bool) {
 			addressesToPrefetch = append(addressesToPrefetch, common.CopyBytes(addr[:])) // Copy needed for closure
 		}
 	}
-	if s.prefetcher != nil && len(addressesToPrefetch) > 0 {
-		s.prefetcher.prefetch(s.originalRoot, addressesToPrefetch, emptyAddr)
+	prefetcher := s.prefetcher
+	if prefetcher != nil && len(addressesToPrefetch) > 0 {
```
guagualvcha, line 1005: can you add some UTs to battle test the concurrence issue of statedb?

qinglin89: pointer assignment is actually atomic, but would fail race detection.

brilliant-lx (approved):
LGTM

guagualvcha (dismissed):

forcodedancing (dismissed):

brilliant-lx (approved):
LGTM

guagualvcha (approved):

PR #1006, opened 2022-07-20
statedb: get rid of activestate, stopPrefetcher manually

qinglin89:
### Description

- Since `statedb` would be accessed concurrently in different routine, it's not proper to use defer with `activeState` in the `insertchain`
- `triePrefetcher` is safe to be accessed like `trie` to get cached node when stopped, and since `triePrefetcher` had been redesigned to concurrency safe, there's no need to use `Lock`.

### Rationale

`StopPrefetcher` manually

### Example

N/A

### Changes

Notable changes: 
* `StopPrefetcher` manually
*  Modify `StopPrefetcher` to keep stopped prefetcher in `statedb`

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1356,19 +1356,17 @@ func (s *StateDB) Commit(failPostCommitFunc func(), postCommitFuncs ...func() er
 	}
 	// Finalize any pending changes and merge everything into the tries
 	if s.lightProcessed {
+		defer s.StopPrefetcher()
 		root, diff, err := s.LightCommit()
 		if err != nil {
-			s.StopPrefetcher()
 			return root, diff, err
 		}
 		for _, postFunc := range postCommitFuncs {
 			err = postFunc()
 			if err != nil {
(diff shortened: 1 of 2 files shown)


guagualvcha:
hi, this PR will not be included in release v1.1.12, so may hold it for a while.

core/state/statedb.go
```diff
@@ -1326,10 +1325,12 @@ func (s *StateDB) LightCommit() (common.Hash, *types.DiffLayer, error) {
 // Commit writes the state to the underlying in-memory trie database.
 func (s *StateDB) Commit(failPostCommitFunc func(), postCommitFuncs ...func() error) (common.Hash, *types.DiffLayer, error) {
 	if s.dbErr != nil {
+		s.StopPrefetcher()
```
brilliant-lx, line 1328: why `StopPrefetcher` in Commit

brilliant-lx (approved):

guagualvcha (approved):

PR #1007, opened 2022-07-20
[R4R] fix: missing contract upgrades and incorrect behavior when miners enable pipecommit

j75689:
### Description

1. The code for the mining block upgrade system contract is missing. https://github.com/bnb-chain/bsc/blob/de4abf9d09afac7ad54d3268647d1c8f5aa61e95/miner/worker.go#L966
2. Incorrect update of `stateb` when miner enables `pipecommit`.

### Rationale

1. Missed when merging ethereum versions.
2. When a miner does a `commit`, a new `*environment` is copied to work, but when `pipecommit` is enabled, `CorrectAccountsRoot` should be called to correct the root hash of the accounts in the snapshot, which results in only the copied ones be updated, and the continuously running `statedb` is not properly updated.

### Example

N/A

### Changes

Notable changes: 
* miner/worker.go

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/consensus/parlia"
 	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/state"
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
@@ -577,7 +578,7 @@ func (w *worker) mainLoop() {
 			if w.isRunning() && w.current != nil && len(w.current.uncles) < 2 {
 				start := time.Now()
 				if err := w.commitUncle(w.current, ev.Block.Header()); err == nil {
-					w.commit(w.current.copy(), nil, false, start)
+					w.commit(w.current, nil, false, start)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -1185,16 +1190,19 @@ func (w *worker) commit(env *environment, interval func(), update bool, start ti
 		if interval != nil {
 			interval()
 		}
+
+		if env.state.IsPipeCommit() {
```
forcodedancing, line 1194: do not need this condition

j75689: fixed

forcodedancing (approved):

guagualvcha (approved):

miner/worker.go
```diff
@@ -1166,7 +1171,7 @@ func (w *worker) commitWork(interrupt *int32, noempty bool, timestamp int64) {
 	}
 	// Fill pending transactions from the txpool
 	w.fillTransactions(interrupt, work)
-	w.commit(work.copy(), w.fullTaskHook, true, start)
```
brilliant-lx, line 1169: There are other 2 spots that `w.commit(work.copy(), ...)` will be executed.
Seems `work.copy()` are all unnecessary?

brilliant-lx (approved):
LGTM

miner/worker.go
j75689: I think yes, it copies twice

guagualvcha (approved):

forcodedancing (approved):

brilliant-lx (approved):
LGTM

PR #1009, opened 2022-07-21
[R4R]fix: resolve the concurrent cache read and write issue for fast node

kyrie-yl:
Signed-off-by: cryyl <yl.on.the.way@gmail.com>

### Description

When fast node insert a block, it will judge whether the 11 ahead of current block has been verified. While, if a block is an empty block, it must be verified. So if the 11 ahead block is an empty block, the AncestorVerified function will return true.

### Changes

Notable changes: 
* In the AncestorVerified function, if the 11 ahead of current block is an empty block, return true.


guagualvcha (approved):

guagualvcha:
LGTM

guagualvcha (approved):

forcodedancing (approved):

yutianwu (approved):

j75689 (approved):

PR #1011, opened 2022-07-22
fix: Incorrect merkle root issue when enabling pipecommit with miner

j75689:
### Description

1. Trie data is incorrect when enabling `pipecommit` with miner.
image
2. Account root hash is incorrect when doing `statedb. CorrectAccountsRoot `.

### Rationale

1. 1. Because the `parlia` consensus will call the system contract in `FinalizeAndAssemble`, and change the state of the world. But before that, the `statedb` is copied, but this copied `statedb` will not be used when writing blocks, resulting in incorrect writing of the world state.
2. 1. When `account root` is an empty hash, the correct `emptyhash` value is not assigned.

### Example

N/A

### Changes

Notable changes: 
* miner/worker.go


realuncle (approved):

guagualvcha (approved):

forcodedancing:
Cool. So we need to put the `env copy` code after  `FinalizeAndAssemble` to make sure `statdb` is corrected all places.

forcodedancing (approved):

yutianwu (approved):

PR #1020, opened 2022-07-27
[R4R] fix: remove diffhash patch introduced from separate node

j75689:
### Description

BSC has merged https://github.com/ethereum/go-ethereum/pull/23628 to avoid hitting outdated snapshots, so we can remove #926 code that was introduced at the time to fix diffHash mismatch. 

### Rationale

image

In previous versions, the snapshot tree executed `parent.flatten()` without first locking the current difflayer when `cap`, but first executed the `parent.flatten()` to change it to be `stale` state and relinks the current `difflayer.parent` (removes the stale's difflayer from the recursive structure).

image

And this has caused a problem. There is a small chance that the stale difflayer will be obtained first when the difflayer is accessed externally, and then the difflayer.parent will be locked and re-linked, resulting in the snapshot being judged when acquiring the Account or Storage data. Returns null for `Stale`.

image

Therefore, in the implementation of #926, a new retry mechanism is added to retry when a null value is obtained, but this approach is not particularly good and may cause over-stack problems.

After the Ethereum version(https://github.com/ethereum/go-ethereum/pull/23628) was merged, it was changed to lock the current difflayer first and then change the parent to avoid the above problems, so we can remove the patch fix introduced at that time.

### Example

N/A

### Changes

Notable changes: 
* remove the patch from separate-node

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -18,7 +18,6 @@ package state
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"io"
 	"math/big"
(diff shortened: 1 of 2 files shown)


brilliant-lx (approved):
LGTM.
It has been protected by diffLayer.lock right now.

guagualvcha (approved):

forcodedancing (approved):

PR #1024, opened 2022-07-28
[R4R]fix: verify node is not treated as verify node

guagualvcha:
### Description
For the fast node, verify node is a special setting that can do remove state root verification.

The development branch does have bugs in mark verify nodes. This PR aims to fix the bug.

### Rationale
image
Both `dialDest` and `n` here is pointers, we shall just compare whether the ID equals.

### Example
After the patch, fast node can work smoothly in `full` and `insecure` mode.

### Changes

No


forcodedancing (approved):

KeefeL (approved):
LGTM

j75689 (approved):

PR #1034, opened 2022-07-30
add --pruneancient flag for snapshot prune-state cmd

barryz:
### Description

suppliment missing flag `--pruneancient` to snapshot sub-command prune-state.

### Rationale

To make sure the command `geth snapshot prune-state --datadir <datadir> --pruneancient` work as expectation after ancient db was pruned inline at runtime.

PR #1053, opened 2022-08-16
state: fix offline tool start failed when start with pruneancient

joeylichang:
### Description

Fixed the bug #1034 that 'prune-state/insecure-prune-all ...'  and some other offline commands will fail after using '--pruneancient' to start geth. 

### Changes

Notable changes: 
* Add ‘skipCheckFreezerType’ parameter to 'NewDatabaseWithFreezer' function.
* When offline commands open database with 'MakeChainDatabase', will set skipCheckFreezerType true, skip freezer type checking. Make sure the command is executed correctly.
* When starting geth, use 'OpenAndMergeDatabase' to open the database, will set skipCheckFreezerType false, force freezer type checking to prevent freezer from being mixed with prunefreezer.

### DIFF
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -93,7 +93,7 @@ func testOfflineBlockPruneWithAmountReserved(t *testing.T, amountReserved uint64
 		t.Fatalf("Failed to back up block: %v", err)
 	}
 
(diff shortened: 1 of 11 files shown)


guagualvcha (approved):

RenRick (approved):

fynn-z (commented):
LGTM

fynn-z (approved):

Mercybudda (approved):

PR #1060, opened 2022-08-19
consensus: fix the GasLimitBoundDivisor

j75689:
### Description
`GasLimitBoundDivisor` should be 256 in BSC, was changed to `1024` in the merged ethereum version, fixed back here

### Rationale

1. change `GasLimitBoundDivisor` to 256
2. Ignore some test cases for eip-1559, bsc doesn't enable it

### Example

1. Start a private network (2 v1.1.11 miners, 1  miner of this version), and continuously initiate transfer transactions, which does not affect block generation

### Changes

N/A


forcodedancing (approved):

guagualvcha (approved):

brilliant-lx (approved):
LGTM

PR #1067, opened 2022-08-23
fix:fix potential goroutine leak

bittrainee:
### Description
There might be no receiver for txCh when interruptCh is closed and goroutine will be blocked here forever

### Rationale

### Example

### Changes

Notable changes: 
* ...


alfiomy (approved):

alfiomy (approved):

brilliant-lx (approved):
nice finding, LGTM.

guagualvcha (approved):

PR #1121, opened 2022-10-08
vm: add two proof verifier to fix the vulnerability in range proof

guagualvcha:
### Description
This PR aims to solve the BSC Bridge exploitation issue.

### Rationale
BC to BSC cross-chain communication is based on the light client technique. A smart contract is deployed on BSC to track the consensus state of BC including the appHash (analogous to stateRoot of Ethereum), which is essentially a root of a verifiable key-value store implemented as an AVL tree. Messages coming from BC will be verified against the appHash so their integrity can be guaranteed.

The store allows consecutive key-value pairs to be proved in batch (better in performance than proving each one individually). However its range proof verification logic contains a critical bug that can be exploited to prove membership of arbitrary key-value pairs chosen by the attacker.

https://github.com/cosmos/iavl/blob/6c1300ae54a9bb851e77dbcc4ba4b21832279027/proof_path.go#L70

https://github.com/cosmos/iavl/blob/6c1300ae54a9bb851e77dbcc4ba4b21832279027/proof.go#L79

The root is calculated by hashing repeatedly (just like in Merkle proof) as in the pseudocode below:
image

 If both left and right exist, the digest will ignore the right field. Because of this, if the verification path contains all items with both left and right, the root is unchanged with any arbitrary right values (using the same leaf). This is one factor to be exploited.

This PR in tendermint will introduce `ProofOpVerifier` to disable following ProofOperation:
1. The absence proof operator;
2. The value proof operator with more than one leaves.
### Example
Refer to the UT.

### Changes
NA


core/vm/contracts_lightclient.go
```diff
@@ -135,27 +187,29 @@ func (c *iavlMerkleProofValidate) Run(input []byte) (result []byte, err error) {
...
-func (c *tmHeaderValidateNano) Run(input []byte) (result []byte, err error) {
-	return nil, fmt.Errorf("suspend")
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------------
-type iavlMerkleProofValidateNano struct{}
-
-func (c *iavlMerkleProofValidateNano) RequiredGas(input []byte) uint64 {
-	return params.IAVLMerkleProofValidateGas
+func forbiddenAbsenceOpVerifier(op merkle.ProofOperator) error {
```
owen-reorg, line 190: This can be done by removing related registers in `multistoreproof.go`

https://github.com/bnb-chain/bsc/blob/1a2da608df5a2e77aedf175225fa3210319f0667/core/vm/lightclient/multistoreproof.go#L131-L138

guagualvcha: We can not remove this, in case the history block or tx  contains `ProofOpIAVLAbsence` op.

core/vm/lightclient/types.go
```diff
@@ -225,11 +227,11 @@ func (kvmp *KeyValueMerkleProof) Validate() bool {
 	kp = kp.AppendKey(kvmp.Key, merkle.KeyEncodingURL)
```
owen-reorg, line 228: How about adding all the check logic in the `Validate()` function?
Thus we don't need to change tendermint dep. The drawback is we have to decode the proof twice.

```suggestion

	decodedProof, err := prt.DecodeProof(kvmp.Proof)
	if err != nil {
		return false
	}
	for _, op := range decodedProof{
		if valueOp, ok := op.(iavl.IAVLValueOp); ok {
			if len(valueOp.Proof.Leaves) != 1 {
				return false
			}
			for _, innerNodde := range valueOp.Proof.LeftPath {
				if len(innerNodde.Right) > 0 && len(innerNodde.Left) > 0 {
					return false
				}
			}
		}
	}
```

core/vm/contracts_lightclient_test.go
```diff
@@ -102,4 +104,34 @@ func TestTmHeaderValidateAndMerkleProofValidate(t *testing.T) {
...
+	moranValidator := iavlMerkleProofValidateMoran{}
+	res, err := moranValidator.Run(bytest)
+	assert.Errorf(t, err, "invalid merkle proof")
```
yutianwu, line 135: can we compare the error message here

core/vm/contracts_lightclient_test.go
```diff
@@ -102,4 +104,34 @@ func TestTmHeaderValidateAndMerkleProofValidate(t *testing.T) {
...
+	validator := iavlMerkleProofValidate{}
+	res, err := validator.Run(bytest)
+	assert.NoError(t, err)
+	assert.Equal(t, res[len(res)-1], uint8(1))
+
+	moranValidator := iavlMerkleProofValidateMoran{}
+	res, err = moranValidator.Run(bytest)
+	assert.Errorf(t, err, "invalid merkle proof")
```
yutianwu, line 126: can we compare the error message

zb3:
Out of curiosity (as an outsider), wouldn't the right package to change be the `iavl` library? The fix could then be merged upstream. Or I'm missing something?

core/vm/lightclient/types.go
guagualvcha: ok, I prefer to avoid `decode the proof twice`.

core/vm/contracts_lightclient_test.go
guagualvcha: `assert.Errorf ` is actually comparing the error message, unfortunately, we can only get high-level error message `invalid merkle proof`.

core/vm/contracts_lightclient_test.go
guagualvcha: as above👆

guagualvcha:
> Out of curiosity (as an outsider), wouldn't the right package to change be the `iavl` library? The fix could then be merged upstream. Or I'm missing something?

We can not change `iavl` library to the right logic directly, because BSC should support replay block from genesis to latest, it means even the exploitation transaction should be properly executed. That is why we did not change the `iavl` lib, but added two verifiers here.

core/vm/contracts_lightclient.go
```diff
@@ -135,27 +186,29 @@ func (c *iavlMerkleProofValidate) Run(input []byte) (result []byte, err error) {
...
-	return nil, fmt.Errorf("suspend")
+func singleValueOpVerifier(op merkle.ProofOperator) error {
+	if op == nil {
+		return nil
+	}
+	if valueOp, ok := op.(iavl.IAVLValueOp); ok {
+		if len(valueOp.Proof.Leaves) != 1 {
+			return cmn.NewError("range proof suspended")
+		}
+		for _, innerNodde := range valueOp.Proof.LeftPath {
```
brilliant-lx, line 207: `innerNodde` could be a typo?

core/vm/contracts_lightclient.go
```diff
@@ -135,27 +186,29 @@ func (c *iavlMerkleProofValidate) Run(input []byte) (result []byte, err error) {
...
-// | payload length | payload    |
-// | 32 bytes       |            |
-func (c *iavlMerkleProofValidateNano) Run(input []byte) (result []byte, err error) {
-	return nil, fmt.Errorf("suspend")
+func singleValueOpVerifier(op merkle.ProofOperator) error {
+	if op == nil {
+		return nil
+	}
+	if valueOp, ok := op.(iavl.IAVLValueOp); ok {
+		if len(valueOp.Proof.Leaves) != 1 {
```
brilliant-lx, line 204: Is it possible `len(valueOp.Proof.Leaves) == 0`?

core/vm/contracts_lightclient_test.go
```diff
@@ -102,4 +104,34 @@ func TestTmHeaderValidateAndMerkleProofValidate(t *testing.T) {
...
+	validator := iavlMerkleProofValidate{}
```
brilliant-lx, line 119: Could add check for the first 100M exploit tx,
https://bscscan.com/tx/0xebf83628ba893d35b496121fb8201666b8e09f3cbadf0e269162baa72efe3b8b
it is input is:
00000000000000000000000000000000000000000000000000000000000005086962630000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e00000100380200000000010dd85c0000000000000000000000000000000000000000000000000000000000000093000000000000000000000000000000000000000000000000000000000000000000f870a0424e4200000000000000000000000000000000000000000000000000000000009400000000000000000000000000000000000000008ad3c21bcecceda100000094489a8756c18c0b8b24ec2a2b9ff3d4d447f79bec94489a8756c18c0b8b24ec2a2b9ff3d4d447f79bec846553f10072cda827a83531ca0fd7ac917a6b65649719aab0836722caafe0603147a523180a8d020a066961766c3a76120e00000100380200000000010dd85c1af201f0010aed010a2b0802100318b091c73422200c10f902d266c238a4ca9e26fa9bc36483cd3ebee4e263012f5e7f40c22ee4d20a4d0801100218b091c7342220e4fd47bffd1c06e67edad92b2bf9ca63631978676288a2aa99f95c459436ef632a20da657c1ffb86c684eb3e265361ef0fa4f9dfa670b45f9f91c5eb6ad84b21a4d112001a370a0e0000010038020000000000000002122011056c6919f02d966991c10721684a8d1542e44003f9ffb47032c18995d4ac7f18b091c7341a340a0e00000100380200000000010dd85c12202c3a561458f8527b002b5ec3cab2d308662798d6245d4588a4e6a80ebdfe30ac18010ad4050a0a6d756c746973746f726512036962631ac005be050abb050a110a066f7261636c6512070a0508b891c7340a0f0a046d61696e12070a0508b891c7340a350a08736c617368696e6712290a2708b891c7341220c8ccf341e6e695e7e1cb0ce4bf347eea0cc16947d8b4e934ec400b57c59d6f860a380a0b61746f6d69635f7377617012290a2708b891c734122042d4ecc9468f71a70288a95d46564bfcaf2c9f811051dcc5593dbef152976b010a110a0662726964676512070a0508b891c7340a300a0364657812290a2708b891c73412201773be443c27f61075cecdc050ce22eb4990c54679089e90afdc4e0e88182a230a2f0a02736312290a2708b891c7341220df7a0484b7244f76861b1642cfb7a61d923794bd2e076c8dbd05fc4ee29f3a670a330a06746f6b656e7312290a2708b891c734122064958c2f76fec1fa5d1828296e51264c259fa264f499724795a740f48fc4731b0a320a057374616b6512290a2708b891c734122015d2c302143bdf029d58fe381cc3b54cedf77ecb8834dfc5dc3e1555d68f19ab0a330a06706172616d7312290a2708b891c734122050abddcb7c115123a5a4247613ab39e6ba935a3d4f4b9123c4fedfa0895c040a0a300a0361636312290a2708b891c734122079fb5aecc4a9b87e56231103affa5e515a1bdf3d0366490a73e087980b7f1f260a0e0a0376616c12070a0508b891c7340a300a0369626312290a2708b891c7341220e09159530585455058cf1785f411ea44230f39334e6e0f6a3c54dbf069df2b620a300a03676f7612290a2708b891c7341220db85ddd37470983b14186e975a175dfb0bf301b43de685ced0aef18d28b4e0420a320a05706169727312290a2708b891c7341220a78b556bc9e73d86b4c63ceaf146db71b12ac80e4c10dd0ce6eb09c99b0c7cfe0a360a0974696d655f6c6f636b12290a2708b891c73412204775dbe01d41cab018c21ba5c2af94720e4d7119baf693670e70a40ba2a52143

core/vm/contracts_lightclient.go
guagualvcha: fixed

core/vm/contracts_lightclient.go
guagualvcha: Not possible, it means nothing to prove.  If `len(valueOp.Proof.Leaves) == 0`, we should return an error.

core/vm/contracts_lightclient_test.go
guagualvcha: good suggestion, added.

brilliant-lx (approved):

params/config.go
```diff
@@ -717,6 +724,14 @@ func (c *ChainConfig) IsOnNano(num *big.Int) bool {
 	return configNumEqual(c.NanoBlock, num)
 }
 
+func (c *ChainConfig) IsMoran(num *big.Int) bool {
+	return isForked(c.MoranBlock, num)
+}
+
+func (c *ChainConfig) IsOnMora(num *big.Int) bool {
```
yutianwu, line 731: typo `IsOnMora`->`IsOnMoran`

guagualvcha: fixed

owen-reorg (approved):

Pythonberg1997 (approved):

Pythonberg1997 (approved):

PR #1138, opened 2022-10-16
all: sync with upstream

qinglin89:
### Description

Sync with `go-ethereum 1.10.19`:
mainly including some bug fixes, and together with some separate `prs`: https://github.com/bnb-chain/bsc/pull/1139, https://github.com/bnb-chain/bsc/pull/1140, https://github.com/bnb-chain/bsc/pull/1141.
### Rationale

N/A

### Example

N/A

### Changes

Notable changes: 
N/A

### DIFF
--- a/core/vm/analysis.go
+++ b/core/vm/analysis.go
@@ -76,7 +76,7 @@ func codeBitmapInternal(code, bits bitvec) bitvec {
 	for pc := uint64(0); pc < uint64(len(code)); {
 		op := OpCode(code[pc])
 		pc++
-		if op < PUSH1 || op > PUSH32 {
+		if int8(op) < int8(PUSH1) { // If not PUSH (the int8(op) > int(PUSH32) is always false).
 			continue
 		}
 		numbits := op - PUSH1 + 1
--- a/accounts/abi/bind/backends/simulated.go
+++ b/accounts/abi/bind/backends/simulated.go
@@ -63,9 +63,10 @@ type SimulatedBackend struct {
 	database   ethdb.Database   // In memory database to store our testing data
 	blockchain *core.BlockChain // Ethereum blockchain to handle the consensus
 
-	mu           sync.Mutex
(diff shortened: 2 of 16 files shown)


core/vm/analysis.go
```diff
@@ -76,7 +76,7 @@ func codeBitmapInternal(code, bits bitvec) bitvec {
 	for pc := uint64(0); pc < uint64(len(code)); {
 		op := OpCode(code[pc])
 		pc++
-		if op < PUSH1 || op > PUSH32 {
+		if int8(op) < int8(PUSH1) { // If not PUSH (the int8(op) > int(PUSH32) is always false).
```
brilliant-lx, line 79: `If not PUSH (the int8(op) > int(PUSH32) is always false)`
why?

guagualvcha (approved):

core/vm/analysis.go
junnmm: The opcode of ["PUSH1" is "0x60"](https://ethervm.io/#60), and all the other `PUSH` opcodes have a greater opCode than "PUSH1".  An opcode that is less than "0x60" can never be a "PUSH32" .

brilliant-lx: but `op > PUSH32` could be true right, say if op is Dup1(0x80)

junnmm:
Yes. You are right.

After realizing that I was wrong about this, I did a little search on this.

I found this function is primarily used for tests. And they never use an opcode greater than PUSH32 in their test cases:

https://github.com/ethereum/go-ethereum/blob/master/core/vm/analysis_test.go#L32


Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

junnmm:
Oh, I think I got.
It is not a bug introduced in benchmark.

0x80 is negative as a signed integer.



Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

setunapo (approved):

brilliant-lx (approved):

Issue #1133, opened 2022-10-13
Validator may get slashed even not be one of mining validator set


48ClubIan:
#### System information
Geth version: `1.1.16`
OS & Version: Linux

#### Expected behaviour
After BEP-131 introduced, 21 mining validators are selected from top 26 (for the moment this issue is raised) validators every 200 blocks.

It is possible for one specific validator to be unauthorized for mining.

#### Actual behaviour
Validator get slashed for missing block while is out of mining set.

#### Steps to reproduce the behaviour
We only observed 2 cases.

Case1:
https://bscscan.com/tx/0x5d57d5ce81f31ca52ab7e14a3253dea776fe0de3036697741573a562dfdc525a
0x72b61c6014342d914470ec7ac2975be345796c2b is slashed by 0xa6f79b60359f141df90a0c745125b131caaffd12 at 21771611

Case2:
https://bscscan.com/tx/0x3312f4597839f794e047944ee739e6c32f66583616dc2a2f5d43650332eea434
0x4396e28197653d0c244d95f8c1e57da902a72b4e is slashed by 0x7ae2f5b9e386cd1b50a4550696d957cb4900f03a at 22127411

Both were at height of 200*N + 11 which are around mining set switching block.

My guess is, there is something misaligned of mining set among active validators thus the slasher considers the slashee as active miner while not.

It's difficult to reproduce, because it doesn't necessarily happen when this pattern is met.

48ClubIan:
This should be fixed by https://github.com/bnb-chain/bsc/pull/1145

PR #1145, opened 2022-10-19
parlia urgent: BEP-131 bug. Cause new validators on blockNumber%Epoch = 11 are slashed often. 

Jolly23:
### Description

Inside parlia snapshot, BEP-131 validator set is updated at `number%s.config.Epoch == uint64(len(snap.Validators)/2)`.

Simplify it is `blockNumber%200 = 10`

`parlia.SignRecently()` is the first function be called when `miner.worker` getting NewChainHead signal, which controls the starting of the packing tasks.

```go
            case head := <-w.chainHeadCh:
			if !w.isRunning() {
				continue
			}
			clearPending(head.Block.NumberU64())
			timestamp = time.Now().Unix()
			if p, ok := w.engine.(*parlia.Parlia); ok {
				signedRecent, err := p.SignRecently(w.chain, head.Block.Header())
				if err != nil {
					log.Info("Not allowed to propose block", "err", err)
					continue
				}
				if signedRecent {
					log.Info("Signed recently, must wait")
					continue
				}
			}
			commit(true, commitInterruptNewHead)
```

But there is bug inside the `parlia.SignRecently()` causes the newValidatorSet is not updated on `blockNumber%200 = 10`

```go
func (p *Parlia) SignRecently(chain consensus.ChainReader, parent *types.Header) (bool, error) {
	snap, err := p.snapshot(chain, parent.Number.Uint64(), parent.ParentHash, nil)
```

Obviously we cannot find a block by mismatched blockNumber and blockHash. In the line above they are `parent.Number.Uint64()` and `parent.ParentHash`.

However the `parent.ParentHash` hits a cache of parlia.Snapshot. The cached snapshot cannot `snap.apply` the latest chain head.

Thus the very import part [*ChangeValidatorSet*](https://github.com/BNB48Club/bsc/blob/6e08fc046ab4a7ee3aa0c9780310a1cbf1b5f884/consensus/parlia/snapshot.go#L191) doesn't run on a RecentSigned validator and new imported validator of current Epoch.

### Rationale

 BNB48 Club reported some related cases. 

### Example

slashlist

#### recent slash list, `blockNumber%Epoch = 11` occur frequently

slashrecord
#### The slashed validator 

epoch
### Changes

```go
func (p *Parlia) SignRecently(chain consensus.ChainReader, parent *types.Block) (bool, error) {
	snap, err := p.snapshot(chain, parent.NumberU64(), parent.Hash(), nil)
```

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -929,8 +929,8 @@ func (p *Parlia) IsLocalBlock(header *types.Header) bool {
 	return p.val == header.Coinbase
 }
 
-func (p *Parlia) SignRecently(chain consensus.ChainReader, parent *types.Header) (bool, error) {
-	snap, err := p.snapshot(chain, parent.Number.Uint64(), parent.ParentHash, nil)
+func (p *Parlia) SignRecently(chain consensus.ChainReader, parent *types.Block) (bool, error) {
+	snap, err := p.snapshot(chain, parent.NumberU64(), parent.Hash(), nil)
 	if err != nil {
 		return true, err
 	}
@@ -941,7 +941,7 @@ func (p *Parlia) SignRecently(chain consensus.ChainReader, parent *types.Header)
 	}
 
 	// If we're amongst the recent signers, wait for the next block
-	number := parent.Number.Uint64() + 1
+	number := parent.NumberU64() + 1
 	for seen, recent := range snap.Recents {
 		if recent == p.val {
 			// Signer is among recents, only wait if the current block doesn't shift it out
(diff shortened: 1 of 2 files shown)


brilliant-lx (approved):
Nice catch, thank you for your contribution.

qinglin89 (dismissed):

Jolly23:
related issue #1133

guagualvcha (dismissed):

guagualvcha (approved):

brilliant-lx (approved):

qinglin89 (approved):

PR #1151, opened 2022-10-28
fair: a hotfix to include more txs before seal block

setunapo:
### Description
As the mining phase, the only criteria for a validator is to broadcast the block on time.
It can keep mining as long as the 3s deadline and GasLimit are not reached.

This is a temporary hot fix to give validator more time to commit transactions, especially when network is not busy.

And the complete network fairness is an open topic, more efforts are need to settle it.

### Changes
It changes the logic of packing transaction in mine phase.
It would be better for validator to apply this change, it helps for APR.


qinglin89:
we do have a resubmit logic to pull in  higher priced transactions with a dynamic resubmit time

brilliant-lx:
> It is a controversial of the resubmit logic, may not suitable for BSC network.

PR #1152, opened 2022-10-28
WIP: mine: a mine enhancement of packing transactions.

setunapo:
code: hot fix part2

code: set DelayLeftOver to 400ms, for block finalize and broadcast

fair: return when no time left

fix: init txs2Ch

fair: add txs2HotfixLoop to avoid flush txsCh on mainLoop

fair: subscribe txs event before commitTransactions

log: improve print log

### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -209,7 +209,6 @@ type worker struct {
 	newWorkCh          chan *newWorkReq
 	getWorkCh          chan *getWorkReq
 	taskCh             chan *task
-	taskinterruptCh    chan struct{}
 	resultCh           chan *types.Block
 	startCh            chan struct{}
 	exitCh             chan struct{}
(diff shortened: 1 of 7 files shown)


Jolly23:
bad idea

miner/worker.go
```diff
@@ -1129,14 +1166,39 @@ func (w *worker) fillTransactions(interrupt *int32, env *environment) {
 	}
 	if len(localTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, localTxs, env.header.BaseFee)
-		if w.commitTransactions(env, txs, interrupt) {
+		reason := w.commitTransactions(env, txs, interrupt)
+		if reason == commitTxsNewHead || reason == commitTxsNoGas || reason == commitTxsNoTime {
```
guagualvcha, line 1170: why not use `reason  != commitTxsDone`?

guagualvcha (approved):

guagualvcha (changes requested):
The transaction auction mechanism may break

PR #1166, opened 2022-11-02
miner: disable enforceTip when get txs from txpool

qinglin89:
### Description

`BSC` doesn't enable `tip` mechanism

### Rationale

NA

### Example

NA

### Changes

NA


brilliant-lx (approved):
BSC would not accept EIP1559 transactions, will reject these transaction in TxPool.
Either set it to true or false has the same result, since BSC should not have these kind of transactions.
And set it to false, could save some unnecessary check.

guagualvcha (approved):

guagualvcha:
Nice catch

PR #1174, opened 2022-11-10
worker: some code enhancement on work.go

setunapo:
### Description
The code of work.go is critical for BSC block generation.
Try to do some code enhancement for it:
1.the resubmit intervalAdjust code was for PoW consensus, can be deleted.
2.change the work interrupt from atomic check to channel communication.

### Rationale

With PoW, there will be a periodic timer to check if it is the time to stop packing transaction and start calculating the desired hash value, since other miner could succeed in hash compute if it spends too much time packing transactions.
It will commit the current fruit to calculate root at a reasonable time.
And it will schedule a new work to get a big block if new transaction was received.

When there are too many transactions in the TxPool, the interval of the resubmit timer would be increased and vice versa.

But it is not needed with PoS consensus, since the block interval is determined in PoS and there is already a timer to stop packing for too long.

### Example
None

### Changes
Remove unused worker code for BSC.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -53,27 +53,12 @@ const (
 	// chainSideChanSize is the size of channel listening to ChainSideEvent.
 	chainSideChanSize = 10
 
-	// resubmitAdjustChanSize is the size of resubmitting interval adjustment channel.
-	resubmitAdjustChanSize = 10
-
 	// sealingLogAtDepth is the number of confirmations before logging successful mining.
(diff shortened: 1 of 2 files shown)


qinglin89 (dismissed):

brilliant-lx (approved):

bitdaddy89 (approved):

KeefeL (approved):
LGTM

guagualvcha (approved):

PR #1182, opened 2022-11-15
Parlia: Some updates of the miner worker

setunapo:
### Description
Some update of the worker.go

1.Delay() with DelayLeftOver 
Right now, DelayLeftOver is used to reserve time for block finalize, not block broadcast.
The general block generation could be described as:
` |- fillTransactions -|- finalize a block -|- wait until the period(3s) reached -|- broadcast -|`
And the code does not work as expected, some code improve on the Delay() logic.

2.no need to add transaction to the current work if mining is stopped.
It seems that it is unnecessary to add transaction when mining is stopped.
It is a old logic of PoW, which is no longer need for BSC.

### Rationale
none

### Example
none

### Changes
none

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -65,7 +65,6 @@ const (
 )
 
 var (
-	commitTxsTimer     = metrics.NewRegisteredTimer("worker/committxs", nil)
 	writeBlockTimer    = metrics.NewRegisteredTimer("worker/writeblock", nil)
 	finalizeBlockTimer = metrics.NewRegisteredTimer("worker/finalizeblock", nil)
 )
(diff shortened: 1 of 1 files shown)


qinglin89 (approved):

KeefeL (approved):
LGTM

miner/worker.go
```diff
@@ -1048,15 +1016,24 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment) {
...
+	if delay != nil {
+		stopTimer = time.NewTimer(*delay)
+		log.Debug("Time left for mining work", "delay", delay.String())
+		defer stopTimer.Stop()
+	}
+
 	if len(localTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, localTxs, env.header.BaseFee)
-		if w.commitTransactions(env, txs, interruptCh) {
+		if w.commitTransactions(env, txs, interruptCh, stopTimer) {
```
guagualvcha, line 1030:
```
	x := time.NewTimer(time.Second)
	<-x.C
	fmt.Println("read delay 1")
	<-x.C
	fmt.Println("read delay 2")
```

It turns out that `fmt.Println("read delay 2")` is not reachable. 

So if the timer trigger when `commitTransactions` for local tx, then dead lock will happen.

PR #1186, opened 2022-11-17
worker: enhancement of the current block generation logic.

setunapo:
### Description
Currently, validator only try once to get transactions from TxPool to produce the block.
However, new transactions could arrive while the validator is committing transaction.
Validator should be allowed to add these new arrived transactions as long as Header.Timestamp is not reached.

### Rationale
This PR will:
- commitTransactions return with error code.
- drop current mining block on new block imported or resubmit timer(default 10s) is triggered.
- try fillTransactions several times for the best, not use append mode to follow the GasPrice rule.
- an algorithm to determine new fillTransactions, based on: the number of new arrived transaction, the time left, the previous cost of fillTransactions, to avoid do fillTransactions too frequently.

And the general workflow of new implementation could be described as:
image

### Example
NA.

### Changes
It could effect the mining logic, validators will be update to align the mine logic.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1063,6 +1063,10 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
 	defer stopTimer.Stop()
 	<-stopTimer.C // discard the initial tick
 
+	stopWaitTimer := time.NewTimer(0)
+	defer stopWaitTimer.Stop()
+	<-stopWaitTimer.C // discard the initial tick
+
 	// validator can try several times to get the most profitable block,
 	// as long as the timestamp is not reached.
 	workList := make([]*environment, 0, 10)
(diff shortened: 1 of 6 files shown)


miner/worker.go
```diff
@@ -1000,26 +1012,23 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment) {
...
+		err = w.commitTransactions(env, txs, interruptCh, stopTimer)
+		if err == errBlockInterruptedByNewHead || err == errBlockInterruptedByOutOfGas || err == errBlockInterruptedByTimeout {
 			return
 		}
 	}
 	if len(remoteTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, remoteTxs, env.header.BaseFee)
-		if w.commitTransactions(env, txs, interruptCh, stopTimer) {
+		err = w.commitTransactions(env, txs, interruptCh, stopTimer)
+		if err == errBlockInterruptedByNewHead || err == errBlockInterruptedByOutOfGas || err == errBlockInterruptedByTimeout {
```
qinglin89, line 1026: this `if` seems useless here, would return anyway

miner/worker.go
```diff
@@ -1000,26 +1012,23 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment) {
...
-		log.Debug("Time left for mining work", "delay", delay.String())
-		defer stopTimer.Stop()
-	}
-
+	err = nil
 	if len(localTxs) > 0 {
 		txs := types.NewTransactionsByPriceAndNonce(env.signer, localTxs, env.header.BaseFee)
-		if w.commitTransactions(env, txs, interruptCh, stopTimer) {
+		err = w.commitTransactions(env, txs, interruptCh, stopTimer)
+		if err == errBlockInterruptedByNewHead || err == errBlockInterruptedByOutOfGas || err == errBlockInterruptedByTimeout {
```
guagualvcha, line 1019: Please comment why `errBlockInterruptedByRecommit` can continue, otherwise it is confusing.

miner/worker.go
```diff
@@ -1049,24 +1058,150 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
+	workList := make([]*environment, 0, 10)
+	var bestWork *environment
+	// workList clean up
+	defer func() {
+		for _, w := range workList {
+			// only keep the best work, discard others.
+			if w == bestWork {
+				continue
+			}
+			w.discard()
```
guagualvcha, line 1081: `w.discard()` will stop prefetcher, I think we should stop prefetcher once the mining is done for each block, no need to wait.

guagualvcha: In this way, we don't have to keep `workList`

miner/worker.go
```diff
@@ -1049,24 +1058,150 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
+		}
+	}()
+LOOP:
+	for {
+		work, err := w.prepareWork(&generateParams{
```
guagualvcha, line 1086: `prepareWork` can do for only one time, like we get `baseenv`, and copy it for each task.

miner/worker.go
```diff
@@ -1049,24 +1058,150 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
+		err = w.fillTransactions(interruptCh, work, stopTimer)
+		fillDuration := time.Since(fillStart)
+		switch {
+		case errors.Is(err, errBlockInterruptedByNewHead):
+			// For Parlia, it will drop the work on receiving new block if it is not inturn.
+			if w.engine.DropOnNewBlock(work.header) {
+				log.Debug("drop the block, when new block is imported")
+				return
+			}
+		case errors.Is(err, errBlockInterruptedByTimeout):
```
guagualvcha, line 1126: we can combine the handler of `errBlockInterruptedByTimeout` and `errBlockInterruptedByOutOfGas`

miner/worker.go
```diff
@@ -1049,24 +1058,150 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
+		if interruptCh == nil || stopTimer == nil {
+			// it is single commit work, no need to try several time.
+			log.Info("commitWork interruptCh or stopTimer is nil")
+			break
+		}
+
+		newTxsNum := 0
+		// stopTimer was the maximum delay for each fillTransactions
+		// but now it is used to wait until (head.Time - DelayLeftOver) is reached.
+		stopTimer.Reset(time.Until(time.Unix(int64(work.header.Time), 0)) - w.config.DelayLeftOver)
```
guagualvcha, line 1144: Is it possible that `stopTimer` reset a very large value if `time.Until(time.Unix(int64(work.header.Time), 0))` is small than `w.config.DelayLeftOver`?

miner/worker.go
setunapo: accepted.

miner/worker.go
setunapo: `time.Until(time.Unix(int64(work.header.Time) < w.config.DelayLeftOver`
it will be negative, stopTimer will be fired immediately.

miner/worker.go
setunapo: accept

miner/worker.go
setunapo: yes, we can remove the `workList`, but it may be better not to discard the bestWork. `TriePrefetch` may not be finished, better to let it keep working, so it will be faster to do block finalize.

miner/worker.go
setunapo: yes, but we have to recreate the StateDB each time.

guagualvcha (approved):

miner/worker.go
```diff
@@ -1049,24 +1058,150 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
+		defer sub.Unsubscribe()
+
+		// Fill pending transactions from the txpool
+		fillStart := time.Now()
+		err = w.fillTransactions(interruptCh, work, stopTimer)
+		fillDuration := time.Since(fillStart)
+		switch {
+		case errors.Is(err, errBlockInterruptedByNewHead):
+			// For Parlia, it will drop the work on receiving new block if it is not inturn.
+			if w.engine.DropOnNewBlock(work.header) {
```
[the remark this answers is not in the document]
setunapo: ok, will drop the current mining block, as long as new block is imported.

guagualvcha (approved):

bitdaddy89 (approved):

PR #1201, opened 2022-11-24
worker: add double sign check for safety.

brilliant-lx:
### Description
Add double sign check for safety.

### Rationale
The current 4 mining loops could have asynchronous risk of mining block with same height in rare case.
Keep recently mined blocks to avoid double sign for safety

### Example
NA

### Changes
NA

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -670,17 +670,28 @@ func (w *worker) resultLoop() {
 			}
 
 			if prev, ok := w.recentMinedBlocks.Get(block.NumberU64()); ok {
-				prevParent, _ := prev.(common.Hash)
-				if prevParent == block.ParentHash() {
-					log.Error("Reject Double Sign!!", "block", block.NumberU64(),
-						"hash", block.Hash(), "root", block.Root(),
-						"ParentHash", block.ParentHash(), "prevParent", prevParent)
+				doubleSign := false
+				prevParents, _ := prev.([]common.Hash)
+				for _, prevParent := range prevParents {
+					if prevParent == block.ParentHash() {
+						log.Error("Reject Double Sign!!", "block", block.NumberU64(),
+							"hash", block.Hash(),
+							"root", block.Root(),
+							"ParentHash", block.ParentHash())
+						doubleSign = true
+						break
+					}
+				}
+				if doubleSign {
 					continue
 				}
+				prevParents = append(prevParents, block.ParentHash())
+				w.recentMinedBlocks.Add(block.NumberU64(), prevParents)
+			} else {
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -661,6 +669,17 @@ func (w *worker) resultLoop() {
...
+			if prev, ok := w.recentMinedBlocks.Get(block.NumberU64()); ok {
+				prevParent, _ := prev.(common.Hash)
+				if prevParent == block.ParentHash() {
+					log.Error("Reject Double Sign!!", "block", block.NumberU64(),
+						"hash", block.Hash(), "root", block.Root(),
+						"ParentHash", block.ParentHash(), "prevParent", prevParent)
+					continue
+				}
+			}
+			w.recentMinedBlocks.Add(block.NumberU64(), block.ParentHash())
```
realuncle, line 681: In most cases,  the item in the LRU cache never been used, so when we pop the item from cache , the order is  not certain.

brilliant-lx: as noted in the comment

miner/worker.go
```diff
@@ -661,6 +669,19 @@ func (w *worker) resultLoop() {
...
+				if prevParent == block.ParentHash() {
+					log.Error("Reject Double Sign!!", "block", block.NumberU64(),
+						"hash", block.Hash(), "root", block.Root(),
+						"ParentHash", block.ParentHash(), "prevParent", prevParent)
+					continue
+				}
+			}
+			// Add will call removeOldest internally to remove the oldest element
+			// if the LRU Cache is full
+			w.recentMinedBlocks.Add(block.NumberU64(), block.ParentHash())
```
realuncle, line 683: There is some corner case here:
1. validator propose block based on a parent blockA at height H, it produce BlockB at height H+1;
2. then a possible reorg may happen;
3. validator propose block based on a parent blockA' at height H, it produce BlockB' at height H+1;
4. the cache at H+1 been overwrite;
5. then a possible reorg may happen;
6. validator propose block based on a parent blockA at height H, it produce BlockC at height H+1; this is no detected by the watch dog.

realuncle (changes requested):

miner/worker.go
realuncle: It is a bit tricky, but any easy way to solve this?

guagualvcha (approved):

miner/worker.go
brilliant-lx: yea, use slice to keep all the history parent blocks, avoid overwritten.

qinglin89 (approved):

j75689 (approved):

yutianwu (approved):

PR #1203, opened 2022-11-25
Prune some useless code

qinglin89:
### Description

This pr is about some code prunning.

### Rationale

Mainly remove some useless code.

### Example

N/A

### Changes

Notable changes: 
* remove useless `waitgroup`
* remove useless `nil` check

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1423,11 +1423,8 @@ func (s *StateDB) Commit(failPostCommitFunc func(), postCommitFuncs ...func() er
 			finishCh := make(chan struct{})
 
 			threads := gopool.Threads(len(s.stateObjectsDirty))
-			wg := sync.WaitGroup{}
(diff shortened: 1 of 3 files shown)


core/state/statedb.go
```diff
@@ -1423,11 +1423,8 @@ func (s *StateDB) Commit(failPostCommitFunc func(), postCommitFuncs ...func() er
 			finishCh := make(chan struct{})
 
 			threads := gopool.Threads(len(s.stateObjectsDirty))
-			wg := sync.WaitGroup{}
```
brilliant-lx, line 1426: why it can be removed?

qinglin89: there's a `read loop`, range for `tasksNums` times, makes it just work as synchronized, no need to `Wait` before the last return.

brilliant-lx (approved):
LGTM

PR #1204, opened 2022-11-25
worker enhancement: NewTxsEvent and triePrefetch reuse

setunapo:
### Description
These two PRs is to enhance worker, after multi-fillTransactions(https://github.com/bnb-chain/bsc/pull/1186) was introduced.
The node could have stress to handle huge transaction traffic.
This PR tries to reduce some CPU cost and improve the logic of NewTxsEvent Subscribe.

### Rationale
**Reused triePrefetch** 
1.reused triePrefetch when multi-fillTransactions to mine a block, to avoid too much trie prefetch routines when several fillTransactions are called.

**Subscribe NewTxsEvent**
 1.remove the unnecessary NewTxsEvent subscriber, which was used for PoW resubmit check.
 2.unsubscribe ASAP before another fillTransactions, to avoid block others.
 3.double txChanSize to **4096 * 2**, so it will less likely to block others when it is running fillTransactions(), which may take 1.5s at most.

### Example
NA

### Changes
NA


guagualvcha (approved):

miner/worker.go
```diff
@@ -1105,27 +1084,29 @@ func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
...
 	defer func() {
-		for _, w := range workList {
+		for _, wk := range workList {
 			// only keep the best work, discard others.
-			if w == bestWork {
+			if wk == w.current {
 				continue
 			}
-			w.discard()
+			wk.discard()
```
qinglin89, line 1095: only one `triePrefetcher` be reused and `transfered`, this `range` may not necessary

brilliant-lx: the `discard()` logic could be changed, although it only stop triePrefecth right now, it could do something later.

qinglin89 (approved):

PR #1217, opened 2022-11-29
build-info causes prom error

nathanhopp:
### Description

Prometheus cannot parse `build-info` which causes metrics not to be exported properly.

### Changes

Notable changes: 
* Change `build-info` -> `build_info`


j75689:
Thanks, but I think we should modify the mutateKey function in `metrics/prometheus` to replace `-` with `_` because this rule should only work for prometheus.

https://github.com/bnb-chain/bsc/blob/03ff2992ab4674c1df8f568ee9a31596f3503c26/metrics/prometheus/collector.go#L128
```go
func mutateKey(key string) string {
	key = strings.Replace(key, "/", "_", -1)
	key = strings.Replace(key, "-", "_", -1)
	return key
}
```

nathanhopp:
> Please refine your commit, it includes some other unrelated commits

Thanks, updated. There's an odd test failing `TestEth2AssembleBlock`, though it was successful in previous commit and locally. I can't trigger a rerun myself

Pythonberg1997 (approved):

j75689 (approved):

anyshy:
expected label name, got "INVALID"

xuboatom:
build_info{"architecture":"amd64",  ,label format is INVALID,   builld_info{"architecture"="amd64".......}

PR #1247, opened 2022-12-20
pruneancient_comments

joeylichang:
### Description

In this pr, some comments about ancient data prune are added.
* Enriched the description of the pruneancient startup parameter, emphasizing that it is not enabled by default, and only the latest 9w block data is kept.
*  Enriched the note of the PruneAncientData confg, emphasizing that irreversible and what kinds of data will be pruned,

### Rationale

Currently, some node operators misunderstand pruneancient flag, which leads to difficulty of maintaining nodes. By adding these comments, it will be easier for them to maintain nodes.

### Example

NA

### Changes

Notable changes: 
* add comments

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -34,11 +34,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/fatih/structs"
-	pcsclite "github.com/gballet/go-libpcsclite"
-	gopsutil "github.com/shirou/gopsutil/mem"
-	"gopkg.in/urfave/cli.v1"
-
 	"github.com/ethereum/go-ethereum/accounts"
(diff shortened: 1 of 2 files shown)


cmd/utils/flags.go
```diff
@@ -74,6 +69,9 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/nat"
 	"github.com/ethereum/go-ethereum/p2p/netutil"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/fatih/structs"
+	pcsclite "github.com/gballet/go-libpcsclite"
+	gopsutil "github.com/shirou/gopsutil/mem"
```
brilliant-lx, line 74: irrelevent?

PR #1268, opened 2023-01-06
parlia: consensus changes for Early broadcast 

kyrie-yl:
### Description

BEP188 introduces a new block broadcast strategy for validators to improve the network capacity and stability. As this BEP allows in-turn validators broadcast mined block earlier, consensus need to adapt to it.

### Rationale

Currently, a mined block will not be broadcasted until the timestamp header.Time is reached, even though the mining process is finished. If an in-turn validator spends too much time in the importing phase, it may not have enough time to mine its own block. In extreme cases, it could generate an empty block or even miss its block. This could be a challenge to the stability of the blockchain.With this BEP, validators could begin to import a block earlier and have more time to mine a new block.

The current consensus engine of the BNB chain does not accept future blocks, which means BSC nodes will reject the block if it is received earlier than its timestamp defined in header.Time. So we need to change the header’s verification logic in the consensus engine as well.
As the broadcast strategy changes, the logic of verifyHeader must be changed accordingly.The new logic should be like this shown in the figure below.
consensus_adaption

### Example
None

### Changes

Notable changes: 
* verifyHeader adapt to the BEP188

### DIFF
--- a/consensus/errors.go
+++ b/consensus/errors.go
@@ -31,6 +31,10 @@ var (
 	// to the current node.
 	ErrFutureBlock = errors.New("block in the future")
 
+	// ErrFutureParentBlock is returned when a block's parent's timestamp is in the future
+	// according to the current node.
+	ErrFutureParentBlock = errors.New("parent block in the future")
+
 	// ErrInvalidNumber is returned if a block's number doesn't equal its parent's
 	// plus one.
 	ErrInvalidNumber = errors.New("invalid block number")
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -327,9 +327,26 @@ func (p *Parlia) verifyHeader(chain consensus.ChainHeaderReader, header *types.H
 	}
 	number := header.Number.Uint64()
 
-	// Don't waste time checking blocks from the future
+	// According to BEP188, after Bohr fork, an in-turn validator is allowed to broadcast
+	// a mined block earlier but not earlier than its parent's timestamp when the block is ready .
 	if header.Time > uint64(time.Now().Unix()) {
-		return consensus.ErrFutureBlock
+		if !chain.Config().IsBohr(header.Number) || header.Difficulty.Cmp(diffInTurn) != 0 {
+			return consensus.ErrFutureBlock
+		}
+		var parent *types.Header
+		if len(parents) > 0 {
+			parent = parents[len(parents)-1]
+		} else {
+			parent = chain.GetHeader(header.ParentHash, number-1)
+		}
+
+		if parent == nil || parent.Number.Uint64() != number-1 || parent.Hash() != header.ParentHash {
+			return consensus.ErrUnknownAncestor
+		}
+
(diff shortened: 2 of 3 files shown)


params/config.go
```diff
@@ -299,6 +299,7 @@ var (
 		NanoBlock:           big.NewInt(21962149),
 		MoranBlock:          big.NewInt(22107423),
 		GibbsBlock:          big.NewInt(23846001),
+		BohrBlock:           big.NewInt(22107423), // todo: TBD
```
brilliant-lx, line 302: set it to nil, since the target block is not determined.

brilliant-lx:
pls follow the commit format.

guagualvcha (changes requested):
Please follow the standard to write the PR description.

consensus/errors.go
```diff
@@ -31,6 +31,10 @@ var (
 	// to the current node.
 	ErrFutureBlock = errors.New("block in the future")
 
+	// ErrBroadcastTooEarly is returned when a block's parent's timestamp is in the future
+	// according to the current node.
+	ErrBroadcastTooEarly = errors.New("block is broadcast too early")
```
guagualvcha, line 36: "block is broadcast too early" ==> "unexpected early broadcasted block"

consensus/parlia/parlia.go
```diff
@@ -327,9 +327,26 @@ func (p *Parlia) verifyHeader(chain consensus.ChainHeaderReader, header *types.H
 	}
 	number := header.Number.Uint64()
 
-	// Don't waste time checking blocks from the future
+	// According to BEP-?, after Bohr fork, an in-turn validator is allowed to broadcast
```
guagualvcha, line 330: `BEP-?` ==> `BEP188`

params/config.go
```diff
@@ -325,6 +326,7 @@ var (
 		GibbsBlock:          big.NewInt(22800220),
 		NanoBlock:           big.NewInt(23482428),
 		MoranBlock:          big.NewInt(23603940),
+		BohrBlock:           big.NewInt(23603940), // todo: TBD
```
guagualvcha, line 329: So do here.

consensus/parlia/parlia.go
```diff
@@ -327,9 +327,26 @@ func (p *Parlia) verifyHeader(chain consensus.ChainHeaderReader, header *types.H
...
+			parent = chain.GetHeader(header.ParentHash, number-1)
+		}
+
+		if parent == nil || parent.Number.Uint64() != number-1 || parent.Hash() != header.ParentHash {
+			return consensus.ErrUnknownAncestor
+		}
+
+		if parent.Time > uint64(time.Now().Unix()) {
+			return consensus.ErrBroadcastTooEarly
+		}
```
guagualvcha, line 349: If we have a hard fork here, why don't we implement the early broadcast logic in this PR?

brilliant-lx (approved):

guagualvcha (approved):

Issue #1286, opened 2023-01-29
Parlia slice bounds out of range


Davincible:
```
nft-marketplace-bsc-miner-1  | panic: runtime error: slice bounds out of range [:-65]
nft-marketplace-bsc-miner-1  | 
nft-marketplace-bsc-miner-1  | goroutine 102 [running]:
nft-marketplace-bsc-miner-1  | github.com/ethereum/go-ethereum/consensus/parlia.(*Parlia).snapshot(0xc000976f00, {0x1ce6d20, 0xc000977400}, 0x8?, {0xa3, 0xfa, 0x5d, 0xc3, 0x93, 0xe4, ...}, ...)
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/consensus/parlia/parlia.go:464 +0xd0a
nft-marketplace-bsc-miner-1  | github.com/ethereum/go-ethereum/consensus/parlia.(*Parlia).Prepare(0xc000976f00, {0x1ce6d20, 0xc000977400}, 0xc000968480)
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/consensus/parlia/parlia.go:602 +0xf8
nft-marketplace-bsc-miner-1  | github.com/ethereum/go-ethereum/miner.(*worker).prepareWork(0xc00064a480, 0xc0022c9b70)
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/miner/worker.go:974 +0x9a5
nft-marketplace-bsc-miner-1  | github.com/ethereum/go-ethereum/miner.(*worker).commitWork(0xc00064a480, 0xc002039650, 0x63d6e8ab)
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/miner/worker.go:1101 +0x3a5
nft-marketplace-bsc-miner-1  | github.com/ethereum/go-ethereum/miner.(*worker).mainLoop(0xc00064a480)
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/miner/worker.go:479 +0x745
nft-marketplace-bsc-miner-1  | created by github.com/ethereum/go-ethereum/miner.newWorker
nft-marketplace-bsc-miner-1  | 	github.com/ethereum/go-ethereum/miner/worker.go:276 +0x7b6
```

brilliant-lx:
investigating.

Davincible:
[...]

PR #1294, opened 2023-02-02
parlia : add a check for the length of extraData.

zzphust:
### Description

add a check for the length of extraData
### Rationale

#1286 
### Example

not related
### Changes

Notable changes:

- consensus/parlia/parlia.go

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -461,6 +461,12 @@ func (p *Parlia) snapshot(chain consensus.ChainHeaderReader, number uint64, hash
 				// get checkpoint data
 				hash := checkpoint.Hash()
 
+				if len(checkpoint.Extra) < extraVanity {
+					return nil, errMissingVanity
+				}
+				if len(checkpoint.Extra) < extraVanity+extraSeal {
+					return nil, errMissingSignature
+				}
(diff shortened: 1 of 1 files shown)


qinglin89:
The logic was fine, but do we really need to deal with this case since this was a `wrong genesis`.
For example, after the vanity and signature check, if user didn't set validator right(like 0), then the system would not be set up successfully (all validator unauthorized)

brilliant-lx (approved):

qinglin89 (dismissed):

brilliant-lx (changes requested):
maybe we only need to add a check and print a error log. then crash

consensus/parlia/parlia.go
```diff
@@ -461,6 +461,12 @@ func (p *Parlia) snapshot(chain consensus.ChainHeaderReader, number uint64, hash
 				// get checkpoint data
 				hash := checkpoint.Hash()
 
+				if len(checkpoint.Extra) < extraVanity {
```
guagualvcha, line 464: Seems this check is unnecessary, because the following one is more restrict.

guagualvcha (approved):

brilliant-lx (approved):

PR #1301, opened 2023-02-05
fix: p2p sync with lagging peer

setunapo:
no need to sync with lagging peer, which could make the local chain stalling as well.

### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/downloader/downloader.go
+++ b/eth/downloader/downloader.go
@@ -68,6 +68,7 @@ var (
 var (
 	errBusy                    = errors.New("busy")
 	errUnknownPeer             = errors.New("peer is unknown or unhealthy")
+	errLaggingPeer             = errors.New("peer is lagging")
 	errBadPeer                 = errors.New("action from bad peer ignored")
 	errStallingPeer            = errors.New("peer is stalling")
 	errUnsyncedPeer            = errors.New("unsynced peer")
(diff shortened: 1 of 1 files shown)


ponury1990 (approved):

ponury1990 (approved):

ponury1990 (approved):

ponury1990 (approved):

ponury1990 (dismissed):

eth/downloader/downloader.go
```diff
@@ -503,22 +504,38 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td *big.I
 		// nil panics on an access.
 		pivot = d.blockchain.CurrentBlock().Header()
 	}
-	height := latest.Number.Uint64()
 
-	origin, err := d.findAncestor(p, latest)
+	// If the remote peer is lagging behind, no need to sync with it, drop the peer.
+	remoteHeight := remoteHeader.Number.Uint64()
+	var localHeight uint64
+	switch mode {
```
guagualvcha, line 511: image

I am curious why the condition here does not stop the syncing from a lag peer.

guagualvcha (changes requested):
Still something not clear.

eth/downloader/downloader.go
brilliant-lx: there could be 2 scenarios:
1.the Peer broadcast its state with a invalid TD, could be a very big TD, but it does not have the block.
2.local node imported new blocks, between `nextSyncOp` and `syncWithPeer, fetchHead`, it is likely to happen, since block fetch and download are 2 asynchronous operation.

ponury1990 (approved):

eth/downloader/peer.go
```diff
@@ -75,6 +76,7 @@ type lightPeerWrapper struct {
 }
 
 func (w *lightPeerWrapper) Head() (common.Hash, *big.Int) { return w.peer.Head() }
+func (w *lightPeerWrapper) MarkLagging()                  { w.peer.MarkLagging() }
```
guagualvcha, line 79: `MarkLagging` can be interface of a `Peer` rather than `LightPeer`

guagualvcha (dismissed):

guagualvcha:
LGTM, but rerun the UT please.

GalaIO (approved):
LGTM

GalaIO (approved):
LGTM

guagualvcha (approved):

PR #1322, opened 2023-02-24
cmd/utils/flags.go: --diffsync flag is deprecate.

blxdyx:
### Description

deprecate diffsync flag

### Rationale

When the time for the handshake is around 10s, when sending a message to the diffwait channel at the same time, the waitDiffExtension enters the timeout logic, which will cause a deadlock.

### Example

no example
### Changes

Notable changes: 
* flag.go:deprecated --diffsync flag and add a warning about it


guagualvcha (changes requested):
`handshakeTimeout` is not a ticker but a timer, so the change seems not make too much sense.

brilliant-lx (approved):

guagualvcha (approved):

PR #1329, opened 2023-03-01
refactor the code in func backOffTime()

brilliant-lx:
### Description
no behavior change, only make the code more readable

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -1346,12 +1346,14 @@ func (p *Parlia) backOffTime(snap *Snapshot, header *types.Header, val common.Ad
 			}
 
 			// Exclude the recently signed validators
-			for index, addr := range validators {
+			temp := make([]common.Address, 0, len(validators))
+			for _, addr := range validators {
 				if _, ok := recentsMap[addr]; ok {
-					validators = append(validators[:index], validators[index+1:]...)
-					break
+					continue
 				}
+				temp = append(temp, addr)
 			}
+			validators = temp
 		}
 
 		// get the index of current validator and its shuffled backoff time.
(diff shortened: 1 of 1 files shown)


consensus/parlia/parlia.go
```diff
@@ -1321,59 +1321,67 @@ func (p *Parlia) backOffTime(snap *Snapshot, header *types.Header, val common.Ad
 	if snap.inturn(val) {
 		return 0
 	} else {
-		idx := snap.indexOfVal(val)
+		delay := initialBackOffTime
+		validators := snap.validators()
+		if p.chainConfig.IsBohr(header.Number) {
+			// reverse the key/value of snap.Recents to get recentsMap
+			recentsMap := make(map[common.Address]uint64, len(snap.Recents))
```
BetZzzzz, line 1328: block height is no need here, value filed could just be struct{}

consensus/parlia/parlia.go
```diff
@@ -1321,59 +1321,67 @@ func (p *Parlia) backOffTime(snap *Snapshot, header *types.Header, val common.Ad
...
+					continue
+				}
+				temp = append(temp, addr)
+			}
+			validators = temp
+		}
+
+		// get the index of current validator and its shuffled backoff time.
+		idx := -1
+		for index, itemAddr := range validators {
```
BetZzzzz, line 1361: could get rid of this for loop, just calculate the `idx` in the loop where calculating updated validators(line:1350)

consensus/parlia/parlia.go
brilliant-lx: actually, it is used bellow for debug purpose.

consensus/parlia/parlia.go
brilliant-lx: but Bohr could have not reached.

BetZzzzz (approved):

NathanBSC:
lgtm

buddh0 (approved):

PR #1357, opened 2023-03-15
Integration API for EIP-4337 bundler with an L2 validator/sequencer

MatusKysel:
# Integration API for EIP-4337 bundler with an L2 validator/sequencer

## Source 
https://notes.ethereum.org/@yoav/SkaX2lS9j

## Introduction

An EIP-4337 "bundler" is a server that handles multiple UserOperations, and submit them as a single transaction to the network.

The bundler validates each UserOperation, and makes sure it is capable (and willing) to pay.

(Each UserOperation has separate "validation" and "execution" phases. the bundler only validates the "validation" phase, since payment is guaranteed even if "execution" reverts, like normal ethereum transcations)

There is a slight chance that the aggregated transaction would fail, and that can only happen if an account happen to modify its own storage between the time the bundler validated the transaction and the time it was submitted.

We need a way for a bundler to guarantee that no change was made to the network that could cause the transaction to revert.

## Specification

* A new RPC request which is an extension to normal "sendRawTransaction"
* The Method and first param have the same meaning as in normal sendRawTransaction. 
* Added a second parameter which is "options"
* within the "options" param, add the following members:
    * **knownAccounts**: a map of accounts with expected storage
        * The key is account address
        * If the value is **hex string**, it is the known storage root hash of that account.
        * If the value is an **object**, then each member is in the format of `"slot": "value"`, which are explicit slot values within that account storage.
    * **blockNumberMin**: [optional] minimal block number for inclusion
    * **blockNumberMax**: [optional] maximum block number for inclusion
    * **timestampMin**: [optional] minimum block timestamp for inclusion
    * **timestampMax**: [optional] maximum block timestamp for inclusion
* Before accepting the request, the sequencer should:
    * If block range was given, check that the block number is within the range.
    * If timestamps range was given, check that the block's timestamp is within the range.
    * For an address with a storage root hash, validate the current root is unmodified.
    * For an address with a list of slots, it should verify that all these slots hold the exact value specified.
* The sequencer should REJECT the request if any address is doesn't pass the above rules.
* The number of the slots/accounts in the knownAccounts structure is known to be below 1000 (TBD) entries.

### Return value

In case of successful inclusion, the call should return the same value as `sendRawTransaction` (namely, the transaction-hash)
In case of failure, it SHOULD return an error with indication of failure reason.
The error code SHOULD be -32003 (transaction rejected) with reason string describing the cause: storage error, out of block/time range,
In case of repeated failures or knownAccounts too large, the error code SHOULD be -32005 (Limit exceeded) with a description of the error

**NOTE:** Even if the transaction was accepted (into the internal mempool), the application MUST NOT assume it actually gets included, and must monitor the blockchain.

## Sample request:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_sendRawTransactionConditional",
    "params": [
        "0x2815c17b00...",
        {
            "blockNumberMax": 12345,
            "knownAccounts": {
                "0xadd1": "0xfedc....",
                "0xadd2": { 
                    "0x1111": "0x1234...",
                    "0x2222": "0x4567..."
                }
            }     
        } 
    ]
}
```

### DIFF
--- a/internal/jsre/deps/web3.js
+++ b/internal/jsre/deps/web3.js
@@ -5438,7 +5438,7 @@ var methods = function () {
         inputFormatter: [null]
     });
 
-    var sendRawTransaction = new Method({
+    var sendRawTransactionConditional = new Method({
         name: 'sendRawTransactionConditional',
         call: 'eth_sendRawTransactionConditional',
         params: 2,
--- a/ethclient/ethclient_test.go
+++ b/ethclient/ethclient_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/ethereum/go-ethereum/eth"
 	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/ethdb/memorydb"
+	"github.com/ethereum/go-ethereum/internal/ethapi"
 	"github.com/ethereum/go-ethereum/node"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rpc"
@@ -376,6 +377,9 @@ func TestEthClient(t *testing.T) {
 		"TestDiffAccounts": {
 			func(t *testing.T) { testDiffAccounts(t, client) },
 		},
+		"TestSendTransactionConditional": {
+			func(t *testing.T) { testSendTransactionConditional(t, client) },
+		},
 		// DO not have TestAtFunctions now, because we do not have pending block now
 	}
 
@@ -659,4 +663,46 @@ func testDiffAccounts(t *testing.T, client *rpc.Client) {
 			t.Fatalf("expect ignore all transactions, but some transaction has recorded")
 		}
 	}
+}
+
+func testSendTransactionConditional(t *testing.T, client *rpc.Client) {
+	ec := NewClient(client)
+
+	if err := sendTransactionConditional(ec); err != nil {
+		t.Fatalf("error: %v", err)
+	}
+}
+
+func sendTransactionConditional(ec *Client) error {
+	chainID, err := ec.ChainID(context.Background())
+	if err != nil {
+		return err
+	}
+
+	nonce, err := ec.PendingNonceAt(context.Background(), testAddr)
+	if err != nil {
+		return err
+	}
+
+	signer := types.LatestSignerForChainID(chainID)
+
+	tx, err := types.SignNewTx(testKey, signer, &types.LegacyTx{
+		Nonce:    nonce,
+		To:       &common.Address{2},
+		Value:    big.NewInt(1),
+		Gas:      22000,
+		GasPrice: big.NewInt(params.InitialBaseFee),
+	})
+	if err != nil {
+		return err
+	}
+
+	root := common.HexToHash("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
+	return ec.SendTransactionConditional(context.Background(), tx, ethapi.TransactionOpts{
+		KnownAccounts: map[common.Address]ethapi.AccountStorage{
+			testAddr: ethapi.AccountStorage{
+				StorageRoot: &root,
+			},
+		},
+	})
 }


internal/ethapi/transaction_options.go
```diff
@@ -0,0 +1,81 @@
...
+	if o.BlockNumberMax != nil && blockNumber > uint64(*o.BlockNumberMax) {
+		return errors.New("BlockNumberMax condition not met")
+	}
+	if o.TimestampMin != nil && timeStamp < uint64(*o.TimestampMin) {
+		return errors.New("TimestampMin condition not met")
+	}
+	if o.TimestampMax != nil && timeStamp > uint64(*o.TimestampMax) {
+		return errors.New("TimestampMax condition not met")
+	}
+	if len(o.KnownAccounts) > 1000 {
```
emailtovamos, line 55: Is `KnownAccounts` supposed to represent total number of smart contract accounts that is supported or something else?

drortirosh: "account" represent a contract which this transaction accessed the storage of.
most of them are "smart accounts", but some might be paymasters or smart-account factory contracts.
Yes, it puts an absolute maximum on the # of accounts in a handleOps bundle.
The number "1000" was chosen arbitrarily, and might change, after we do some performance analysis of implementation, to understand the impact of this size

emailtovamos: Thanks @drortirosh . 
With regards to this, how exactly are the `SlotValue`s determined? 
I assume `SlotValue` is a reference to the address of the storage of the account?

drortirosh: The slot value is the value returned by `eth_getStorageAt(address, position)`
(we just use the term "slot" and not "position")

guagualvcha:
I change the merge base from master to develop branch.

internal/ethapi/api.go
```diff
@@ -2137,6 +2137,23 @@ func (s *PublicTransactionPoolAPI) SendRawTransaction(ctx context.Context, input
...
+func (s *PublicTransactionPoolAPI) SendRawTransactionConditional(ctx context.Context, input hexutil.Bytes, opts TransactionOpts) (common.Hash, error) {
+	tx := new(types.Transaction)
+	if err := tx.UnmarshalBinary(input); err != nil {
+		return common.Hash{}, err
+	}
+	state, _, err := s.b.StateAndHeaderByNumber(ctx, rpc.BlockNumber(s.b.CurrentBlock().Number().Int64()))
+	if state == nil || err != nil {
+		return common.Hash{}, err
+	}
+	if err := opts.Check(s.b.CurrentBlock().NumberU64(), s.b.CurrentBlock().Time(), state); err != nil {
```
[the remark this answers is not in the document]
guagualvcha: Aslo suggest using `CurrentHeader` than `CurrentBlock`

internal/ethapi/transaction_options.go
```diff
@@ -0,0 +1,89 @@
...
+	if counter > 1000 {
+		return errors.New("knownAccounts too large")
+	}
+	return o.CheckOnlyStorage(statedb)
+}
+
+func (o *TransactionOpts) CheckOnlyStorage(statedb *state.StateDB) error {
+	for address, accountStorage := range o.KnownAccounts {
+		if accountStorage.RootHash != nil {
+			trie := statedb.StorageTrie(address)
```
guagualvcha, line 72: To improve the performance of BSC client, we have introduced the `separate node` feature. Please refer to: https://github.com/bnb-chain/bsc/pull/926.

The fast node does not have MPT, and thus can not open tries here. Actually, a lot of infra are running fast node now, especially RPC nodes. So need come up a way to workaround this, please try `snapshot` when Trie is not available.

internal/ethapi/transaction_options.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if o.TimestampMax != nil && timeStamp > uint64(*o.TimestampMax) {
+		return errors.New("TimestampMax condition not met")
+	}
+	counter := 0
+	for _, account := range o.KnownAccounts {
+		if account.RootHash != nil {
+			counter += 1
+		} else if account.SlotValue != nil {
+			counter += len(account.SlotValue)
```
Mister-EA, line 60: I thought the limit was on the number of known accounts, not on the number of storage slots.

internal/ethapi/transaction_options.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	if counter > 1000 {
```
Mister-EA, line 63: Probably best not to hardcode 1000.

guagualvcha:
LGTM

internal/ethapi/transaction_options_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/internal/ethapi"
+)
+
+func ptr(hash common.Hash) *common.Hash {
+	return &hash
+}
+
+func u64Ptr(v hexutil.Uint64) *hexutil.Uint64 {
```
Mister-EA, line 17: Why have this as a function instead of using `&v` directly?

internal/ethapi/transaction_options.go
emailtovamos: Hi @drortirosh 
To make BSC work with the eth-infinitism bundler, does BSC have to enable London hardfork? 
Or can I modify the bundler to make it work with the current BSC?

internal/jsre/deps/web3.js
```diff
@@ -5438,6 +5438,13 @@ var methods = function () {
         inputFormatter: [null]
     });
 
+    var sendRawTransaction = new Method({
```
emailtovamos, line 5441: Should this be `sendRawTransactionConditional` ?

emailtovamos (changes requested):

brilliant-lx (approved):

guagualvcha (approved):

PR #1360, opened 2023-03-16
.travis: use go 1.19

markya0616:
### Description

Use go 1.19 in travis.

### Rationale

The go version in go.mod should be the same in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ jobs:
   allow_failures:
     - stage: build
       os: osx
-      go: 1.17.x
+      go: 1.19.x
       env:
         - azure-osx
         - azure-ios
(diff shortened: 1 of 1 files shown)


brilliant-lx:
Thx, BTW, is this `.travis.yml` useful to you?

brilliant-lx:
Checked, we don't Travis CI, no need to update, so close it.

Lusitaniae:
Could we possibly remove it instead?

We had some bsc build automation that was reading go version from .travis.yml which is now outdated :)

GalaIO:
> Could we possibly remove it instead?
> 
> We had some bsc build automation that was reading go version from .travis.yml which is now outdated :)

Could you read go version from `go.mod` file? Here no plan remove all travis deps, and another question if `.travis.yml` file removed, how can you get the go version?

Lusitaniae:
I did that

It’s just that it’s misleading to have an old Travis config which doesn’t
work anymore I’d say

Anyway; not important, feel free to close

PR #1363, opened 2023-03-17
bugfix: failed to freeze genesis block with prunedfreezer

jsvisa:
### Description

I'm running bsc syncing from the genesis block, and failed to sync with errors below:

```
 Error writing genesis to ancients        err="this operation is not supported"
```

So let's skip the error return

### DIFF
--- a/core/rawdb/prunedfreezer.go
+++ b/core/rawdb/prunedfreezer.go
@@ -300,5 +300,5 @@ func (f *prunedfreezer) AncientRange(kind string, start, count, maxBytes uint64)
 }
 
 func (f *prunedfreezer) ModifyAncients(func(ethdb.AncientWriteOp) error) (int64, error) {
-	return 0, errNotSupported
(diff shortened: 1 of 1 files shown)


core/rawdb/prunedfreezer.go
```diff
@@ -300,5 +300,5 @@ func (f *prunedfreezer) AncientRange(kind string, start, count, maxBytes uint64)
 }
 
 func (f *prunedfreezer) ModifyAncients(func(ethdb.AncientWriteOp) error) (int64, error) {
-	return 0, errNotSupported
+	return 0, nil
```
brilliant-lx, line 303: `prunedfreezer` should not support `ModifyAncients()`, since it does not contain ancient data. So unreasonable to modify ancient.

jsvisa: thanks, seems this PR is unnecessary, I'll close it right now

PR #1382, opened 2023-03-22
consnesus/parlia: abort sealing when block in the same height has upd…

qinglin89:
### Description

 abort sealing when block in the same height has updated the canonical head while this is the `offturn` validator

### Rationale
log_validators
log_validator_upper:

- receive B_71317(...682) at 00:56:55,  and its timestamp should before 00:56:54 according to the log from the lower part of the picture (backofftime<=2, timestamp<=prev+5)
- mining B_71317(...e95), commit at 00:56:55 and elapsed time 5.194s => backofftime>=6, and the timestamp of 71316 is 00:56:54, so the backofftime of B_71317 is 3 (timestamp=prev+6)
- from line:4474, shows that at 00:56:55 B_71317(...682) has been imported and the miner.worker had received the new head event, but the B_71317(...e95) was in the sealing stage, so it couldn't be interrupt, and finally sealed and post the offturn block B_71317(...e95) which would be the ultimate canonical block after some reorgs.

log_validator_blow:
- B_71317(...682) received as canonical head at 00:56:54, and received B_71317(...e95, mining stage shown from upper log) as sidechain at 00:56:56

---

BSC have changed the logical of `backoffTime` calculation, this modification make the `backoffTime` more reasonable and expected, but it reduces the interval between `offturn` block's timestamp, this shouldn't be called as `bug`, but could increase the chance of `reorganization` especially when the network not stability or big traffic happens compared to the previous version since it made bigger average interval among `offturn` blocks' timestamp.

reorgs

### Example

N/A

### Changes

Notable changes: 
*

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -891,7 +891,7 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 				log.Info("Received block process finished, abort block seal")
 				return
 			case <-time.After(time.Duration(processBackOffTime) * time.Second):
-				if number := header.Number.Uint64(); chain.CurrentHeader().Number.Uint64() >= number && chain.GetHeaderByNumber(number-1).Hash() == (header.ParentHash) {
+				if chain.CurrentHeader().Number.Uint64() >= header.Number.Uint64() {
 					log.Info("Process backoff time exhausted, and current header has updated to abort this seal")
 					return
 				}
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -891,7 +891,7 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 				log.Info("Received block process finished, abort block seal")
 				return
 			case <-time.After(time.Duration(processBackOffTime) * time.Second):
-				if chain.CurrentHeader().Number.Uint64() == header.Number.Uint64() {
+				if number := header.Number.Uint64(); chain.CurrentHeader().Number.Uint64() >= number && chain.GetHeaderByNumber(number-1).Hash() == (header.ParentHash) {
 					log.Info("Process backoff time exhausted, and current header has updated to abort this seal")
 					return
 				}
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -891,6 +891,10 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 				log.Info("Received block process finished, abort block seal")
 				return
 			case <-time.After(time.Duration(processBackOffTime) * time.Second):
+				if chain.CurrentHeader().Number.Uint64() == header.Number.Uint64() {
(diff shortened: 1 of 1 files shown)


buddh0 (approved):

consensus/parlia/parlia.go
```diff
@@ -891,6 +891,10 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 				log.Info("Received block process finished, abort block seal")
 				return
 			case <-time.After(time.Duration(processBackOffTime) * time.Second):
+				if chain.CurrentHeader().Number.Uint64() == header.Number.Uint64() {
```
brilliant-lx, line 894: how about >=?
it could happen, in case:
- local mined has unstable network
- it mined a off-turn block, with backOffTime =4, say: T0 + 4
- its network resumed at T0 + 3, it could import 2 in-turn blocks simultaneously
- then, currentHeader would > header

NathanBSC: I think >= is better

qinglin89: It's not how it works.
There's a `delay` logic before this `shouldWaitCheck`, so it only would happen when T0-Block started to insert before T0+4, and this offturn validator would wait to T0+5 for the `verifiedHeader` block's body/state be verified and check whether it is `updated as canonical`. and it's not possible for T0+6 block imported(future block)

qinglin89: It's not how it works.
There's a delay logic before this shouldWaitCheck, so it only would happen when T0-Block started to insert before T0+4, and this offturn validator would wait to T0+5 for the verifiedHeader block's body/state be verified and check whether it is updated as canonical. and it's not possible for T0+6 block imported(future block)

qinglin89: It's not how it works.
There's a delay logic before this shouldWaitCheck, so it only would happen when T0-Block started to insert before T0+4, and this offturn validator would wait to T0+5 for the verifiedHeader block's body/state be verified and check whether it is updated as canonical. and it's not possible for T0+6 block imported(future block)

qinglin89: might insist use `==`, since `>` would happen on another fork

NathanBSC (approved):

buddh0 (dismissed):

consensus/parlia/parlia.go
```diff
@@ -891,6 +891,10 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
 				log.Info("Received block process finished, abort block seal")
 				return
 			case <-time.After(time.Duration(processBackOffTime) * time.Second):
+				if number := header.Number.Uint64(); chain.CurrentHeader().Number.Uint64() >= number && chain.GetHeaderByNumber(number-1).Hash() == (header.ParentHash) {
```
brilliant-lx, line 894: 1. it is too long
2. why add the hash check?

NathanBSC (approved):

NathanBSC (dismissed):

buddh0 (approved):

PR #1383, opened 2023-03-22
miner: no need to broadcast sidechain header mined by this validator

qinglin89:
### Description

if the mined block inserted as `sidechain` by the commiter, then shouldn't broadcast it.
details ref: #1382 

test with stress test (tps=800, duration=600s):
- before: reorg_execution=13
- with pr#1382+pr#1383: reorg_executio=0
reorgs

### Rationale

Reduce `offturn/sidechain` blocks traffic, reduce the probability of reorgs under big transactions traffic.
This could reduce the `offturn/sidechain` network traffic, and 
- since`parlia` is order based, it's unnecessary for validator to broadcast the block mined by itself and considered as `sidechain`.
- under big traffic or bad network condition, the `sidechain` blocks would cause more chaos(reorgs) since validators decided to reorg the head of canonical chain by 50% probability when the ttd were equal.

### Example

N/A

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -674,7 +674,7 @@ func (w *worker) resultLoop() {
 				if err != nil {
 					log.Error("Failed writing block to chain", "err", err, "status", status)
 				} else {
-					log.Info("Writen block as SideChain and avoid broadcasting", "status", status)
+					log.Info("Written block as SideChain and avoid broadcasting", "status", status)
 				}
 				continue
 			}
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -671,7 +671,11 @@ func (w *worker) resultLoop() {
 			start := time.Now()
 			status, err := w.chain.WriteBlockAndSetHead(block, receipts, logs, task.state, true)
 			if status != core.CanonStatTy {
-				log.Error("Failed writing block to chain", "err", err, "status", status)
+				if err != nil {
+					log.Error("Failed writing block to chain", "err", err, "status", status)
+				} else {
+					log.Info("Writen block as SideChain and avoid broadcasting", "status", status)
+				}
 				continue
 			}
 			writeBlockTimer.UpdateSince(start)
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -666,20 +666,19 @@ func (w *worker) resultLoop() {
 				w.recentMinedBlocks.Add(block.NumberU64(), []common.Hash{block.ParentHash()})
 			}
 
-			// Broadcast the block and announce chain insertion event
-			w.mux.Post(core.NewMinedBlockEvent{Block: block})
-
 			// Commit block and state to database.
 			task.state.SetExpectedStateRoot(block.Root())
 			start := time.Now()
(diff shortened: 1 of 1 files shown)


miner/worker.go
```diff
@@ -666,20 +666,19 @@ func (w *worker) resultLoop() {
...
-			w.mux.Post(core.NewMinedBlockEvent{Block: block})
-
 			// Commit block and state to database.
 			task.state.SetExpectedStateRoot(block.Root())
 			start := time.Now()
-			_, err := w.chain.WriteBlockAndSetHead(block, receipts, logs, task.state, true)
-			if err != nil {
-				log.Error("Failed writing block to chain", "err", err)
+			status, err := w.chain.WriteBlockAndSetHead(block, receipts, logs, task.state, true)
+			if status != core.CanonStatTy {
```
buddh0, line 673: when status==SideStatTy and err==nil
Log.Error will output,  but no error happens indeed

buddh0:
local canonial chain
block   n   difficulty 1
block   n+1   difficulty 2

now, local mined block n with difficulty 2, 
even if it's not in the canonial chain, should we propogate it？

qinglin89:
@buddh0 It's not a `should/shouldn't` case, this change would reduce the `unexpected traffic`, led to less `reorg execution`. This case you mentioned, follow the miner/worker logic, the expected behaviour should be `it be interrupted when the n(diff=1)`, rather to say it had imported `n+1(diff2)`

miner/worker.go
brilliant-lx: case 1: NonStatTy & err !=nil, ==>  likely the local DB is in trouble, better not broadcast
case 2: CanonStatTy    ==>   reorg done,  should not broadcast?
case 3: SideStatTy   ==> valid block is mined, should broad?

brilliant-lx: there are many cases which could lead to case 1:
- chain stopped
- local DB error: fail to get TD, fail to commit, invalid merkle root
- reorg failed,

brilliant-lx: I think it is ok, but could you add some comments?

brilliant-lx:
> local canonical chain block n difficulty 1 block n+1 difficulty 2
> 
> now, local mined block n with difficulty 2, even if it's not in the canonical chain, should we propagate it？
> 
> so the condition of propagating a new mined block is: it's in the canonical chain or it's a inturn block

In this case, I think the local mined in-turn block n should not be propagated, since it already exceeded its mining window. 
And it is unlikely to happen.

The more likely case could be:
- local canonical chain: block n difficulty 1
- now, local mined block n with difficulty 2,
in this case, the mined block will be propagated.

miner/worker.go
NathanBSC: my concern here is not output Error Log when not necessary
if err！=nil then  out put error log
if status == core.CanonStatTy then broadcast block 

but now when status!=core.CanonStatTy and err==nil, then output Error Log

qinglin89: updated

brilliant-lx (approved):

NathanBSC (dismissed):

buddh0 (approved):

PR #1479, opened 2023-04-18
fix: solve potential deadlock for bsc p2p protocol handshake

NathanBSC:
### Description

fix potential deadlock by bsc p2p protocol handshake

### Rationale

**Process description:**
1. runEthPeer wait bsc extension in waitBscExtension
    timeout is 10 seconds
2. bscHandler.RunPeer, then do handshake protocol, has a smaller timeout
3. timeout (10 seconds) is reached,
     but ps.lock is got by other go routines, such as bscHandler.RunPeer and registerBscExtension
     they are blocked in such code now
            `wait <- peer`
    then deadlock happens
    
**Root cause :** 
func RunPeer  is designed as not cost much time before.
but when handshake process added in, handshake also has a timeout, 
so ethPeer have to wait more time than before, more close to  timeout (10 seconds),
thus leading to improve chance of potential deadlock greatly.

**Solve out:**
we can move handshake of sub protocol such as bsc and diff
to behind handshake of main protocol

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/peerset.go
+++ b/eth/peerset.go
@@ -458,7 +458,12 @@ func (ps *peerSet) registerPeer(peer *eth.Peer, ext *snap.Peer, diffExt *diff.Pe
 		eth.trustExt = &trustPeer{trustExt}
 	}
 	if bscExt != nil {
-		eth.bscExt = &bscPeer{bscExt}
+		if err := bscExt.Handshake(); err == nil {
+			eth.bscExt = &bscPeer{bscExt}
+		} else {
+			peer.Log().Debug("bsc protocol handshake failed", "err", err)
+			return err
+		}
 	}
 	ps.peers[id] = eth
 	return nil
--- a/eth/protocols/bsc/handler.go
+++ b/eth/protocols/bsc/handler.go
@@ -1,6 +1,7 @@
 package bsc
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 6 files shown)


eth/peerset.go
```diff
@@ -458,7 +458,12 @@ func (ps *peerSet) registerPeer(peer *eth.Peer, ext *snap.Peer, diffExt *diff.Pe
 		eth.trustExt = &trustPeer{trustExt}
 	}
 	if bscExt != nil {
-		eth.bscExt = &bscPeer{bscExt}
+		if err := bscExt.Handshake(); err == nil {
```
guagualvcha, line 461: Does this seem against the design of `geth`, as if every protocol follow the same pattern, the handshake of the different protocol will happen serially.

eth/protocols/bsc/handler.go
```diff
@@ -74,6 +86,11 @@ func MakeProtocols(backend Backend, dnsdisc enode.Iterator) []p2p.Protocol {
 // Handle is the callback invoked to manage the life cycle of a `bsc` peer.
 // When this function terminates, the peer is disconnected.
 func Handle(backend Backend, peer *Peer) error {
+	select {
+	case <-peer.handshaked:
```
guagualvcha, line 90: This seems not extensible, as the channel is only used by `bsc` protocol, it wont work if another extension is needed in the future.

brilliant-lx:
Try to illustrate how the deadlock happen: 
- routine of the BSC protocol
image

- routine of the Eth protocol
image

eth/peerset.go
NathanBSC: the design of geth has no handshake exclude eth protocol
waitBscExtension and registerBscExtension can end quickly
when handshake of sub protocol involved in above two , it' begin to slow down and lead to problem

eth/protocols/bsc/handler.go
NathanBSC: snap has no handshake, it's easy.
all problem begin from a sub protocol need handshake

NathanBSC:
I will shutdown this PR
another PR with light change will solve this problem.
https://github.com/bnb-chain/bsc/pull/1484

eth/protocols/bsc/handler.go
NathanBSC: > This seems not extensible, as the channel is only used by `bsc` protocol, it wont work if another extension is needed in the future.


another sub protocol need handshake should have it's own `handshaked ` channel

PR #1481, opened 2023-04-18
core: delay blsSignatureVerify precompile to Lynn from Boneh

NathanBSC:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -141,9 +141,9 @@ var PrecompiledContractsBerlin = map[common.Address]PrecompiledContract{
 	common.BytesToAddress([]byte{9}): &blake2F{},
 }
 
-// PrecompiledContractsBoneh contains the default set of pre-compiled Ethereum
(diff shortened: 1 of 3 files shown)


core/vm/contracts.go
```diff
@@ -141,9 +141,9 @@ var PrecompiledContractsBerlin = map[common.Address]PrecompiledContract{
 	common.BytesToAddress([]byte{9}): &blake2F{},
 }
 
-// PrecompiledContractsBoneh contains the default set of pre-compiled Ethereum
+// PrecompiledContractsLynn contains the default set of pre-compiled Ethereum
 // contracts used in the Boneh release.
-var PrecompiledContractsBoneh = map[common.Address]PrecompiledContract{
```
guagualvcha, line 146: @KeefeL do have a precompile contract that want to enable at Boneh, so such modification may conflict with his intension.

Issue #1483, opened 2023-04-19
deadlock: due to P2P handshake of diffsync


brilliant-lx:
#### System information

Geth version: geth v1.1.21
OS & Version: Linux
Commit hash : v1.1.21

#### Expected behaviour
The node can keep running.

#### Actual behaviour
The node was stuck.
I dumped the routines and found there are tens of thousand routines and a deadlock bug.

#### Steps to reproduce the behaviour
Not easy to reproduce.

#### Backtrace
Try to illustrate how the deadlock happen: 
- routine of the BSC protocol
![image](https://user-images.githubusercontent.com/92799281/232947409-78b8108f-bef5-452e-b3c0-5694f4ac3217.png)

- routine of the Eth protocol
![image](https://user-images.githubusercontent.com/92799281/232947758-14fe87cd-2a67-4b9a-b865-08cf164b43f6.png)

PR #1484, opened 2023-04-19
fix: a deadlock caused by bsc protocol handeshake timeout

brilliant-lx:
### Description
The bug has been described in: https://github.com/bnb-chain/bsc/issues/1483

### Rationale
The current code of P2P module has the concurrent bug, it will lock a channel operation, which is not suggested.
To fully address the issue, we need to refactor the code, prefer not to carry it out now.
Just use TryLock & Wait to fix the deadlock.

### Example
NA

### Changes
NA

### DIFF
--- a/eth/handler_bsc.go
+++ b/eth/handler_bsc.go
@@ -27,7 +27,7 @@ func (h *bscHandler) RunPeer(peer *bsc.Peer, hand bsc.Handler) error {
 		ps.lock.Lock()
 		if wait, ok := ps.bscWait[id]; ok {
 			delete(ps.bscWait, id)
-			wait <- peer
+			wait <- nil
 		}
 		ps.lock.Unlock()
 		return err
--- a/eth/peerset.go
+++ b/eth/peerset.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/eth/protocols/eth"
 	"github.com/ethereum/go-ethereum/eth/protocols/snap"
 	"github.com/ethereum/go-ethereum/eth/protocols/trust"
+	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p"
 )
 
@@ -70,6 +71,7 @@ const (
 	// extensionWaitTimeout is the maximum allowed time for the extension wait to
 	// complete before dropping the connection as malicious.
 	extensionWaitTimeout = 10 * time.Second
+	tryWaitTimeout       = 200 * time.Millisecond
 )
 
 // peerSet represents the collection of active peers currently participating in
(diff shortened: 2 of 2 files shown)


eth/peerset.go
```diff
@@ -402,7 +404,15 @@ func (ps *peerSet) waitBscExtension(peer *eth.Peer) (*bsc.Peer, error) {
 		return peer, nil
 
 	case <-time.After(extensionWaitTimeout):
-		ps.lock.Lock()
+		// once timeout is reached, we won't wait for the peer any more, bsc protocol won't be enabled on it.
+		log.Warn("waitBscExtension", "peer", id, "bsc protocol timeout, it won't be enabled.")
+		for {
+			// it is potential deadlock, so we use TryLock to avoid it.
+			if ps.lock.TryLock() {
```
NathanBSC, line 411: still deadlock:
    wait <- nil, blocked because no one to accept nil from channel
    ps.lock.TryLock() always return false at the same time

brilliant-lx: yes, forget to clean the channel... will update

eth/handler_bsc.go
```diff
@@ -27,7 +27,7 @@ func (h *bscHandler) RunPeer(peer *bsc.Peer, hand bsc.Handler) error {
 		ps.lock.Lock()
 		if wait, ok := ps.bscWait[id]; ok {
 			delete(ps.bscWait, id)
-			wait <- peer
+			wait <- nil
```
NathanBSC, line 30: this go routine still have chance to be blocked？

eth/peerset.go
```diff
@@ -402,10 +404,27 @@ func (ps *peerSet) waitBscExtension(peer *eth.Peer) (*bsc.Peer, error) {
 		return peer, nil
 
 	case <-time.After(extensionWaitTimeout):
-		ps.lock.Lock()
-		delete(ps.bscWait, id)
-		ps.lock.Unlock()
-		return nil, errPeerWaitTimeout
+		// once timeout is reached, bsc protocol won't be enabled on this peer.
+		log.Warn("waitBscExtension", "peer", id, "bsc protocol timeout, it won't be enabled.")
+		// could be deadlock, so we use TryLock to avoid it.
```
NathanBSC, line 409: can delete 409～414？

brilliant-lx: yes, can be deleted, but it will cause an extra tryWaitTimeout, so not prefer to delete it.

eth/peerset.go
```diff
@@ -402,10 +404,27 @@ func (ps *peerSet) waitBscExtension(peer *eth.Peer) (*bsc.Peer, error) {
 		return peer, nil
```
NathanBSC, line 404: output a warn when peer is nil

NathanBSC, line 404: output a warn when peer is nil

brilliant-lx: ok, it will be printed as Warning log level in `startProtocols ()`

NathanBSC (approved):

blxdyx (approved):

zzphust (approved):

guagualvcha (approved):

PR #1510, opened 2023-04-23
fix: all votes rejected for span voting caused by reorg

NathanBSC:
### Description

image
after luban upgrade, voting is enabled, but not use justified number in fork choice;
so it can generate reorg easily as before by shutdown one validator
as in picture:
validator vote for 49578-->49579
then reorg and sourceNumber reset to 49577
all votes after reorg will rejected for span voting

### Rationale
1. there may be great distance between blockNumber and targetNumber
    so iterator from sourceNumber + 1 to targetNumber is dangerous
2. just ignore votes for target before targetNumber - maxSizeOfRecentEntry when checking rules
    
by the way:
this case is just a reason why we require sourceNumber of vote is in a range before block.number
image

set maliciousVoteSlashScope to 256 is safe:
if current block is n
then target block number is n-1
then iterate begin target number is n-1-256
then                    max(begin source number) is n-1-256-1 = n-258
in contracts _begin source number_ is n-255

n-288<n-255, so scan scope in bsc is larger, it's safe to set  maliciousVoteSlashScope to 256

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vote/vote_journal.go
+++ b/core/vote/vote_journal.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	maxSizeOfRecentEntry = 512
+	maxSizeOfRecentEntry    = 512
+	maliciousVoteSlashScope = 256
 )
 
 type VoteJournal struct {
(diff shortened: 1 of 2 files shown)


guagualvcha:
Please refine the title of this PR.

guagualvcha (approved):

brilliant-lx (dismissed):

brilliant-lx (approved):

guagualvcha (approved):

PR #1547, opened 2023-04-28
fix: recently signed check when slashing unavailable validator

minh-bq:
### Description

The first validator in snap.Recents is shifted out and allowed to seal the block already. However, when determining if we should slash the validator in Finalize and FinalizeAndAssemble, we mark that first validator as unable to seal block and don't slash this validator. This commit fixes that bug and creates a separate helper function to check if the validator recently signed a block for consistency in all places the check happens.

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -754,13 +754,8 @@ func (p *Parlia) verifySeal(chain consensus.ChainHeaderReader, header *types.Hea
 		return errUnauthorizedValidator
 	}
 
-	for seen, recent := range snap.Recents {
-		if recent == signer {
-			// Signer is among recents, only fail if the current block doesn't shift it out
-			if limit := uint64(len(snap.Validators)/2 + 1); seen > number-limit {
-				return errRecentlySigned
-			}
-		}
+	if snap.SignRecently(signer) {
+		return errRecentlySigned
 	}
 
 	// Ensure that the difficulty corresponds to the turn-ness of the signer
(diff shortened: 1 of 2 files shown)


consensus/parlia/parlia.go
```diff
@@ -1294,14 +1298,8 @@ func (p *Parlia) Seal(chain consensus.ChainHeaderReader, block *types.Block, res
...
-	for seen, recent := range snap.Recents {
-		if recent == val {
-			// Signer is among recents, only wait if the current block doesn't shift it out
-			if limit := uint64(len(snap.Validators)/2 + 1); number < limit || seen > number-limit {
-				log.Info("Signed recently, must wait for others")
-				return nil
-			}
-		}
+	if snap.SignRecently(val) {
+		return nil
```
buddh0, line 1302: keep log here

consensus/parlia/snapshot.go
```diff
@@ -207,6 +207,17 @@ func (s *Snapshot) updateAttestation(header *types.Header, chainConfig *params.C
 	}
 }
 
+func (s *Snapshot) SignRecently(validator common.Address) bool {
+	for seen, recent := range s.Recents {
+		if recent == validator {
+			if limit := uint64(len(s.Validators)/2 + 1); s.Number+1 < limit || seen > s.Number+1-limit {
```
BetZzzzz, line 213: `seen>s.Number+1-limit` includes  the `s.Number+1<limit`, 
remove `s.Number+1<limit` would be fine.

NathanBSC: I think ‘s.Number+1<limit’ should be kept.
becase  `s.Number+1-limit` would overflow  when s.Number+1 < limit

NathanBSC: > `seen>s.Number+1-limit` includes the `s.Number+1<limit`, remove `s.Number+1<limit` would be fine.

I think ‘s.Number+1<limit’ should be kept.
becase s.Number+1-limit would overflow when s.Number+1 < limit

BetZzzzz: that's right.

Kingo033:
Screenshot_۲۰۲۳۰۵۰۳-۰۶۱۳۳۷_Chrome

consensus/parlia/parlia.go
minh-bq: I've added log

buddh0 (approved):

BetZzzzz (approved):

Issue #1605, opened 2023-05-12
getBlockByNumber support `safe` and `finalized` in cosole


NathanBSC:
#### System information

curl --location --request POST 'https://data-seed-prebsc-1-s1.binance.org:8545/' --header 'Content-Type: application/json' --data '{"jsonrpc": "2.0", "method": "eth_getBlockByNumber", "params":["safe", true], "id": 1}'
works right 

but when ./geth attach geth.ipc
 
![image](https://github.com/bnb-chain/bsc/assets/122502194/5360bec0-fe43-4857-838e-60dfed834084)

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

PR #1617, opened 2023-05-15
jsre: add new block types

MatusKysel:
### Description

As we introduced new RPC block types recently with FF https://github.com/bnb-chain/bsc/blob/develop/rpc/types.go#L63
```
	SafeBlockNumber      = BlockNumber(-4)
	FinalizedBlockNumber = BlockNumber(-3)
```
we need to add support to JS console too

### Rationale

Resolves https://github.com/bnb-chain/bsc/issues/1605

### Example

### Changes

Notable changes: 
* ...


emailtovamos:
@MatusKysel I guess this change already happened through some other commit in `develop`?

Issue #1419, opened 2023-04-07
Too many Erigon P2P Connections, it made the P2P service hard to maintain


brilliant-lx:
#### System information
Geth version: `geth v1.1.21`.  // It is irrelevant to Geth version
OS & Version: Linux

#### Expected behavior
The P2P connection should be at a reasonable level, better not reach the up limit, defined by `MaxPeers`

#### Actual behavior
MaxPeers has been set to 3500, but the P2P connection reached the up limit very soon.
I dump the peers list, found 2 weird things:
1.Erigon client takes ~2/3 of the peers;
```
"name": "erigon/v2.32.0/linux-amd64/go1.18.7",
...
```
2.Erigon client would keep tens of p2p connections with same IP address.
Attach the log of `admin.peers`
[admin_peers_1.log](https://github.com/bnb-chain/bsc/files/11176434/admin_peers_1.log)

Here is some screenshot from the log:
![image](https://user-images.githubusercontent.com/92799281/230551650-58271565-cb1e-44d9-94b1-3ee6ffe9309f.png)
![image](https://user-images.githubusercontent.com/92799281/230551672-17fb1c56-8dda-4cf8-8fa3-4504863c2162.png)
![image](https://user-images.githubusercontent.com/92799281/230551692-3a4d9bd5-36ea-4b1f-8077-3513aa853cd3.png)

#### Steps to reproduce the behavior
You should setup an public bsc node, let many other peers connect to you. And you can observe the connected peer list.
But may not easy to reproduce in you device, since we have famous P2P nodes, which is much likely to be connected.

#### Backtrace
NA

When submitting logs: please submit them as text and not screenshots.

wellgamer789:
They are almost all fake nodes, they don't send anything, they just consume and receive

It's been more than 4 months that I've noticed that their number is very high, in any region of the planet they will only connect to you if you are using the official build

fbit:
Just an idea:

Wouldn't it be possible to create a patch/fix so that only one or two peers with the same remoteAddress (IP) are accepted at the same time? This way those evil nodes wouldn't be able to spam and block other nodes.... well unless they deploy on hundreds of servers with different IP addresses.

brilliant-lx:
> 

we have applied it in some of the P2P nodes, may consider to raise a formal patch.

kaber2:
[...]

PR #1623, opened 2023-05-19
[WIP] P2P: try to limit the connection number per IP address

brilliant-lx:
### Description
background: https://github.com/bnb-chain/bsc/issues/1419

### Rationale
We have observed tens of connections from a single IP, there could have 2 cases:
- case 1: they all comes a single device
- case 2: they are on different devices, with NAT(Network Address Translation), they all share a same public IP address.

For case 1, it seems unnecessary to establish more than 1 connections on a single device, so we can limit their connection number to avoid P2P attack.
For case 2, since they are on different devices, their connection request should be accepted. But the problem is that there are too many of them, which makes the network congested. It makes new peers difficult to find a good peer. I think, we still can apply this IP limit on some of the public nodes, so these pubic nodes would have capacity to serve these newly joined peers.

### Example
NA

### Changes
Could have impact to P2P connection, nodes running in cloud with single IP could be rejected.


eth/handler.go
```diff
@@ -140,6 +141,8 @@ type handler struct {
 	maliciousVoteMonitor *monitor.MaliciousVoteMonitor
 	chain                *core.BlockChain
 	maxPeers             int
+	maxPeersPerIp        int
+	peersPerIp           map[string]int
```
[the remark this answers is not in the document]
brilliant-lx: I think I can simply add Lock to protect, rather than sync.Map, ok for you?

guagualvcha:
Did not get time to fully walk through the code, but the idea LGTM, and the implementation is also LGTM.

cmd/geth/main.go
```diff
@@ -132,6 +132,7 @@ var (
 		utils.PruneAncientDataFlag,
 		utils.ListenPortFlag,
 		utils.MaxPeersFlag,
+		utils.MaxPeersPerIpFlag,
```
sysvm, line 135: IP is a proper noun, so this var can be renamed to `MaxPeersPerIPFlag`. Reference: [Initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

brilliant-lx: accept, thank you for your input.

sysvm: The IP used below should also be capitalized.

brilliant-lx: yes, updated, thx

GalaIO (approved):

blxdyx (approved):

zzphust (approved):

PR #1624, opened 2023-05-19
core,eth: add rpc method to trie flush frequency

jsvisa:
### Description

backport from https://github.com/ethereum/go-ethereum/pull/24785


zzzckck:
already supported after v1.3.0

PR #1648, opened 2023-05-24
miner: adding MEV to validator

Uri-Margalit-bloXroute:
### Description

This PR allows a validator node to utilize bloXroute MEV services. The validator node
* registers itself to MEV relay upon restart or an epoch change
* processes eth_proposedBlock from MEV relays

### Rationale

The introduction of mev-boost in Ethereum enabled a win-win situation for Searchers, Block builders and Proposers which allows all three to profit off of potential arbitrage opportunities, and do so not at the expense of other traders and dApps executing transactions on chain.

This PR implements the validator to MEV relay integration

### Example

eth_proposedBlock endpoint receives

```
{
   "id": "1",
   "method": "eth_proposedBlock",
   "params": [{
      "blockNumber": "0xba10d0",
      "prevBlockHash": "0xabc123...555",
      "blockReward": 1230000000000000,
      "gasLimit": 140000000,
      "gasUsed": 70000000,
      "payload": ["ab..ab", "cd..cd"],
   }]
}
```

### Changes

Notable changes: 
* Added couple of startup argument
  * miner.mevrelays - list of uris of MEV relays the validator works with
  * miner.mevproposedblockuri - the callback uri which the MEV relay should send the proposed blocks
* Added web3 endpoint eth_proposedBlock - the MEV relays use this endpoint to send their proposed blocks
* Added a logic in the miner to validate the proposed blocks from the MEV relays and consider them when the validator choose which block it should propose


Uri-Margalit-bloXroute:
@brilliant-lx , is there an option to rerun the Unit test?

kaber2:
According to https://bloxroute.com/pricing/, this requires a subscription costing at least $1250/month. My personal opinion is, its a bad idea to integrate proprietary services into the reference code base for BSC. There's a maintenance cost and a risk of bugs while having zero benefit unless you're willing to shell out quite a lot of money to Bloxroute.

Validators can easily add this patch themselves.

github1youlc:
This will provide convenience for sandwich attack, front-running attack, is it a good thing?
Can your solution consider the suffering of the normal traders on BSC.

kaber2:
> The introduction of mev-boost in Ethereum enabled a win-win situation for Searchers, Block builders and Proposers which allows all three to profit off of potential arbitrage opportunities, and do so not at the expense of other traders and dApps executing transactions on chain.

Just to be very clear about this. There is no win-win and this is a false narrative. For searchers, sure, it enables the ones too lazy to do a proper implementation to cheat their way to a successful trade, for now. The long term economic incentives are purely in favor of miners/validators. Any kind of easily accessible block manipulation for a bribe results in a race to the bottom (top), who offers the most to the block producer. The end result can only be searchers offering anything they can without going bankrupt. This in effect has a great effect on many protocols, which rely on external transactors to keep prices in balance, clear out bad debt, ... The end result will be that most searchers will go out of business and these vital transactions will be highly centralized.

There is no win-win here, just win, and in my opinion, you trying to externalize your costs while pocketing the gains should just shove off.

brilliant-lx:
> @brilliant-lx , is there an option to rerun the Unit test?

@Uri-Margalit-bloXroute sorry, missed the message, UT can be rerun just click into the failed page, there will be a rerun icon. And could you rebase to the latest develop branch?

cmd/utils/flags.go
```diff
@@ -1571,6 +1593,60 @@ func setWhitelist(ctx *cli.Context, cfg *ethconfig.Config) {
 	}
 }
 
+func setMEV(ctx *cli.Context, ks *keystore.KeyStore, cfg *miner.Config) {
```
guagualvcha, line 1596: As some validator are running in a private network and being protected by a sentry node, is there any solution for them?

internal/ethapi/api.go
```diff
@@ -1332,6 +1332,57 @@ func (s *PublicBlockChainAPI) GetVerifyResult(ctx context.Context, blockNr rpc.B
...
+	BlockNumber   rpc.BlockNumber `json:"blockNumber"`
+	PrevBlockHash common.Hash     `json:"prevBlockHash"`
+	BlockReward   *big.Int        `json:"blockReward"`
+	GasLimit      uint64          `json:"gasLimit"`
+	GasUsed       uint64          `json:"gasUsed"`
+	Payload       []hexutil.Bytes `json:"payload"`
+}
+
+// ProposedBlock will submit the block to the miner worker
+func (s *PublicBlockChainAPI) ProposedBlock(ctx context.Context, args ProposedBlockArgs) error {
```
guagualvcha, line 1355: Any authorization method to verify the request comes from the relayer?

miner/worker.go
```diff
@@ -1057,10 +1129,147 @@ func (w *worker) generateWork(params *generateParams) (*types.Block, error) {
...
+	processorCapacity := 100
+	bloomProcessors := core.NewAsyncReceiptBloomGenerator(processorCapacity)
+
+	signal := commitInterruptNone
+	for _, tx := range block.txs {
+		// If we don't have enough gas for any further transactions then we're done
+		if env.gasPool.Gas() < params.TxGas {
+			log.Trace("Not enough gas for further transactions", "have", env.gasPool, "want", params.TxGas)
+			signal = commitInterruptOutOfGas
+			break
```
github1youlc, line 1173: If some transactions are lost in case of out of gas, it will be a disaster when existing sandwich attack.

Jolly23:
Good, let’s make BSC even worse

Jolly23:
It is fucking ridiculous to put this shit inside BSC official repo. To force searchers to buy your $1250 stupid monthly subscription shit. 
Couldn’t you find a restroom?

lss15:
I don't think it's a good idea because it will benefit the subscription sellers and those who purchase the $1250.00 subscription

48ClubIan:
Why on earth do you think it is proper to embed a specific mev service provider's code into official repo ?

Maintain your own repo.

giovannicarlos77:
I think it’s a bad ideia, benefiting just people that bought 1250$/monthly plan.

internal/ethapi/api.go
Jolly23: Good, pretend like a blockchain genius but never let it stop.

cmd/utils/flags.go
Uri-Margalit-bloXroute: Yes, there is another branch for the sentry node to be a proxy between the MEV relays and the validator

Jolly23:
commit shits non-stop

Uri-Margalit-bloXroute:
moving the PR to a different place

PR #1651, opened 2023-05-25
fix: remove naturally finality

NathanBSC:
### Description

remove naturally finality

### Rationale
naturally finality is alias of probabilistic finality in the code.
1. The probabilistic finality is only probabilistic. In extreme cases, it may be reverted without penalty, so it is not suitable to acknowledge it through the interface
2. Different application scenarios of probabilistic finality may choose different numbers of blocks. For example, 15 is recommended, while the main station uses 21 for deposits and withdrawals. but only a fixed number can be specified in the code
3. The Ethereum client will also degenerate to a probabilistic finality when the vote fails, but the number of blocks is not specified in the code; Bitcoin has probabilistic finality, and the number of blocks is not specified in the code

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/consensus/parlia/snapshot.go
+++ b/consensus/parlia/snapshot.go
@@ -46,7 +46,7 @@ type Snapshot struct {
 	Validators       map[common.Address]*ValidatorInfo `json:"validators"`            // Set of authorized validators at this moment
 	Recents          map[uint64]common.Address         `json:"recents"`               // Set of recent validators for spam protections
 	RecentForkHashes map[uint64]string                 `json:"recent_fork_hashes"`    // Set of recent forkHash
-	Attestation      *types.VoteData                   `json:"attestation:omitempty"` // Attestation for fast finality
+	Attestation      *types.VoteData                   `json:"attestation:omitempty"` // Attestation for fast finality, but `Source` used as `Finalized`
 }
 
 type ValidatorInfo struct {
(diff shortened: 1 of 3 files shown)


consensus/parlia/snapshot.go
```diff
@@ -199,11 +199,11 @@ func (s *Snapshot) updateAttestation(header *types.Header, chainConfig *params.C
 	}
 
 	// Update attestation
-	s.Attestation = &types.VoteData{
-		SourceNumber: attestation.Data.SourceNumber,
-		SourceHash:   attestation.Data.SourceHash,
-		TargetNumber: attestation.Data.TargetNumber,
-		TargetHash:   attestation.Data.TargetHash,
+	s.Attestation.TargetNumber = attestation.Data.TargetNumber
```
NathanBSC, line 202: It's a bug here
Attestation is nill, will crash

emailtovamos:
What exactly is the reason for removal of `NaturallyFinalizedDist` ? I didn't quite understand the rationale given.

NathanBSC:
> What exactly is the reason for removal of `NaturallyFinalizedDist` ? I didn't quite understand the rationale given.

the reason remove naturally finality in code:
1. The probabilistic finality is only probabilistic. In extreme cases, it may be reverted without penalty, so it is not suitable to acknowledge it through the interface
2. Different application scenarios of probabilistic finality may choose different numbers of blocks. For example, 15 is recommended, while the main station uses 21 for deposits and withdrawals. but only a fixed number can be specified in the code
3. The Ethereum client will also degenerate to a probabilistic finality when the vote fails, but the number of blocks is not specified in the code; Bitcoin is a probabilistic finality, and the number of blocks is not specified in the code

NathanBSC:
> What exactly is the reason for removal of `NaturallyFinalizedDist` ? I didn't quite understand the rationale given.

reason updated

consensus/parlia/snapshot.go
```diff
@@ -199,11 +199,11 @@ func (s *Snapshot) updateAttestation(header *types.Header, chainConfig *params.C
...
-	s.Attestation = &types.VoteData{
-		SourceNumber: attestation.Data.SourceNumber,
-		SourceHash:   attestation.Data.SourceHash,
-		TargetNumber: attestation.Data.TargetNumber,
-		TargetHash:   attestation.Data.TargetHash,
+	if s.Attestation != nil && attestation.Data.SourceNumber+1 != attestation.Data.TargetNumber {
+		s.Attestation.TargetNumber = attestation.Data.TargetNumber
+		s.Attestation.TargetHash = attestation.Data.TargetHash
+	} else {
+		s.Attestation = attestation.Data
```
brilliant-lx, line 206: there could be corner case? i.e. the first Attestation in the network, what's its SourceNumber?

brilliant-lx: ok, it will be zero, no problem

brilliant-lx (approved):

guagualvcha (approved):

PR #1671, opened 2023-06-05
fix: voting can only be enabled when mining

NathanBSC:
### Description

voting can only be enabled when mining

### Rationale

The node operator generally starts two nodes with the same consensus key. After having ff, the node will configure the same vote key.
After startup, use miner.stop() and miner.start() in geth console to switch out block nodes.
If the --vote option is used when both nodes are started, conflicting votes will be cast, which will be judged as malicious voting.

the solution is as following:
neither mining decide voting, nor they are independent
Instead, let mining be a prerequisite of voting

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vote/vote_manager.go
+++ b/core/vote/vote_manager.go
@@ -16,9 +16,15 @@ import (
 
 var votesManagerCounter = metrics.NewRegisteredCounter("votesManager/local", nil)
 
+// Backend wraps all methods required for voting.
+type Backend interface {
+	IsMining() bool
+	EventMux() *event.TypeMux
+}
+
 // VoteManager will handle the vote produced by self.
 type VoteManager struct {
-	mux *event.TypeMux
+	eth Backend
 
 	chain       *core.BlockChain
 	chainconfig *params.ChainConfig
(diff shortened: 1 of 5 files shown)


eth/backend.go
```diff
@@ -299,16 +274,39 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
...
 		return nil, err
 	}
-	if eth.votePool != nil {
-		eth.handler.votepool = eth.votePool
+
+	eth.miner = miner.New(eth, &config.Miner, chainConfig, eth.EventMux(), eth.engine, eth.isLocalBlock)
+	eth.miner.SetExtra(makeExtraData(config.Miner.ExtraData))
+
+	// Create voteManager instance
+	if posa, ok := eth.engine.(consensus.PoSA); ok {
```
brilliant-lx, line 282: why move these code after `eth.miner` create? any difference?

brilliant-lx, line 282: can we just add a check here, if mining is not enabled, no need to create: VotePool, VoteManager?

core/vote/vote_manager.go
```diff
@@ -16,9 +16,15 @@ import (
 
 var votesManagerCounter = metrics.NewRegisteredCounter("votesManager/local", nil)
 
+// Backend wraps all methods required for voting.
+type Backend interface {
+	IsMining() bool
+	EventMux() *event.TypeMux
+}
```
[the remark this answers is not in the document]
buddh0: result of IsMining() changes dynamically
image

eth/backend.go
buddh0: no 
validator can start without the `mine` flag
but start mining later through geth console

brilliant-lx (approved):

Pythonberg1997 (approved):

Issue #1674, opened 2023-06-07
It took quite a long time for node to shutdown


brilliant-lx:
#### System information

Geth version: `geth v1.2.4`
OS & Version: Linux
Commit hash : just the tag v1.2.4

#### Expected behaviour
When node shutdown, it can be shutdown gracefully within a few second.

#### Actual behaviour
It took more than several minutes for node to shutdown.

#### Steps to reproduce the behaviour
Start a BSC full node with v1.2.4, sync to the latest block, shut it down by: `kill <pid of bsc>`
The process can not be shutdown, it keeps printing the following logs.
```
t=2023-06-06T14:50:53+0000 lvl=info msg="Got interrupt, shutting down..."
t=2023-06-06T14:50:53+0000 lvl=warn msg="HTTP server graceful shutdown timed out"
t=2023-06-06T14:50:53+0000 lvl=info msg="HTTP server stopped"                 endpoint=...:8545
t=2023-06-06T14:50:53+0000 lvl=info msg="IPC endpoint closed"                 url=/.../geth.ipc
t=2023-06-06T14:50:53+0000 lvl=info msg="Les server stopped"
t=2023-06-06T14:50:53+0000 lvl=info msg="Ethereum protocol stopped"
t=2023-06-06T14:50:53+0000 lvl=info msg="Transaction pool stopped"
t=2023-06-06T14:50:53+0000 lvl=eror msg="Bsc extension Handshake failed"      peer=803058b1 err=EOF
t=2023-06-06T14:50:53+0000 lvl=eror msg="Bsc extension Handshake failed"      peer=a7ca772b err=EOF
t=2023-06-06T14:50:53+0000 lvl=eror msg="Bsc extension Handshake failed"      peer=3370c8d6 err=EOF
t=2023-06-06T14:50:54+0000 lvl=info msg="commitWork SubscribeNewTxsEvent return nil"
t=2023-06-06T14:50:55+0000 lvl=info msg="Writing cached state to disk"        block=28,867,106 hash=0x84d3a9ee5dcfcc08508835f26fdb15c6957e7459b26404130d4a74facd7c14bb root=0xaccf7080ec70b8f33944cb1a0d29c75f4a3e20b5e93e0d6b8d2a52b7a3c107b9
[...]

PR #1675, opened 2023-06-07
goleveldb: downgrade the version for performance

brilliant-lx:
### Description
this PR: https://github.com/bnb-chain/bsc/pull/936 upgraded the goleveldb indirectly, since it has dependency of prysm, which also depends on goleveldb indrectly.

But the new version has some performance issue on Database Close, see: https://github.com/syndtr/goleveldb/issues/434

So we need to rollback to the previous verison
The issue has been described in: https://github.com/bnb-chain/bsc/issues/1674

### Rationale
NA

### Example
NA

### Changes
No impact to users.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -62,7 +62,7 @@ require (
 	github.com/shirou/gopsutil v3.21.11+incompatible
 	github.com/status-im/keycard-go v0.2.0
 	github.com/stretchr/testify v1.8.1
-	github.com/syndtr/goleveldb v1.0.1-0.20220721030215-126854af5e6d
+	github.com/syndtr/goleveldb v1.0.1
 	github.com/tendermint/go-amino v0.14.1
 	github.com/tendermint/iavl v0.12.0
 	github.com/tendermint/tendermint v0.31.15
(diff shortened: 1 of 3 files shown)


weiihann (dismissed):

zzphust (dismissed):

buddh0 (dismissed):

go.mod
```diff
@@ -62,7 +62,7 @@ require (
 	github.com/shirou/gopsutil v3.21.11+incompatible
 	github.com/status-im/keycard-go v0.2.0
 	github.com/stretchr/testify v1.8.1
-	github.com/syndtr/goleveldb v1.0.1-0.20220721030215-126854af5e6d
+	github.com/syndtr/goleveldb v1.0.1
```
buddh0, line 65: just change to following?
github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7

so not need to replace later
and no need to modify accounts/abi/bind/bind_test.go

I think if path  changed, like github.com/**tendermint**/tendermint => github.com/**bnb-chain**/tendermint
we need replace it later
for goleveldb, there is no need.

buddh0: can't do as above
because pyrsm depend on v1.0.1-0.20220721030215-126854af5e6d

buddh0 (approved):

weiihann (approved):

zzphust (approved):

PR #1676, opened 2023-06-07
EIP-3529 (BEP-212) Unit tests for Parlia Config

Mister-EA:
### Description
In this PR a test config `TestChainParliaConfig` that uses the Parlia engine is introduced for unit testing purposes.

This config is used to add unit tests for [EIP-3529/BEP-212](https://github.com/bnb-chain/BEPs/pull/212) . 

There are already tests for EIP-3529 targeting the  default `TestChainConfig` which uses a fake ethash engine. 

In this PR, additional test cases are included targeting `TestChainConfig`, and the same test cases are replicated for `TestChainParliaConfig`.

### Rationale
Since in the Hertz hard fork Berlin and London EIPs will be applied simultaneously, the expected gas usage values between the ethash unit tests and  Parlia config unit tests for EIP-3529 are different in some cases.

This is because of different VM instructionsets before the EIP-3529 is applied:

| Config  | Instructionset before EIP-3529   | Instructionset after EIP-3529|
|---|---|---|
| `TestChainConfig`  | Berlin   | London  |
|`TestChainParliaConfig`| Istanbul | London |

### Changes

Notable changes: 
* `params/config.go` : `TestChainParliaConfig` is defined.
* `core/blockchain_eip3529_tests.go` : Additional EIP-3529 test cases for `TestChainConfig`.
* `core/parlia_tests/blockchain_eip3529_tests.go` : EIP-3529 test cases for `TestChainParliaConfig`.

### DIFF
--- a/core/parlia_tests/blockchain_eip_3529_test.go
+++ b/core/parlia_tests/blockchain_eip_3529_test.go
@@ -144,7 +144,7 @@ func TestSstoreClearGasPostEIP3529(t *testing.T) {
 	initialStorage[common.HexToHash("01")] = common.HexToHash("01")
 	initialStorage[common.HexToHash("02")] = common.HexToHash("02")
 
-	// Expected gas is intrinsic +  2*pushGas + cold load (due to legacy tx) + SstoreReset (a->b such that a!=0) - sstoreClearGasRefund
+	// Expected gas is intrinsic +  2*pushGas + SstoreReset (a->b such that a!=0) - sstoreClearGasRefund
 	expectedGasUsage := params.TxGas + 2*vm.GasFastestStep + params.SstoreResetGasEIP2200 - params.SstoreClearsScheduleRefundEIP3529
 	testGasUsage(t, postHertzConfig(), ethash.NewFaker(), bytecode, initialStorage, 60_000, expectedGasUsage)
 }
--- a/core/parlia_tests/blockchain_eip_3529_test.go
+++ b/core/parlia_tests/blockchain_eip_3529_test.go
@@ -1,4 +1,4 @@
-package core
+package parlia_tests
 
 import (
 	"math/big"
@@ -7,6 +7,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/consensus/ethash"
+	"github.com/ethereum/go-ethereum/core"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/core/vm"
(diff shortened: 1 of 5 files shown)


core/parlia_tests/blockchain_eip_3529_test.go
```diff
@@ -0,0 +1,210 @@
...
+		byte(vm.PUSH1), 0x1, // location
+		byte(vm.SSTORE), // Set slot[1] = 0
+	}
+	// initialize contract storage
+	initialStorage := make(map[common.Hash]common.Hash)
+	// Populate two slots
+	initialStorage[common.HexToHash("01")] = common.HexToHash("01")
+	initialStorage[common.HexToHash("02")] = common.HexToHash("02")
+
+	// Expected gas is intrinsic +  2*pushGas + cold load (due to legacy tx) + SstoreReset (a->b such that a!=0) - sstoreClearGasRefund
```
emailtovamos, line 147: This is mentioning 5 components but the formula in the next line has 4 components. Which two components are combined?

Mister-EA: Good that you noticed! It looks like I forgot to remove the mention of the cold load in the comment.

sunny2022da (approved):
LGTM.
Just minor comments (may comes from my personal inner struggling) that it seems most of the code in parlia_tests are same as the one outside (which targeting to London),with just minor changes for choosing parlia config, maybe in future we can figure out a way for reusing those codes, which is expected to simplify the code maintenance work. 
On the other side, considering that parlia is vital for bsc so it may worthwhile to have a separate folder for emphasis and clarify.   I think it is also make sense to dup the test code for parlia separately. 

I approve this change as the maintenance thing mentioned has no reason to block the merge of this PR IMHO. 
Thanks!

core/parlia_tests/blockchain_eip_3529_test.go
```diff
@@ -0,0 +1,210 @@
+package parlia_tests
```
sunny2022da, line 1: Do we need add header here for Copyright info?

Mister-EA: I haven't seen any copyright headers in BSC specific files like `parlia.go`.

Mister-EA:
@sunny2022da Good thought, but unfortunately Go does not allow you to import functions from test files. Only way would be to declare a `TestGasUsage` function in a non-test package, and import it from the two test files we have. I am not a fan of this solution, since I don't think it's a good idea to have testing code in the compiled binaries.

So unless you have any other suggestion, I think we have to accept this code duplication in the tests.

sunny2022da:
> @sunny2022da Good thought, but unfortunately Go does not allow you to import functions from test files. Only way would be to declare a `TestGasUsage` function in a non-test package, and import it from the two test files we have. I am not a fan of this solution, since I don't think it's a good idea to have testing code in the compiled binaries.
> 
> So unless you have any other suggestion, I think we have to accept this code duplication in the tests.

Agree, I have also tried the way you mentioned and found it is hard to do that. So let's leave it as it is at present.

core/blockchain_eip3529_test.go
```diff
@@ -39,13 +39,49 @@ func TestSelfDestructGasPreEIP3529(t *testing.T) {
...
+		byte(vm.SSTORE), // Set slot[1] = 3
+	}
+	// initialize contract storage
+	initialStorage := make(map[common.Hash]common.Hash)
+	// Populate two slots
+	initialStorage[common.HexToHash("01")] = common.HexToHash("01")
+	initialStorage[common.HexToHash("02")] = common.HexToHash("02")
+	// Expected gas is intrinsic +  2*pushGas + cold load (due to legacy tx) + SstoreReset (a->b such that a!=0)
+	// i.e. no refund
+	expectedGasUsed := params.TxGas + 2*vm.GasFastestStep + params.ColdSloadCostEIP2929 + (params.SstoreResetGasEIP2200 - params.ColdSloadCostEIP2929)
```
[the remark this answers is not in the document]
Mister-EA: This is a unit test for ethash test chain before London, which means the instructionset for this test is the Berlin instructionset.

brilliant-lx (approved):

sunny2022da (approved):
LGTM

PR #1682, opened 2023-06-08
fix: set the signer of parlia to the most permissive one

NathanBSC:
### Description

update signer of parlia after london upgrade

### Rationale

chain will be blocked when get a EIP1559 tx.

EIP1559 tx can be handled when generating a new block, 
because  func MakeSigner will return the newest signer
```
func (w *worker) makeEnv(parent *types.Block, header *types.Header, coinbase common.Address...
	// Note the passed coinbase may be different with header.Coinbase.
	env := &environment{
		signer:    types.MakeSigner(w.chainConfig, header.Number),
		state:     state,
		coinbase:  coinbase,
                ...
	}
```

but when a validator replay txs of blocks form others, it use the signer of parlia 
which is a EIP155Signer
```
func (p *Parlia) IsSystemTransaction(tx *types.Transaction, header *types.Header) (bool, error) {
	// deploy a contract
	if tx.To() == nil {
		return false, nil
	}
	sender, err := types.Sender(p.signer, tx)
	if err != nil {
		return false, errors.New("UnAuthorized transaction")
	}
	if sender == header.Coinbase && isToSystemContract(*tx.To()) && tx.GasPrice().Cmp(big.NewInt(0)) == 0 {
		return true, nil
	}
	return false, nil
}
```
we should update signer of parlia to londonSigner from EIP155Signer

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -278,7 +278,7 @@ func New(
 		validatorSetABIBeforeLuban: vABIBeforeLuban,
 		validatorSetABI:            vABI,
 		slashABI:                   sABI,
-		signer:                     types.NewEIP155Signer(chainConfig.ChainID),
+		signer:                     types.LatestSigner(chainConfig),
 	}
 
 	return c


consensus/parlia/parlia.go
```diff
@@ -278,7 +278,7 @@ func New(
 		validatorSetABIBeforeLuban: vABIBeforeLuban,
 		validatorSetABI:            vABI,
 		slashABI:                   sABI,
-		signer:                     types.NewEIP155Signer(chainConfig.ChainID),
```
sunny2022da, line 281: maybe types.LatestSigner() is better as it can make choice from the ChainID.

NathanBSC: accpeted

sunny2022da (approved):

brilliant-lx (approved):

PR #1686, opened 2023-06-09
eip3529tests: refactor tests

Mister-EA:
### Description

Refactor tests to avoid duplication of helper `testGasUsage` function by putting eip3529 tests in a new package.


sunny2022da:
I like this idea of reusing the code for `testGasUsage`! Let me try to extend it for other eip such as 2930 in Berlin, and I will try to see whether we can make it more general to not bind to 3529~.

PR #1695, opened 2023-06-13
test: Add EIP-2718 unit test for parlia ChainConfig

sunny2022da:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


sunny2022da:
This PR is a simplified replacement of #1693, as it is complicated to reuse methods in test as we found.

emailtovamos (approved):

Issue #1625, opened 2023-05-19
wanted: specify single hard fork height in config.toml


brilliant-lx:
# Rationale
There is a configure section in config.toml, like:
```
[Eth.Genesis.Config]
ChainID = 56
HomesteadBlock = 0
...
GibbsBlock = 23846001
PlanckBlock = 27281024
```
Currently, if you download the mainnet.zip or testnet.zip from https://github.com/bnb-chain/bsc/releases
, unzip it, then you get the config.toml without the above config section.
People can run full sync using this config.toml, since the default block height for each hard fork have been defined in code:  https://github.com/bnb-chain/bsc/blob/master/params/config.go#L172

But if people setup a single hard fork height, then the full sync could break, like:
```
[Eth.Genesis.Config]
ChainID = 56
PlanckBlock = 27281024
```
That is people can either skip the section or set all the hard forks, but can not just set a part of the fork height, it is not reasonable.
The expected behaviors should: if the dedicated fork height is not configured in config.toml, it should use the default value in code: https://github.com/bnb-chain/bsc/blob/master/params/config.go#L172

# Implementation
Just to fix it.

PR #1704, opened 2023-06-15
core: use reflection to set default block values

Mister-EA:
### Description
Set the deault block values of genesis config  using reflection, to avoid a long sequence of if-statements.

For more context:
https://github.com/bnb-chain/bsc/issues/1625

### Rationale
Maintaining a long sequence of if-statements with hardcoded block names is not scalable as more hard forks are introduced in BSC. Forgetting to update the code when a new hard fork happens would result in a wrong config. 

A more robust way is to use reflection to achieve the same result, without having to worry about keeping the file up-to-date every time there is a hard fork. This way the code is also kept short.

### Changes

Notable changes: 
* `genesis.go` : replace long series of if-statements with a `setDefaultBlockValues` method which uses the `.*Block` regex and reflection to achieve the same result.
* `genesis_test.go`: add tests to ensure that reflection modifies the relevant block values correctly, and leaves other fields untouched.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -23,6 +23,8 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


core/genesis.go
```diff
@@ -246,6 +248,32 @@ func SetupGenesisBlockWithOverride(db ethdb.Database, genesis *Genesis, override
 	return newcfg, stored, nil
 }
 
+// For any block in g.Config which is nil but the same block in defaultConfig is not
+// set the block in genesis config to the block in defaultConfig.
+// Reflection is used to avoid a long series of if statements with hardcoded block names.
+func (g *Genesis) setDefaultBlockValues(defaultConfig *params.ChainConfig) {
+	// Regex to match block names
+	blockRegex := regexp.MustCompile(`.*Block$`)
```
alexlucaci, line 256: I would suggest moving this into a `global` variable thus avoid compiling it every time the method is called

brilliant-lx: I think it is fine, since it will only be called only once on  node start up.

PR #1725, opened 2023-06-20
Uncomment log trace

alexlucaci:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


miner/worker.go
```diff
@@ -852,28 +852,30 @@ LOOP:
...
 		case errors.Is(err, core.ErrNonceTooLow):
 			// New head notification data race between the transaction pool and miner, shift
-			//log.Trace("Skipping transaction with low nonce", "sender", from, "nonce", tx.Nonce())
+			log.Trace("Skipping transaction with low nonce", "sender", from, "nonce", tx.Nonce())
 			txs.Shift()
 
 		case errors.Is(err, core.ErrNonceTooHigh):
 			// Reorg notification data race between the transaction pool and miner, skip account =
-			//log.Trace("Skipping account with hight nonce", "sender", from, "nonce", tx.Nonce())
+			log.Trace("Skipping account with hight nonce", "sender", from, "nonce", tx.Nonce())
```
Mister-EA, line 878: There is a typo here.

miner/worker.go
```diff
@@ -852,28 +852,30 @@ LOOP:
...
 		if tx.Protected() && !w.chainConfig.IsEIP155(env.header.Number) {
-			//log.Trace("Ignoring reply protected transaction", "hash", tx.Hash(), "eip155", w.chainConfig.EIP155Block)
+			log.Trace("Ignoring reply protected transaction", "hash", tx.Hash(), "eip155", w.chainConfig.EIP155Block)
 			txs.Pop()
 			continue
 		}
 		// Start executing the transaction
 		env.state.Prepare(tx.Hash(), env.tcount)
 
+		from, _ := types.Sender(env.signer, tx)
```
emailtovamos, line 862: Why no error check here?

emailtovamos (changes requested):

miner/worker.go
alexlucaci: I don't know if we should throw an error from here. I copy/paste what was written on line 851

Mister-EA (approved):

miner/worker.go
brilliant-lx: It was commented in line:847, the tx should have been check in the txpool, no they skip the error check here.
```
// Error may be ignored here. The error has already been checked
// during transaction acceptance is the transaction pool.
```

brilliant-lx (approved):

emailtovamos (approved):

PR #1727, opened 2023-06-22
discov: increase bucket size for bootnodes

MatusKysel:
### Increase bucket size for bootnodes
The present discovery protocol employs a Distributed Hash Table (DHT), which is organized into "buckets." Each bucket presently stores 16 entries, and there is a total of 17 buckets. This configuration allows for a capacity of 272 node records per individual node.
```
    bucketSize        = 16 // Kademlia bucket size
    hashBits          = len(common.Hash{}) * 8
    nBuckets          = hashBits / 15       // Number of buckets
```
I propose increasing the `bucketSize` value to 256, which is 16 times its current size. Since the bootnode binary is extremely lightweight, this change should not impact overall performance, and it could substantially enhance the discovery process.

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -40,9 +40,10 @@ import (
 )
 
 const (
-	alpha           = 3  // Kademlia concurrency factor
-	bucketSize      = 16 // Kademlia bucket size
-	maxReplacements = 10 // Size of per-bucket replacement list
+	alpha              = 3   // Kademlia concurrency factor
+	bucketSize         = 16  // Kademlia bucket size
+	bootNodeBucketSize = 256 // Bigger bucket size for boot nodes
+	maxReplacements    = 10  // Size of per-bucket replacement list
 
 	// We keep buckets for the upper 1/15 of distances because
 	// it's very unlikely we'll ever encounter a node that's closer.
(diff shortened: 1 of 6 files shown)


p2p/discover/table.go
```diff
@@ -115,6 +117,11 @@ func newTable(t transport, db *enode.DB, bootnodes []*enode.Node, log log.Logger
 		log:        log,
 		enrFilter:  filter,
 	}
+	if bootnode {
```
alexlucaci, line 120: Just a personal opinion, I would avoid using the else statement and assign `bucketSize` by default to the `tab.bucketSize` and change it in the if statement

emailtovamos:
Not sure if I asked this before during the call but any numbers on the size changes of the binary and performance changes? 
I am not sure how increasing bucketSize 16 times doesn't impact performance.

alexlucaci (approved):

brilliant-lx (approved):

kaber2:
Just a datapoint. I've manually added this change to the last version of BSC to improve the discovery process, and this has propelled a lot of the discovery and goleveldb methods to the top positions in the CPU profile.

So at least for me, there is a quite noticable impact on performance, partially caused by secondary effects like increased goleveldb load.

So you might actually want to profile this yourself.

kaber2:
Let me run a comparison with and without the change again.

Issue #114, opened 2021-03-15
[Feature] Provide a configuration option for logs rotation


gituser:
#### System information

Geth version: 1.0.6
OS & Version: Linux
Commit hash : latest release

#### Expected behaviour
Currently, geth creates every hour a new log file even if you set the `MaxBytesSize` to relatively large value (e.g. 100MB), e.g.:
```
[Node.LogConfig]
FileRoot = ""
FilePath = "bsc.log"
MaxBytesSize = 104857600
Level = "info" 
```

This results in:
```
-rw-r--r-- 1 bsc bsc 318390 Mar 12 17:00 bsc.log.2021-03-12_16
-rw-r--r-- 1 bsc bsc 317626 Mar 12 17:59 bsc.log.2021-03-12_17
-rw-r--r-- 1 bsc bsc 317914 Mar 12 19:00 bsc.log.2021-03-12_18
-rw-r--r-- 1 bsc bsc 318458 Mar 12 20:00 bsc.log.2021-03-12_19
-rw-r--r-- 1 bsc bsc 318739 Mar 12 20:59 bsc.log.2021-03-12_20
-rw-r--r-- 1 bsc bsc 319184 Mar 12 22:00 bsc.log.2021-03-12_21
-rw-r--r-- 1 bsc bsc 318645 Mar 12 23:00 bsc.log.2021-03-12_22
-rw-r--r-- 1 bsc bsc 318582 Mar 13 00:00 bsc.log.2021-03-12_23
-rw-r--r-- 1 bsc bsc 320143 Mar 13 01:00 bsc.log.2021-03-13_00
-rw-r--r-- 1 bsc bsc 319308 Mar 13 02:00 bsc.log.2021-03-13_01
-rw-r--r-- 1 bsc bsc 319031 Mar 13 03:00 bsc.log.2021-03-13_02
-rw-r--r-- 1 bsc bsc 319838 Mar 13 04:00 bsc.log.2021-03-13_03
-rw-r--r-- 1 bsc bsc 318024 Mar 13 05:00 bsc.log.2021-03-13_04
-rw-r--r-- 1 bsc bsc 317292 Mar 13 06:00 bsc.log.2021-03-13_05
-rw-r--r-- 1 bsc bsc 316828 Mar 13 06:59 bsc.log.2021-03-13_06
-rw-r--r-- 1 bsc bsc 319357 Mar 13 08:00 bsc.log.2021-03-13_07
[...]

PR #1735, opened 2023-06-26
log: added daily rotate option which overrides hourly

alexlucaci:
### Description

Adding an option to rotate longs on a daily basis rather to only rotate them on an hourly basis.

### Rationale

One wants the ability to have more flexibility when using log rotation. This PR tries to fix that by also adding an option to rotate the logs based on days. This new config option does not change the the mechanics of `MaxBytesSize` so its value also needs to be increased if one expects to have only one log file per day.

For backwards compatibility, if you don't provide `DailyRotate` in the config file, it defaults to false.

### Example

```
[Node.LogConfig]
FileRoot = ""
FilePath = "bsc.log"
MaxBytesSize = 104857600
Level = "info"
DailyRotate = true
```

### Changes

Notable changes: 
* ...


log/async_file_writer_test.go
```diff
@@ -1,24 +1,41 @@
...
-func TestWriter(t *testing.T) {
-	w := NewAsyncFileWriter("./hello.log", 100)
+func TestWriterHourly(t *testing.T) {
+	w := NewAsyncFileWriter("./hello.log", 100, false)
 	w.Start()
 	w.Write([]byte("hello\n"))
 	w.Write([]byte("world\n"))
 	w.Stop()
-	files, _ := ioutil.ReadDir("./")
+	files, _ := os.ReadDir("./")
```
Mister-EA, line 15: Why os instead of already existing ioutil?

log/async_file_writer.go
```diff
@@ -10,32 +10,43 @@ import (
...
 	stop chan struct{}
 	C    <-chan time.Time
 }
 
-func NewHourTicker() *HourTicker {
-	ht := &HourTicker{
+// NewTimeTicker notifies on a daily basis if dailyRotate is true, otherwise on an hourly basis
+func NewTimeTicker(dailyRotate bool) *TimeTicker {
+	ch := make(chan time.Time)
+	ht := TimeTicker{
```
Mister-EA, line 21: Since this could be a day ticker or hour ticker, the variable should rather be named `tt` ?

log/async_file_writer_test.go
alexlucaci: `ioutil.ReadDir` is deprecated since go 1.16

From `ioutil` package:

```
// Deprecated: As of Go 1.16, os.ReadDir is a more efficient and correct choice:
// it returns a list of fs.DirEntry instead of fs.FileInfo,
// and it returns partial results in the case of an error
// midway through reading a directory.
```

Mister-EA:
I think the request here https://github.com/bnb-chain/bsc/issues/114 is to make the rotation interval configurable by the user. In this implementation we are providing only one other extra option (daily) on top of the default hourly interval. I think it would make sense to have a more generic implementation, so that the rotation can be daily, weekly, monthly etc.

log/async_file_writer_test.go
```diff
@@ -1,24 +1,41 @@
...
 			t.Log(fn)
-			content, _ := ioutil.ReadFile(fn)
+			content, _ := os.ReadFile(fn)
+			t.Log(content)
+			os.Remove(fn)
+		}
+	}
+}
+
+func TestWriterDaily(t *testing.T) {
```
Mister-EA, line 27: I don't think these tests cases test if the ticker logic is correct - they only check that something was written to the log file. Can you think of a way to test the rotation logic? (I am thinking a more generic implementation with a variable rotation interval also has advantages in terms of ease of testing this logic).

brilliant-lx (approved):

PR #1754, opened 2023-07-07
feature: create a sentry node between validator node and mev relay

brilliant-lx:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


fetihzamani571 (approved):

marrkee:
hello could you explain what is this? is there related BEP? What will prevent validators and builders from exploiting users?

PR #1776, opened 2023-07-21
eth/protocols/handler: add packet sending condition, prevent send small packets frequently;

galaio:
### Description

eth/protocols/handler: add packet sending condition, prevent sending small packets frequently;

When profiling the p2p nodes, I found that `sendPooledTransactionHashes` func wastes the most CPU time.

image

After inspecting the egress packets, it only contains ~1.14 hash per msg with 2500 peers, which sends very small packets too frequently.

So if it's possible to let msg collect more data to send once?

Below is the optimised performance profile:
image

Now it contains ~13 hashes per msg with 1800 peers:
image

### Rational

Here add new config params:
1. `minTxPacketSize` = 1024, it's very hard to reach. When the size is reached, it will packet ~32 hashes per msg.;
2. `sendPacketTimeout` = 300 * `time.Millisecond`, less time will cause more times sending, and large time will cause more data. I tried 500ms, it will packet > 20 hashes per msg.

### Changes

Notable changes: 
* eth/protocols/handler: add packet sending condition, prevent send small packets frequently;

### DIFF
--- a/eth/protocols/eth/broadcast.go
+++ b/eth/protocols/eth/broadcast.go
@@ -18,6 +18,7 @@ package eth
 
 import (
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/gopool"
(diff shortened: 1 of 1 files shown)


melissamforbs (approved):

AlphaOmega113 (approved):

allformless:
will impact txs arriving in time

PR #1778, opened 2023-07-24
vote: backup validator sync votes from corresponding mining validator

NathanBSC:
### Description

backup validator sync votes from corresponding mining validator

### Rationale

node operator often start two nodes, one is mining and the other one backup,
these two nodes have the same consensus address and voting key.

when switch the backup validator to mining, because of lacking of votes recently, It may violate voting rules.
this PR address this issue, and sync votes for backup validators from mining validators and write synced votes to `voteJournal`

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vote/vote_manager.go
+++ b/core/vote/vote_manager.go
@@ -32,6 +32,10 @@ type VoteManager struct {
 	chainHeadCh  chan core.ChainHeadEvent
 	chainHeadSub event.Subscription
 
+	// used for backup validators to sync votes from corresponding mining validator
+	syncVoteCh  chan core.NewVoteEvent
+	syncVoteSub event.Subscription
+
 	pool    *VotePool
 	signer  *VoteSigner
 	journal *VoteJournal
@@ -46,9 +50,9 @@ func NewVoteManager(eth Backend, chainconfig *params.ChainConfig, chain *core.Bl
 		chain:       chain,
 		chainconfig: chainconfig,
 		chainHeadCh: make(chan core.ChainHeadEvent, chainHeadChanSize),
-
-		pool:   pool,
-		engine: engine,
+		syncVoteCh:  make(chan core.NewVoteEvent, voteBufferForPut),
+		pool:        pool,
+		engine:      engine,
 	}
 
 	// Create voteSigner.
@@ -69,6 +73,7 @@ func NewVoteManager(eth Backend, chainconfig *params.ChainConfig, chain *core.Bl
 
 	// Subscribe to chain head event.
 	voteManager.chainHeadSub = voteManager.chain.SubscribeChainHeadEvent(voteManager.chainHeadCh)
+	voteManager.syncVoteSub = voteManager.pool.SubscribeNewVoteEvent(voteManager.syncVoteCh)
 
 	go voteManager.loop()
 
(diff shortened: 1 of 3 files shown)


core/vote/vote_pool.go
```diff
@@ -151,6 +155,8 @@ func (pool *VotePool) putIntoVotePool(vote *types.VoteEnvelope) bool {
 		return false
 	}
 
+	pool.basicVerifiedVotesCh <- vote
```
NathanBSC, line 158: when `voting` is not enabled, here will dead lock

core/vote/vote_manager.go
```diff
@@ -164,6 +169,21 @@ func (voteManager *VoteManager) loop() {
 				voteManager.pool.PutVote(voteMessage)
 				votesManagerCounter.Inc(1)
 			}
+		case event := <-voteManager.syncVoteCh:
+			voteMessage := event.Vote
+			if voteManager.eth.IsMining() || !voteManager.signer.UsingKey(&voteMessage.VoteAddress) {
```
guagualvcha, line 174: Will it panic if `voteManager.signer` is nil?

core/vote/vote_signer.go
```diff
@@ -104,3 +104,7 @@ func (signer *VoteSigner) SignVote(vote *types.VoteEnvelope) error {
 	copy(vote.Signature[:], signature.Marshal()[:])
 	return nil
 }
+
+func (signer *VoteSigner) UsingKey(bLSPublicKey *types.BLSPublicKey) bool {
+	return types.BLSPublicKey(signer.pubKey) == *bLSPublicKey
```
guagualvcha, line 109: use `bytes.Equal(..,..)` instead.

core/vote/vote_manager.go
buddh0: voteManager.signer is assigned when New VoteManager
if voteManager.signer is nil，then `NewVoteManager` will fail, so loop will not to start

core/vote/vote_signer.go
buddh0: accepted and updated

guagualvcha (approved):

core/vote/vote_manager.go
```diff
@@ -69,6 +73,7 @@ func NewVoteManager(eth Backend, chainconfig *params.ChainConfig, chain *core.Bl
 
 	// Subscribe to chain head event.
 	voteManager.chainHeadSub = voteManager.chain.SubscribeChainHeadEvent(voteManager.chainHeadCh)
+	voteManager.syncVoteSub = voteManager.pool.SubscribeNewVoteEvent(voteManager.syncVoteCh)
```
brilliant-lx, line 77: need call `Unsubscribe()` when before exit the loop.
both `chainHeadSub` and `syncVoteSub`

NathanBSC: accepted.

normally, votepool and votemanager will only exit the loop when bsc is shutting down,
so not to unsubscribe() has no bad affect.
but there is really mini chance for votepool and votemanager exiting the loop unnormally.

brilliant-lx (approved):

GalaIO (approved):

PR #1784, opened 2023-07-28
fix: exclude same votes when doing malicious voting check

NathanBSC:
### Description

fix: exclude same votes when doing malicious voting check

### Rationale

even we have used `receivedVotes` to mark votes received,
in a corner cases, a vote has been pruned, then a reorg to older block happened
so the pruned votes can be received again, this will lead to alert malicious vote in a wrong way

so add more check in monitor.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/monitor/malicious_vote_monitor.go
+++ b/core/monitor/malicious_vote_monitor.go
@@ -58,6 +58,7 @@ func (m *MaliciousVoteMonitor) ConflictDetect(newVote *types.VoteEnvelope, pendi
 	if !(blockNumber+maliciousVoteSlashScope > pendingBlockNumber) {
 		blockNumber = pendingBlockNumber - maliciousVoteSlashScope + 1
 	}
+	newVoteHash := newVote.Data.Hash()
 	for ; blockNumber <= pendingBlockNumber+upperLimitOfVoteBlockNumber; blockNumber++ {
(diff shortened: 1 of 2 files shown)


consensus/parlia/parlia.go
```diff
@@ -868,6 +868,9 @@ func (p *Parlia) assembleVoteAttestation(chain consensus.ChainHeaderReader, head
 	voteAddrSet := make(map[types.BLSPublicKey]struct{}, len(votes))
 	signatures := make([][]byte, 0, len(votes))
 	for _, vote := range votes {
+		if _, ok := voteAddrSet[vote.VoteAddress]; ok {
+			return fmt.Errorf("assemble VoteAttestation failed, double sign by %s", common.Bytes2Hex(vote.VoteAddress[:]))
```
brilliant-lx, line 872: Would it be possible that a validator could have 2 votes with same `vote.Data`?
since we have already checked the `vote.Data`

NathanBSC: this check both work for 'same votes' and 'different votes' with the same targetNumber from one VoteAddress

NathanBSC: @brilliant-lx u r right, no need check here

brilliant-lx (approved):

GalaIO (approved):

PR #1789, opened 2023-07-31
rpc: add GetFinalizedHeader/Block to simplify using the fast finality feature

NathanBSC:
### Description

add GetFinalizedHeader/Block to simplify using the fast finality feature

### Rationale

the fast finality feature  may not work stable in early time or in some extreme scenes, 
then the chain will decade back to probabilistic finality.
we need more simple APIs to handle block finality easily regardless of the status of fast finality mechanism.
here comes in this PR
```
//  eth_getFinalizedHeader returns the requested finalized block header.
//   - probabilisticFinalized should be in range [2,21],
//     then the block header with number `max(fastFinalized, latest-probabilisticFinalized)` is returned
eth_getFinalizedHeader(probabilisticFinalized int64)

//  eth_getFinalizedBlock returns the requested finalized block.
//   - probabilisticFinalized should be in range [2,21],
//     then the block with number `max(fastFinalized, latest-probabilisticFinalized)` is returned
//   - When fullTx is true all transactions in the block are returned, otherwise
//     only the transaction hash is returned.
eth_getFinalizedBlock(probabilisticFinalized int64 ,fullTx bool)
```
### Example
you can use it as following:
```
curl -X POST "http://localhost:8545/" -H "Content-Type: application/json"  --data '{"jsonrpc":"2.0","method":"eth_getFinalizedHeader","params":[11],"id":1}'

curl -X POST "http://localhost:8545/" -H "Content-Type: application/json"  --data '{"jsonrpc":"2.0","method":"eth_getFinalizedBlock","params":[14, true],"id":1}'
```
### Changes

Notable changes: 
* ...

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -54,6 +54,14 @@ import (
 
 const UnHealthyTimeout = 5 * time.Second
 
+// max is a helper function which returns the larger of the two given integers.
+func max(a, b int64) int64 {
+	if a > b {
+		return a
+	}
+	return b
+}
+
 // PublicEthereumAPI provides an API to access Ethereum related information.
 // It offers only methods that operate on public data that is freely available to anyone.
 type PublicEthereumAPI struct {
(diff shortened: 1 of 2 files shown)


internal/ethapi/api.go
```diff
@@ -773,6 +781,27 @@ func (s *PublicBlockChainAPI) Health() bool {
 	return true
 }
 
+// GetFinalizedBlock returns the requested finalized block.
+// * probabilisticFinalized should be in range [2,21], then the block with number `max(voteFinalized, latest-probabilisticFinalized)` is returned
+func (s *PublicBlockChainAPI) GetFinalizedBlock(ctx context.Context, probabilisticFinalized int64, fullTx bool) (map[string]interface{}, error) {
```
guagualvcha, line 786: what does fulltx mean?

Pythonberg1997 (dismissed):

internal/ethapi/api.go
NathanBSC: add in comment
When fullTx is true all transactions in the block are returned, otherwise  only the transaction hash is returned.

guagualvcha (approved):

Pythonberg1997 (approved):

bascrew:
But these two new APis is not supported by web3j for instance.

The API to get the finality block vis status can be stable with later fix?

bascrew:
hey guy,

why don't you add this into the `eth_getBlockByNumber` API then it will be compatible with other EVM chains?

PR #1791, opened 2023-08-02
finality: add more check to ensure reuslt of assembleVoteAttestation

NathanBSC:
### Description

finality: add more check to ensure reuslt of assembleVoteAttestation

### Rationale

drop attestation when two same votes included or vote with invalid validator included

by the way, these two senses have been excluded by func basicVerify and verifyVote, 
but assembleVoteAttestation should depend them less

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -882,6 +882,11 @@ func (p *Parlia) assembleVoteAttestation(chain consensus.ChainHeaderReader, head
 			attestation.VoteAddressSet |= 1 << (valInfo.Index - 1) //Index is offset by 1
 		}
 	}
+	validatorsBitSet := bitset.From([]uint64{uint64(attestation.VoteAddressSet)})
+	if validatorsBitSet.Count() < uint(len(signatures)) {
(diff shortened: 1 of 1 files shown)


consensus/parlia/parlia.go
```diff
@@ -882,6 +882,11 @@ func (p *Parlia) assembleVoteAttestation(chain consensus.ChainHeaderReader, head
 			attestation.VoteAddressSet |= 1 << (valInfo.Index - 1) //Index is offset by 1
 		}
 	}
+	validatorsBitSet := bitset.From([]uint64{uint64(attestation.VoteAddressSet)})
+	if validatorsBitSet.Count() < uint(len(signatures)) {
```
brilliant-lx, line 886: why not just: 
```
if validatorsBitSet.Count()  < cmath.CeilDiv(len(snap.Validators)*2, 3) {
return fmt.Errorf(...)
}
```
if there are duplicated data in signatures,  bitset count would not equal to len(signatuer), but could still >=2/3 validators size, right?

brilliant-lx (approved):

PR #1798, opened 2023-08-03
4844: All changes

emailtovamos:
### Description

Missing things that are undergoing:

- Ensure the blob transaction goes through without errors: currently there's RLP-SSZ mismatch error.
- Storage of blobs
- Query functionality
- P2P
- Some more...

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck:
4844 implementation has redesigned and implemented in this PR: https://github.com/bnb-chain/bsc/pull/2279

PR #1810, opened 2023-08-08
4844 blobs storage

alexlucaci:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


emailtovamos:
It would be great to have functionality to create database if it doesn't exist. 
If it exist then open the existing one. 
Similar to [here](https://github.com/bnb-chain/bsc/blob/e4381a5d250b177950cdff6855c00ae14b70365c/node/node.go#L604). 
Or perhaps the way Prysm does it (I haven't looked at how exactly they do it)

I think currently it always opens a new fresh database unless I am mistaken?

PR #1839, opened 2023-08-23
Update init-network command

Mister-EA:
### Description
This PR modifies the behavior of the `init-network` command so that:

- it no longer writes the genesis config to the `config.toml` of the nodes, but writes it to `genesis.json` instead.
- nodes on the same host use different ports, while nodes in different hosts use the same port.
-  split the long function into smaller functions.
- Uses bootstrap nodes instead of static nodes to setup the p2p connections among nodes.

### Rationale
The `init-network` command is used by the [node-deploy](https://github.com/bnb-chain/node-deploy/) repository to generate the configs for a local cluster of BSC nodes. The command receives a size for the network and `config.toml` and `genesis.json` and produces a folder structure where the subfolders are the configurations for the individual nodes in the network.

The problems with the current implementation that are addressed in this PR are the following:
- The command parses the input `config.toml` and `genesis.json` files, and combines them into a `config.toml` file written in each node's directory . The problem is that some hard fork blocks in the genesis config might be nil, and if the genesis config is written to the `config.toml`, these blocks will be serialized into `<nil>`, which will lead to parsing errors when the nodes in the cluster are started. The `node-deploy` repo deals with this by modifying each generated `config.toml` file to delete lines with nil blocks. However, a better solution would be to not include the genesis config in `config.toml` as it was never intended to be included there, and instead write the `genesis.json` to the node's directories.
- Currently all nodes use the same port to listen for p2p connections, however this doesn't work when the nodes are all running on the same host. `node-deploy` deals with this by manually editing the port numbers in the generated `config.toml` files, but a better solution would be to have custom logic so that when the host is localhost the nodes are listening on different ports, while if the nodes are running on different hosts the same port may be used.

### Example

`bin/geth init-network --init.dir ./.local/bsc/clusterNetwork --init.size=3  --config ./config.toml ./genesis/genesis.json`

### Changes

Notable changes: 
* `cmd/geth/chaincmd.go`

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -318,16 +318,18 @@ func initNetwork(ctx *cli.Context) error {
 		utils.Fatalf("init.dir is required")
 	}
 	size := ctx.Int(utils.InitNetworkSize.Name)
+	if size <= 0 {
+		utils.Fatalf("size should be greater than 0")
+	}
 	port := ctx.Int(utils.InitNetworkPort.Name)
+	if port <= 0 {
+		utils.Fatalf("port should be greater than 0")
+	}
 	ipStr := ctx.String(utils.InitNetworkIps.Name)
 	cfgFile := ctx.String(configFileFlag.Name)
 
 	if len(cfgFile) == 0 {
 		utils.Fatalf("config file is required")
-	}
-
-	if size <= 0 {
-		utils.Fatalf("size should be greater than 0")
 	}
 
 	ips, err := parseIps(ipStr, size)
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


cmd/geth/chaincmd.go
```diff
@@ -270,61 +257,140 @@ func initNetwork(ctx *cli.Context) error {
...
+// Create config for node i in the cluster
+func createNodeConfig(baseConfig gethConfig, enodes []*enode.Node, ip string, port int, size int, i int) gethConfig {
+	baseConfig.Node.HTTPHost = ip
+	baseConfig.Node.P2P.ListenAddr = fmt.Sprintf(":%d", port+i)
+	baseConfig.Node.P2P.BootstrapNodes = make([]*enode.Node, size-1)
+	// Set the P2P connections between this node and the other nodes
+	for j := 0; j < i; j++ {
+		baseConfig.Node.P2P.BootstrapNodes[j] = enodes[j]
+	}
+	for j := i + 1; j < size; j++ {
```
emailtovamos, line 286: Should this be `size` or `size+1` ?

emailtovamos: Is size of `enodes` and size of `baseConfig.Node.P2P.BootstrapNodes` supposed to be the same?

Mister-EA: No, `baseConfig.Node.P2P.BootstrapNodes` length is `size-` while `enodes` length is `size`. This is because the node does not include itself in the list of bootstrap nodes (i.e. only the other nodes).

emailtovamos (approved):

NathanBSC (approved):

brilliant-lx (approved):

PR #1858, opened 2023-09-07
vote: check consensus key match vote key before voting

NathanBSC:
### Description

add a description of your changes here...

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes: 
* add each change in a bullet point here
* ...

PR #1852, opened 2023-09-05
discov: add hardcoded bootnodes

MatusKysel:
### Description

This change will add hardcoded addresses to our bootnodes

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes: 
* add each change in a bullet point here
* ...

PR #1860, opened 2023-09-08
draft release v1.2.12

brilliant-lx:
### Description
Release v1.2.12 is a maintenance release for BSC mainnet and testnet.
It mainly support BootNodes on BSC mainnet, it would be easier for user to sync with the BSC network.
And it also fixed three bugs to make the node more reliable, not critical bugs, pls check the change list for detail.

### Change Log
FEATURE
* [\#1852](https://github.com/bnb-chain/bsc/pull/1852) discov: add hardcoded bootnodes

BUGFIX
* [\#1844](https://github.com/bnb-chain/bsc/pull/1844) crypto: Update BLST to v0.3.11
* [\#1854](https://github.com/bnb-chain/bsc/pull/1854) fetcher: no import blocks before or equal to the finalized height
* [\#1855](https://github.com/bnb-chain/bsc/pull/1855) eth/tracers: trace system tx should add intrinsicGas

IMPROVEMENT
* [\#1839](https://github.com/bnb-chain/bsc/pull/1839) Update init-network command
* [\#1858](https://github.com/bnb-chain/bsc/pull/1858) vote: check consensus key match vote key before voting

### Example
NA

### Compatibility
No Impact to users.


core/vote/vote_manager.go
```diff
@@ -132,11 +135,22 @@ func (voteManager *VoteManager) loop() {
 
 			curHead := cHead.Block.Header()
 			// Check if cur validator is within the validatorSet at curHead
-			if !voteManager.engine.IsActiveValidatorAt(voteManager.chain, curHead) {
+			if !voteManager.engine.IsActiveValidatorAt(voteManager.chain, curHead,
+				func(bLSPublicKey *types.BLSPublicKey) bool {
```
Mister-EA, line 139: Question: Why use a custom function here when all you need to check is if the `blsPublicKey` matches `voterManager.signer.Pubkey` ? Are there any other use cases?

NathanBSC: > Question: Why use a custom function here when all you need to check is if the `blsPublicKey` matches `voterManager.signer.Pubkey` ? Are there any other use cases?

as before, `IsActiveValidatorAt` only check consensus key 
after fast finality feature brought in,  IsActiveValidatorAt may want to check votekey,
so a new param `checkVoteKeyFn` is added, 
      if it's nil, meaning no need to check; if it's not nil, then should check
if in instead, we pass a pointer `* blsPublicKey`, when we pass nil, the meaning is confusing:
    `nil` meaning not to checking or desire votoKey is nil?

Mister-EA: I think you can define `isActiveValidatorAt(chain consensus.ChainHeaderReader, header *types.Header, blsPublicKey *types.BLSPublicKey) bool` , and pass `voteManager.signer.Pubkey[:]`  to it.  The implementation of `isActiveValidatorAt` should check that this voting key matches the validator's vote address.  

This way you don't need the custom function.

Mister-EA (approved):

NathanBSC (approved):

buddh0 (approved):

PR #1865, opened 2023-09-12
Big merge v1.10.16 v1.12.2 fix integration test

NathanBSC:
### Description

fix integration test

### Rationale

1. miner.etherbase must be set for validator now
2. if BootstrapNodes is a empty slice ,not nil, and BootnodesFlag is set,
	we should add extra bootnodes from BootnodesFlag
3. copy nil baseFee will cause panic

### Example
```
make truffle-test pass
transfer using metamask pass
```
### Changes

Notable changes: 
* ...

### DIFF
--- a/p2p/discover/table.go
+++ b/p2p/discover/table.go
@@ -108,7 +108,7 @@ type bucket struct {
 	index        int
 }
 
-func newTable(t transport, db *enode.DB, cfg Config, filter NodeFilterFunc) (*Table, error) {
+func newTable(t transport, db *enode.DB, cfg Config) (*Table, error) {
 	cfg = cfg.withDefaults()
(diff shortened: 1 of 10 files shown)


brilliant-lx (approved):

p2p/discover/table.go
```diff
@@ -108,7 +108,7 @@ type bucket struct {
 	index        int
 }
 
-func newTable(t transport, db *enode.DB, cfg Config, filter NodeFilterFunc) (*Table, error) {
```
brilliant-lx, line 111: @MatusKysel could you help check if this change is ok?

CripTz00:
Good

CripTz00 (approved):

PR #1873, opened 2023-09-15
core/state: handle account destruction after updating account state

NathanBSC:
### Description

handle account destruction after updating account state

#### Bugfix:
1. `handleDestruction` depend on `accounts` and this map will be updated when `updateStateObject`,
so move `handleDestruction` after `updateStateObject`, otherwise will **pathdb** will behavior unexpectedly.
2. check err returned by `commmitTrie`
3. revert logic of New statedb to support new a statedb when OpenTrie failed(for example, no trie data)

#### Improve performance
1.  avoid copy accounts and storages twice in `copyInternal`
2. avoid both update accounts in `AccountsIntermediateRoot` and `StateIntermediateRoot`
3. make two funcs in `commitFuncs` run in parallel

#### Others:
 use `types.EmptyRootHash` instead of `common.Hash{}` to express empty root hash of account

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -336,7 +336,6 @@ func (s *stateObject) updateTrie() (Trie, error) {
 		}
 		var v []byte
 		if value != (common.Hash{}) {
-			// Encoding []byte cannot fail, ok to ignore the error.
 			value := value
 			v = common.TrimLeftZeroes(value[:])
 		}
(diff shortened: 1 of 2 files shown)


NathanBSC:
Integration test failed, to be resolved.
PS: even checkout code from develop and add a comment will fail, may be some problems like lib version

brilliant-lx (approved):

GalaIO (approved):

Issue #1125, opened 2022-10-10
deadlock when calling eth.getLogs if fromBlock is small


github1youlc:
#### System information

Geth version: 
Geth
Version: 1.1.15
Architecture: amd64
Go Version: go1.18.3
Operating System: linux

OS & Version: Linux

#### Expected behaviour
After attaching geth, this script
> eth.getLogs({"fromBlock": "0x0", "toBlock": "0x3E8", "topics": [["0x3c13bc30b8e878c53fd2a36b679409c073afd75950be43d8858768e956fbc20e"]]})

should response normally.

#### Actual behaviour
 It block forever. And some goroutines trapped in deadlock.

sync.runtime_SemacquireMutex+0x24                                               /home/ubuntu/opt/go/src/runtime/sema.go:71
sync.(*Mutex).lockSlow+0x164                                                    /home/ubuntu/opt/go/src/sync/mutex.go:162
sync.(*Mutex).Lock+0x52                                                         /home/ubuntu/opt/go/src/sync/mutex.go:81
sync.(*Once).doSlow+0x26                                                        /home/ubuntu/opt/go/src/sync/once.go:64
sync.(*Once).Do+0x44                                                            /home/ubuntu/opt/go/src/sync/once.go:59
github.com/ethereum/go-ethereum/core/bloombits.(*MatcherSession).Close+0x14     /home/ubuntu/projects/bsc-workspace/bsc-geth/core/bloombits/matcher.go:528
github.com/ethereum/go-ethereum/eth/filters.(*Filter).indexedLogs+0x4c9         /home/ubuntu/projects/bsc-workspace/bsc-geth/eth/filters/filter.go:201
github.com/ethereum/go-ethereum/eth/filters.(*Filter).Logs+0x1ab                /home/ubuntu/projects/bsc-workspace/bsc-geth/eth/filters/filter.go:162
github.com/ethereum/go-ethereum/eth/filters.(*PublicFilterAPI).GetLogs+0x192    /home/ubuntu/projects/bsc-workspace/bsc-geth/eth/filters/api.go:355

sync.runtime_Semacquire+0x24                                                            /home/ubuntu/opt/go/src/runtime/sema.go:56
sync.(*WaitGroup).Wait+0x51                                                             /home/ubuntu/opt/go/src/sync/waitgroup.go:136
[...]

PR #1895, opened 2023-09-25
core/bloombits: fix deadlock when matcher session hits an error

msmania:
### Description

Calling `eth_getLogs` for pruned blocks causes deadlock.  The exact same issue was reported as #1125 but it was never fixed.

This patch has been merged in the upstream and will be included in the next version v1.13.2.  More technical details are described in https://github.com/ethereum/go-ethereum/pull/28184.

### Rationale

Every node runners running pruned nodes encounter this deadlock.  If a request causes deadlock, the request fail with timeout.  To make matters worse, piled-up deadlock goroutines keep consuming system resource and the process will eventually be killed. due to OOM.

With this patch, the request immediately returns a jsonrpc error.

### Example

Sending a single `eth_getLogs` for a single pruned block is enough to reproduce deadlock.

### Changes

Notable changes: 
* Porting https://github.com/ethereum/go-ethereum/commit/c2cfe35f121cb88650b4d90c958bcc4214d0ce7f from go-ethereum to bsc

### DIFF
--- a/core/bloombits/matcher.go
+++ b/core/bloombits/matcher.go
@@ -633,13 +633,16 @@ func (s *MatcherSession) Multiplex(batch int, wait time.Duration, mux chan chan 
 			request <- &Retrieval{Bit: bit, Sections: sections, Context: s.ctx}
 
 			result := <-request
+
+			// Deliver a result before s.Close() to avoid a deadlock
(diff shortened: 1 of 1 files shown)


brilliant-lx (approved):
thx

buddh0 (approved):

zzphust (approved):

PR #1897, opened 2023-09-26
cmd/geth: fix listern port when initNetwork

NathanBSC:
### Description

cmd/geth: fix listern port when initNetwork

### Rationale

otherwise, nodes may can't link to each other

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -269,7 +269,7 @@ func createPorts(ipStr string, port int, size int) []int {
 // Create config for node i in the cluster
 func createNodeConfig(baseConfig gethConfig, enodes []*enode.Node, ip string, port int, size int, i int) gethConfig {
 	baseConfig.Node.HTTPHost = ip
-	baseConfig.Node.P2P.ListenAddr = fmt.Sprintf(":%d", port+i)
+	baseConfig.Node.P2P.ListenAddr = fmt.Sprintf(":%d", port)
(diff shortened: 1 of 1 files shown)


cmd/geth/chaincmd.go
```diff
@@ -277,7 +277,7 @@ func createPorts(ipStr string, port int, size int) []int {
 // Create config for node i in the cluster
 func createNodeConfig(baseConfig gethConfig, enodes []*enode.Node, ip string, port int, size int, i int) gethConfig {
 	baseConfig.Node.HTTPHost = ip
-	baseConfig.Node.P2P.ListenAddr = fmt.Sprintf(":%d", port+i)
+	baseConfig.Node.P2P.ListenAddr = fmt.Sprintf(":%d", port)
```
Mister-EA, line 280: What was the reason for this change? The nodes in the cluster need to have different listen addresses.

NathanBSC: ListenAddr is not the same with enodes now

NathanBSC: if they are in one machine , they should listen different ports
if they are in different machines , they can listen the same ports ( or not)

but the listen port ,should keep the same with port in enodes

NathanBSC: for example
using node-deploy
image

NathanBSC:
have test init-network in different ips mode, works well now

Mister-EA (approved):

brilliant-lx (approved):

GalaIO (approved):

PR #1908, opened 2023-10-08
core/state: skip handleDestruction in hash based mode

NathanBSC:
### Description

core/state: skip handleDestruction in hash based mode

### Rationale

1. storage deletion is useless for hash based db, so skip it.
2. `TestStateChanges` fails now, but it is designed for path based only,
    and is fixed in https://github.com/bnb-chain/bsc/pull/1909,
     so disable it for now.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1422,7 +1422,13 @@ func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root 
 // In case (d), **original** account along with its storages should be deleted,
 // with their values be tracked as original value.
 func (s *StateDB) handleDestruction(nodes *trienode.MergedNodeSet) (map[common.Address]struct{}, error) {
+	// Short circuit if geth is running with hash mode. This procedure can consume
+	// considerable time and storage deletion isn't supported in hash mode, thus
+	// preemptively avoiding unnecessary expenses.
 	incomplete := make(map[common.Address]struct{})
+	if s.db.TrieDB().Scheme() == rawdb.HashScheme {
+		return incomplete, nil
+	}
 	for addr, prev := range s.stateObjectsDestruct {
 		// The original account was non-existing, and it's marked as destructed
 		// in the scope of block. It can be case (a) or (b).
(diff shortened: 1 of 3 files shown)


brilliant-lx (approved):
it is ok for Hash based, but Path based will need to handle it in the future.

core/state/statedb.go
```diff
@@ -44,6 +44,14 @@ import (
 
 const defaultNumOfSlots = 100
 
+const (
+	// storageDeleteLimit denotes the highest permissible memory allocation
+	// employed for contract storage deletion.
+	// for now, only hashScheme-based  supported, so no need to delete storages,
+	// pathScheme-based should be supported with `fastDeleteStorage` together.
```
Fynnss, line 51: Does the `FastDeleteStorage` mentioned here refer to this code section (https://github.com/ethereum/go-ethereum/blob/4e1e37323d87967708b58546f040990db7c2820c/core/state/statedb.go#L951-L993) in the go-ethereum repository?

 If so, I don't think pbss is relevant because FastDeleteStorage only deletes the storage inside the snapshot. It is only executed when snap is enabled.

core/state/statedb.go
```diff
@@ -1362,7 +1370,7 @@ func (s *StateDB) deleteStorage(addr common.Address, addrHash common.Hash, root
 	)
 	for it.Next(true) {
 		// arbitrary stateSize limit, make it configurable
-		if stateSize+nodeSize > 512*1024*1024 {
+		if stateSize+nodeSize > storageDeleteLimit {
```
RenRick, line 1373: It is better to skip handling destructions than to change the threshold to 0.

NathanBSC: image
ok. go-ethereum has skip it now

core/state/statedb.go
NathanBSC: It leverages the associated state snapshot for fast storage iteration
It's just a improvement.

I will modify this PR

RenRick (approved):

joeylichang (approved):

PR #1909, opened 2023-10-08
core/state: implement fast storage deletion (#27955)

NathanBSC:
This changes implements faster post-selfdestruct iteration of storage slots for deletion, by using snapshot-storage+stacktrie to recover the trienodes to be deleted. This mechanism is only implemented for path-based schema.

For hash-based schema, the entire post-selfdestruct storage iteration is skipped, with this change, since hash-based does not actually perform deletion anyway.

---------

### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -42,7 +42,12 @@ import (
 	"github.com/ethereum/go-ethereum/trie/triestate"
 )
 
-const defaultNumOfSlots = 100
+const (
+	defaultNumOfSlots = 100
+	// storageDeleteLimit denotes the highest permissible memory allocation
+	// employed for contract storage deletion.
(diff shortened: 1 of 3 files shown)


brilliant-lx (changes requested):
pls hold this PR, until v1.3.0 release

joeylichang:
I think we should follow the PR template spec to organize PR descriptions.
BTW, lint commur reports errors.

buddh0:
> I think we should follow the PR template spec to organize PR descriptions. BTW, lint commur reports errors.

I will close this PR, https://github.com/bnb-chain/bsc/pull/1916 is prefered

buddh0:
I will close this PR, https://github.com/bnb-chain/bsc/pull/1916 is preferred

PR #1939, opened 2023-10-24
dependency: go version to 1.20 and some dependencies in go.mod

brilliant-lx:
### Description
upgrade some dependencies for quality enhancement.
- prysm from v4.0.2 to v4.0.8
  It is mainly for go-libp2p upgrade, which has a potential OOM risk. go-libp2p was added by prysm indirectly, so upgrade prysm instead. It will also upgrade go-libp2p to v0.27.8 and it needs golang v1.20+ to compile, but prsym v4.0.8 can not be compiled with golang v1.21. So only upgrade to v1.20.
- github.com/docker/docker from v20.10.19 to 20.10.24
 actually it is not an issue for BSC, since BSC only use it for test purpose, but upgrade also is ok.

And after golang v1.20, the usage of math/rand module has been changed, see: https://pkg.go.dev/math/rand
### Example
NA

### Changes
NA


buddh0:
fail to compile with go@1.19
so we should modify README.md
```
Building `geth` requires both a Go (version 1.19 or later) and a C compiler (GCC 5 or higher). You can install
```

buddh0 (approved):

Mister-EA (approved):

PR #1948, opened 2023-10-26
performance: commitTire concurrently during Commit

brilliant-lx:
### Description
In `StateDB::Commit()`, run `commmitTrie()` concurrently to improve the performance.
Previous commitTrie was changed to be run serially for `--pipecommit`, but actually, it is ok to run it concurrently.

### Rationale
It was broken after v1.3.0 due to a object concurrent access safety, to reenable it

### Implementation
the safety was about: `StateDB.storages`
```
type StateDB struct {
  ...
  storages       map[common.Hash]map[common.Hash][]byte 
  ...
}
```
it is used to store the modified KV pair for each address on updateTrie which is called during block validation phase, the key is with Keccak hashed.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1808,6 +1808,7 @@ func (s *StateDB) Commit(block uint64, failPostCommitFunc func(), postCommitFunc
 		go commmitTrie()
 	} else {
 		defer s.StopPrefetcher()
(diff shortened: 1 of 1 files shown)


buddh0 (approved):

GalaIO (approved):

brilliant-lx:
run full sync for 2 days, it is ok, so merge the PR.

PR #1979, opened 2023-11-10
fix: upgrade pebble and improve config

flywukong:
### Description

1. upgrade pebble and improve config to enhance pebble as the bankup of leveldb
2. improve config of pebble:  increased BlockSizes so that the fixed cost of a block cache miss is reduced and  use larger target file sizes which is same as CockroachDB . The config change is good for read caching performance.

### Rationale

the block chasing performance of the geth with pebble （after upgrade） is not bad than leveldb,  so it is suitable as an alternative to leveldb
image
the yellow one is leveldb ,  pebble  avg:  371ms (avg value of 24 hours), leveldb avg:   380ms (avg value of 24 hours)

### Example

### Changes

Notable changes: 
* ...


ethdb/pebble/pebble.go
```diff
@@ -116,6 +117,16 @@ func (d *Database) onWriteStallEnd() {
 	d.writeDelayTime.Add(int64(time.Since(d.writeDelayStartTime)))
 }
 
+// panicLogger is just a noop logger to disable Pebble's internal logger.
+type panicLogger struct{}
+
+func (l panicLogger) Infof(format string, args ...interface{}) {
+}
```
sysvm, line 124: move `}` to 124 lines

RenRick (approved):

sysvm (approved):
LGTM

Fynnss (approved):

joeylichang (approved):

Issue #1984, opened 2023-11-13
v1.3.1-beta  can't start with flags `--tries-verify-mode=none --pruneancient=true`


du5:
It works fine on v1.3.0, please fix it

link #1974 

<img width="1217" alt="image" src="https://github.com/bnb-chain/bsc/assets/54979375/edaf88b8-a9aa-426b-989b-b421021eb13c">

<img width="1274" alt="image" src="https://github.com/bnb-chain/bsc/assets/54979375/d4b1f21e-5f81-4f07-9cb4-7333e55d7273">

Issue #1982, opened 2023-11-12
v1.3.1-beta stuck, updated from v1.3.0


tebeletbl:
i just tried to update my geth from v1.3.0 to v.1.3.1 
![image](https://github.com/bnb-chain/bsc/assets/102418244/f57fb3dc-50eb-4ad7-a576-8e4537b1f0c0)
but stuck here

CocoStarZ:
could you pls provide more info? are you able to upgrade to 1.2.13?

tebeletbl:
1.2.13 working just fine, i tried 1.3.0 but sometimes misses few pending tx, then i updated to 1.3.1 and node wont start, now i downgraded to 1.2.13 since 1.3.1 wont start and 1.3.0 seems unstable

weiihann:
I just tried starting up a new node with v1.3.1-beta and it works fine. Perhaps there are some configurations previously that caused the issue. Can you provide the following details:
1. Node start-up command
2. Logs with trace enabled (i.e. --verbosity=5)

tebeletbl:
this is my start up command 
for 1.3.0 and 1.2.13, is there anything i need to change for 1.3.1
./geth_linux --config ./config.toml --datadir ./mainnet --cache 48000 --rpc.allow-unprotected-txs --txlookuplimit 0 --http --maxpeers 850 --ws --syncmode=full --snapshot=true --tries-verify-mode=none --pruneancient=true --diffblock=5000

weiihann:
can you try to run it again with `--verbosity=5` can show the logs?

tebeletbl:
Nov 13 15:21:01 mine2 bash[55126]: Fatal: Could not open database: open /home/geth/mainnet/geth/chaindata/ancient/bodies.cidx: no such file or directory
Nov 13 15:21:01 mine2 systemd[1]: geth.service: Main process exited, code=exited, status=1/FAILURE
Nov 13 15:21:01 mine2 systemd[1]: geth.service: Failed with result 'exit-code'.

tebeletbl:
thanks, waiting for next update

PR #1986, opened 2023-11-14
fix: v1.3.1-beta cannot start with `--pruneancient=true`

weiihann:
### Description
Fix the node unable to start issue if `--pruneancient=true` by creating a non-read-only database. 

Refer to #1984 and #1982.

### Rationale
When initializing a readonly database, it will attempt to open a freezer file for read only (refer to code [here](https://github.com/bnb-chain/bsc/blob/v1.3.1-beta/core/rawdb/freezer_table.go#L147-L164)). Since ancient files are pruned and no longer exist, the read file operation will fail and return error. 

### Example
Refer to #1984 and #1982.

### Changes
Create a non-read-only database if `--pruneancient=true`.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2355,7 +2355,7 @@ func tryMakeReadOnlyDatabase(ctx *cli.Context, stack *node.Node) ethdb.Database 
 	// If datadir doesn't exist we need to open db in write-mode
 	// so database engine can create files.
 	readonly := true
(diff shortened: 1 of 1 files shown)


zzphust (approved):

buddh0:
avoid to open freezer is more rational when --pruneancient=true ?

buddh0:
approved

buddh0 (approved):

tangzhiwei429:
i got this issue also

Issue #1999, opened 2023-11-24
the flag: --txlookuplimit no longer works, it has been replaced by --history.transactions after v1.3.1-beta


brilliant-lx:
#### System information
Geth version: v1.3.1-beta
OS & Version: Linux
Commit hash : 1dca486622ae46636b888fa10f8a8dbf0879e7e8

#### Expected behaviour
With this flag: `--txlookuplimit 0`, the node can maintain the txIndex of the entire chain.

#### Actual behaviour
But it only maintains the default number of blocks, which is 2,350,000.
It would even cause the node run `unindexTransactions(...)`, to remove the txIndex of ancient blocks, which would have great impact the performance, since `unindexTransactions(...)` will remove lots of KV item from the database. The `unindexTransactions(...)` could last for more then 10 hours, during this period the node will have very poor performance.

The code that deprecated `--txlookuplimit`
https://github.com/bnb-chain/bsc/commit/720ff1fe57810128031d32c202f0437fababa55e

#### Suggestion
To avoid misuse of this flag, it would be better to make it crash on start and show the error message.
We can add a log like: log.Fatal("--txlookuplimit is deprecated, use --history.transactions instead ")

PR #2000, opened 2023-11-24
cmd/utils: exit process if txlookuplimit flag is set

weiihann:
### Description
Crash node and exit process if `--txlookuplimit` flag is set instead of `history.transactions`.

### Rationale
If users still use `--txlookuplimit` which is deprecated instead of `history.transactions`, then it maintains the default number of blocks to maintain the transaction index for. This may cause the node to unindex transactions, which would affect the performance.

Refer to #1999 for more details.

### Example
N/A

### Changes
- Wait 3 seconds before `os.Exit(1)`, this ensures the log message is printed before exiting.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1948,14 +1948,12 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	// Parse transaction history flag, if user is still using legacy config
 	// file with 'TxLookupLimit' configured, copy the value to 'TransactionHistory'.
 	if cfg.TransactionHistory == ethconfig.Defaults.TransactionHistory && cfg.TxLookupLimit != ethconfig.Defaults.TxLookupLimit {
-		log.Warn("The config option 'TxLookupLimit' is deprecated and will be removed, please use 'TransactionHistory'")
-		cfg.TransactionHistory = cfg.TxLookupLimit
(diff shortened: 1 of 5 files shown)


cmd/utils/flags.go
```diff
@@ -1954,8 +1954,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	if ctx.IsSet(TransactionHistoryFlag.Name) {
 		cfg.TransactionHistory = ctx.Uint64(TransactionHistoryFlag.Name)
 	} else if ctx.IsSet(TxLookupLimitFlag.Name) {
-		log.Warn("The flag --txlookuplimit is deprecated and will be removed, please use --history.transactions")
-		cfg.TransactionHistory = ctx.Uint64(TransactionHistoryFlag.Name)
+		Fatalf("The flag --txlookuplimit is deprecated and may cause unexpected performance degradation issues. Please use --history.transactions instead.")
```
NathanBSC, line 1957: use log.Crit?
if use `Fatalf`, we may lose the log
for example, run in a docker without a terminal

NathanBSC: should line 1951-1952 also be modified?
they handle the cases when TxLookupLimit is configured in config file

buddh0 (approved):

buddh0 (approved):

brilliant-lx (approved):

PR #2012, opened 2023-11-28
fix: support  db engine with pebble on 32 bits and OpenBSD 

flywukong:
### Description

https://github.com/cockroachdb/pebble/pull/2878 the pebble version that used by bsc has already supported 32-bit . So the judgment logic which disabled  the 32-bit environments needs to be removed
The default db is still leveldb.

### Rationale
support  db engine with pebble on 32 bits and OpenBSD too

### Example

### Changes

Notable changes: 
* ...


sysvm (approved):
LGTM

Fynnss (approved):
LGTM

Issue #31, opened 2020-08-27
Stake Wars Rewards Claim


huangsuyu:

tigdar:
{
  "address": "0x0af7d8b7d4eb50fa0eddd643d11120c94ad61248",
  "msg": "bnb149m2zlvnn223yk28wjjxwurlmu5nmeg7ew4us0",
  "sig": "0x956c73f435450df67a8d3c8719ebd0e061c3060bbba689d13df39bfa9b048a4d531374b407b722179ab56ab160dba3482ec68130629cc05a29cfd201c8cae0f01c",
  "version": "3",
  "signer": "MEW"
}

Metas-taz:
{
  "address": "0xbc93f0661a49822b66b07e0727848ed3598551e0",
  "msg": "bnb1rydsekq4283qfcudfkzevsge9u7kl7xfke68p7",
  "sig": "0x51df018cddf18fd011929b397e3208dc4a1e054dd836d8322093ca2ad526d5c0790dbc50a0936f10d6ba7aa4f13dad4dce0bdd26fe229c9f7e0323fed7c295811c",
  "version": "3",
  "signer": "MEW"
}

utkarsssh17:
{
  "address": "0xa2e5f9e8db4b38ac8529f79c4f3b582952b3d3dc",
  "msg": "bnb16nwsuwgqzwaxljufw7u4zt3vqs6l6fxjwrd9wh",
  "sig": "0xb6697e14bfc3cb4b2ff579e30042437691c10ab65e8ae386c5188e423d00b7180de59bb403c41ac3d21e0c34e2c6a001f14ecf263ca55dacb6ef71abd389bd971c",
  "version": "3",
  "signer": "MEW"
}

zzzTheWinner:
{
  "address": "0x3efb67b34a9b69b54d4027e044954f483dc31678",
  "msg": "bnb10sem3qt9hrm45cmdh9w6h7ulllp5u2m0ylqjkz",
  "sig": "0x3c7eb87bafa0e44b7c0d88d8159318146589ef5c452bd281a566e2c61a136a1e14eed74ec385f44c1e82d8922f2037b8f9f45e5a4ec277cfacb7e65ee02b5fdc1c",
  "version": "3",
  "signer": "MEW"
}

redredemptionmark:
{
  "address": "0x4083dcf5ba3d1506e3d43067cdbe262124adbf8f",
[...]

PR #2017, opened 2023-11-29
BEP-283: Segmented History Data Maintenance

0xbundler:
# BEP-283: Segmented History Block Maintenance

- [BEP-283: Segmented History Block Maintenance]((https://github.com/bnb-chain/BEPs/pull/283))
    - [1. Summary](#1-summary)
    - [2. Motivation](#2-motivation)
    - [3. Specification](#3-specification)
        - [3.1 Asset Reconciliation](#31-asset-reconciliation)
        - [3.2 Halt on Reconciliation Error](#32-halt-on-reconciliation-error)
    - [4. License](#4-license)

## 1. Summary
This BEP proposes a practical solution to address the issue of increasing history data storage on the BSC(BNB Smart Chain) for full nodes, so they only need to maintain a limited range of blocks. With the introduction of a new synchronization protocol and storage protocol, nodes can synchronize from a specific checkpoint and skip historical blocks. This eliminates the mandatory requirement to keep historical blocks.
## 2. Motivation

The history block data includes block header, body and receipt, they are useless to execute the latest blocks. A recent storage profile on Jul-02-2023 shows that the size of history block has reached ~1288 GB, which makes it a big burden to run a BSC full node. As BSC keeps generating new blocks, the history block data also keeps growing.
Actually the history block data can be maintained by some archive node or by a DA layer, a full node does not need to keep a full copy of the history block. But simply deleting the history block data is also not acceptable, it will make the P2P network in chaos and hard to sync.
We need to make a rule so that full nodes only need to keep a recent period of blocks, like several months, it would be better to keep the bounded history block data size within 200GB. It could reduce both node’s storage pressure and the network traffic cost.

## 3. Specification
|        Parameter       |   Value      |        Description        |
| ----------------- | ----------- | ------------------------  |
|       BoundStartBlock       |   31,268,530    | The starting block height of the first segment, block 31,268,530 was produced on Aug-29-2023, 3 years after the BSC genesis block.        |
|    HistorySegmentLength   |   2,592,000    | assume 1 block for every 3 second, 2,592,000 blocks will be produced in 90 days. |

### 3.1 General Workflow

The history block data will be divided into several segments, the 1st segment is segment_0, which is from genesis to BoundStartBlock-1, then following segments will have the same length: HistorySegmentLength.
```
HistorySegment_0 = [0, BoundStartBlock)
HistorySegment_1 = [BoundStartBlock, BoundStartBlock+HistorySegmentLength)
...
HistorySegment_N = [BoundStartBlock+(N-1)*HistorySegmentLength, BoundStartBlock+N*HistorySegmentLength)
```
The BSC node only needs to maintain the latest 2 segments, in case of block reorg, the current segment number is calculated based on the “finalized” block.
```
ImmutabilityThreshold = 90000

func GetFinalizedBlock() {
  blockByAttestation := GetFinalizedByAttestation()
  blockByThreshold := GetCurrentBlockNumber() - ImmutabilityThreshold
  return max(blockByAttestation, blockByThreshold)
}

func GetCurrentSegmentIndex(blockNum int) {
  if blockNum < BoundStartBlock {
    return 0
  }
  boundBlocks = blockNum - BoundStartBlock
  return (boundBlocks/HistorySegmentLength) + 1
}

finalizedBlock = GetFinalizedBlock()
segIndex = GetCurrentSegmentIndex(finalizedBlock)
if  segIndex > = 2 {
  // the segments before (segIndex - 2) can be pruned
}

```

### 3.2 Prune

The offline block prune would need to be boundary aligned and leave the most recent 2 segments. The most recent segment is determined by the finalized block, since FastFinality is to be enabled, we can take use of the feature to determine the finalized block and then get the current segment index.

#### 3.2.1 Prune details
Since we introduced segment definition, the historical block data will be organized on a segment-by-segment basis within the file directory, making it easy to prune the block data by deleting the corresponding file folder.
```
if segIndex > 2 {
  // delete the file folder of segIndex0 to segIndex - 2
  // os.RemoveAll(segIndex0..segIndex-2)
}

```
### 3.3 Node Sync
 It would be difficult to sync from genesis, since most nodes may choose to not preserve these old blocks, but it is still possible as long as some nodes still keep the whole history data.
There could have 2 approaches to do sync after this proposal
directly download a snapshot of a recent state, from a snapshot service provider or DA layer like GreenField
Segment based snap sync: user can take the boundary blockhash as the new `GenesisHash` and start from it directly. 

### 3.4 Data Availability
Some of the nodes like archive nodes would keep maintaining the whole history data.
And meanwhile, could take use of a DA layer, like greenfield to make sure the whole block data is available.

### 4.1 BoundStartBlock & HistorySegmentLength
As the current history block data is already very large, we prefer to enable this proposal faster, so
BoundStartBlock is still more than 1 month ahead, it could be an acceptable date.
HistorySegmentLength, we did a profile, for the past 6 months(Jan-2023 to July 2023), ~1.2GB history data was generated per day on average. But it is somehow low traffic during this period. Traffic volume could be 3 times if the bull market starts, that is ~3.6GB per day. To keep the historical data size within 500GB in the bull market and 200GB in the bear market, 90 days could be a reasonable value.

### 4.2 Why Segment The History Data
No Big fundamental differences to the EIP-4444 proposal
Pros: aligned
easier to determine the current living blocks?
easier to provide the snapshot download service?

### 5.1 Portal Network
Portal network is a hot topic to solve the storage pressure, once it is ready, it is possible that portal network can replace this proposal if it has a more applicable solution.

### 6.1 Archive Node
If you run an archive node, you can just keep all the history block data, no impact to its business.
### 6.2 RPC API
If users query history block data that have been pruned, could return a new error code to show it is expired and removed.

## License
The content is licensed under the [CC0](https://creativecommons.org/publicdomain/zero/1.0/) license.


cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,205 @@ func dump(ctx *cli.Context) error {
...
+
+	genesisHash := rawdb.ReadCanonicalHash(db, 0)
+	td := rawdb.ReadTd(db, genesisHash, 0)
+	chainConfig, engine, headerChain, err := simpleHeaderChain(db, genesisHash)
+	if err != nil {
+		return err
+	}
+	latest := headerChain.CurrentHeader()
+	if _, ok := engine.(consensus.PoSA); !ok {
+		return errors.New("current chain is not POSA consensus, cannot generate history segment")
```
Mister-EA, line 729: Rephrase: "cannot generate history segment because consensus engine is not PoSA"

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,205 @@ func dump(ctx *cli.Context) error {
...
+	chainConfig := rawdb.ReadChainConfig(db, genesisHash)
+	if chainConfig == nil {
+		return nil, nil, nil, errors.New("failed to load chainConfig")
+	}
+	engine, err := ethconfig.CreateConsensusEngine(chainConfig, db, nil, genesisHash)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	if _, ok := engine.(consensus.PoSA); !ok {
+		return nil, nil, nil, errors.New("current chain is not POSA, cannot generate history segment")
```
Mister-EA, line 906: "current engine is not PoSA ... "

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+			segments[i].ReGenesisNumber <= segments[i-1].ReGenesisNumber {
+			return fmt.Errorf("wrong segement, index: %v, segment: %v", i, segments[i])
+		}
+	}
+
+	return nil
+}
+
+// HisSegments return all history segments
+func (m *HistorySegmentManager) HisSegments() []HistorySegment {
```
Mister-EA, line 206: Rename to `HistoricalSegments` for clarity .

eth/ethconfig/gen_config.go
```diff
@@ -17,59 +17,62 @@ import (
...
+		BlobPool                 blobpool.Config
+		GPO                      gasprice.Config
+		EnablePreimageRecording  bool
+		DocRoot                  string `toml:"-"`
+		RPCGasCap                uint64
+		RPCEVMTimeout            time.Duration
+		RPCTxFeeCap              float64
+		OverrideCancun           *uint64 `toml:",omitempty"`
+		OverrideVerkle           *uint64 `toml:",omitempty"`
+		HistorySegmentEnable     bool    `toml:",omitempty"`
```
Mister-EA, line 74: Rename to either `HistorySegmentEnabled` or `EnableHistorySegment` .

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+	return HistoryBlock{
+		Number: num,
+		hash:   hash,
+		TD:     td,
+	}
+}
+
+type HistorySegment struct {
+	Index           uint64      `json:"index"`                    // segment index number
+	ReGenesisNumber uint64      `json:"re_genesis_number"`        // new history segment start at a finality block number, called ReGenesisNumber
```
Mister-EA, line 76: Maybe it would be a good idea not to confuse this with the genesis block, and rather call this `StartingBlockNumber` . And the field below `StartingBlockHash` .

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+	ReGenesisHash   common.Hash `json:"re_genesis_hash"`          // new history segment start at a finality block hash, called ReGenesisHash
+	TD              uint64      `json:"td"`                       // the ReGenesisBlock's TD
+	ConsensusData   string      `json:"consensus_data,omitempty"` // the ReGenesisBlock's consensus data
+}
+
+func (s *HistorySegment) String() string {
+	return fmt.Sprintf("{Index: %v, ReGenesisNumber: %v, ReGenesisHash: %v, TD: %v}", s.Index, s.ReGenesisNumber, s.ReGenesisHash, s.TD)
+}
+
+func (s *HistorySegment) MatchBlock(h common.Hash, n uint64) bool {
```
Mister-EA, line 86: Rename to `MatchesBlockAtHeight`

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	return &HistorySegmentManager{
+		segments: segments,
+		cfg: &HistorySegmentConfig{
+			Genesis: genesis,
+		},
+	}, nil
+}
+
+func ValidateHisSegments(genesis HistoryBlock, segments []HistorySegment) error {
```
Mister-EA, line 182: Rename to `ValidateHistorySegments`

eth/ethconfig/config.go
```diff
@@ -197,6 +197,10 @@ type Config struct {
 
 	// OverrideVerkle (TODO: remove after the fork)
 	OverrideVerkle *uint64 `toml:",omitempty"`
+
+	// History Segment
+	HistorySegmentEnable     bool   `toml:",omitempty"`
```
Mister-EA, line 202: Rename to either `HistorySegmentEnabled` or `EnabledHistorySegment`

eth/downloader/downloader.go
```diff
@@ -1726,3 +1760,37 @@ func (d *Downloader) reportSnapSyncProgress(force bool) {
...
+	}
+	// Make sure the peer actually gave something valid
+	if len(headers) != 1 {
+		return 0, fmt.Errorf("%w: multiple headers (%d) for single request", errBadPeer, len(headers))
+	}
+
+	// check if it matches local last segment
+	h := hashes[0]
+	n := headers[0].Number.Uint64()
+	if d.lastSegment.MatchBlock(h, n) {
```
Mister-EA, line 1785: You can combine these if statements in one statement using `||` .

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,205 @@ func dump(ctx *cli.Context) error {
...
+		return errors.New("current chain is not enable Luban hard fork, cannot generate history segment")
+	}
+
+	var (
+		boundStartBlock      = params.BoundStartBlock
+		historySegmentLength = params.HistorySegmentLength
+	)
+	switch genesisHash {
+	case params.BSCGenesisHash, params.ChapelGenesisHash, params.RialtoGenesisHash:
+		boundStartBlock = params.BoundStartBlock
```
Mister-EA, line 741: Not possible to configure BoundStartBlock based on the flag value on mainnet?

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
...1396466346132222c223331323638353834223a226231396466346132222c223331323638353835223a226231396466346132222c223331323638353836223a226231396466346132222c223331323638353837223a226231396466346132222c223331323638353838223a226231396466346132222c223331323638353839223a226231396466346132222c223331323638353930223a226231396466346132222c223331323638353931223a226231396466346132222c223331323638353932223a226231396466346132222c223331323638353933223a226231396466346132222c223331323638353934223a226231396466346132222c223331323638353935223a226231396466346132222c223331323638353936223a226231396466346132222c223331323638353937223a226231396466346132222c223331323638353938223a226231396466346132222c223331323638353939223a226231396466346132227d2c226174746573746174696f6e3a6f6d6974656d707479223a7b22536f757263654e756d626572223a33313236383539372c22536f7572636548617368223a22307863653762343062643836353463306339653961653131373063393839666531353138323562356235306163313063313838326539383331366164643362343833222c225461726765744e756d626572223a33313236383539382c2254617267657448617368223a22307837323335343633626530633139623361626461366666653766393035633936613861643632616565393663616239396465613937303531626433346564313366227d7d"
```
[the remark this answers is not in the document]
0xbundler: Hard code is key for history segment feature, nodes could follow the history prune rule as same, it's also a simplest way.  And user can use custom history segment config in private chain.

params/history_segment.go
0xbundler: It's better to use re-genesis, the BEP emphasize that it a new finality range like genesis, so called "ReGenesis".

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+		TD:     td,
+	}
+}
+
+type HistorySegment struct {
+	Index           uint64      `json:"index"`                    // segment index number
+	ReGenesisNumber uint64      `json:"re_genesis_number"`        // new history segment start at a finality block number, called ReGenesisNumber
+	ReGenesisHash   common.Hash `json:"re_genesis_hash"`          // new history segment start at a finality block hash, called ReGenesisHash
+	TD              uint64      `json:"td"`                       // the ReGenesisBlock's TD
+	ConsensusData   string      `json:"consensus_data,omitempty"` // the ReGenesisBlock's consensus data
```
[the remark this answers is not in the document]
0xbundler: No, it's a parlia's snapshot of target block for verifying header or attestation.

params/history_segment.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if !strings.EqualFold(s.ConsensusData, compared.ConsensusData) {
+		return false
+	}
+	return true
+}
+
+type HistorySegmentConfig struct {
+	CustomPath string       // custom HistorySegments file path, need read from the file
+	Genesis    HistoryBlock // specific chain genesis, it may use hard-code config
```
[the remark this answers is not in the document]
0xbundler: This field is the real chain genesis, it's used to init history segment manager.

cmd/geth/chaincmd.go
0xbundler: yes, mainnet or testnet is hard code, every node should follow the same rules.

params/history_segment.go
Mister-EA: What does ReGenesis mean ?

params/history_segment.go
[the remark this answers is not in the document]
Mister-EA: Can't we use the actual struct instead of the serialized string?

params/history_segment.go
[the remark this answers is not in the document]
Mister-EA: Can't we make it so that these bytes are read from a file at least?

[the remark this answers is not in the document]
0xbundler: If these data read from file, it will cause more maintenance resource. I think hard code is ok, like `core/systemcontracts/upgrade.go` does.

params/history_segment.go
0xbundler: It provides a finality block to start sync/store block from there, like genesis.

params/history_segment.go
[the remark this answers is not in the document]
0xbundler: It aims to contain a data, not target type, so it could be extend other consensus later, like clique, ethash, etc.

params/history_segment.go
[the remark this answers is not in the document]
Mister-EA: The pattern in `upgrade.go` is not great either, because it crashes the IDE loading all the data.  If you can't think of any better way, at least one thing we can do is keep this data in a separate .go file so that the struct definitions below are easier to locate.

cmd/geth/chaincmd.go
zzzckck: mainnet, chapel and rialto should have different `BoundStartBlock`, right?

cmd/geth/dbcmd.go
```diff
@@ -257,7 +257,7 @@ WARNING: This is a low-level operation which may cause database corruption!`,
 	}
 	ancientInspectCmd = &cli.Command{
 		Action: ancientInspect,
-		Name:   "inspect-reserved-oldest-blocks",
+		Name:   "inspect-ancient",
```
zzzckck, line 260: why change the name?

eth/backend.go
```diff
@@ -259,7 +274,23 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 
 	peers := newPeerSet()
 	bcOps = append(bcOps, core.EnableBlockValidator(chainConfig, eth.engine, config.TriesVerifyMode, peers))
-	eth.blockchain, err = core.NewBlockChain(chainDb, cacheConfig, config.Genesis, &overrides, eth.engine, vmConfig, eth.shouldPreserve, &config.TransactionHistory, bcOps...)
+	txLookupLimit := &config.TransactionHistory
+	// if enable HistorySegment, just skip txLookupLimit params,
+	// may cause regenerate tx index, but it will also generate new block index
+	if config.HistorySegmentEnabled {
+		txLookupLimit = nil
```
zzzckck, line 281: why disable transaction index generation when this BEP is enabled?

params/history_segment_config.go
```diff
@@ -0,0 +1,56 @@
+package params
+
+const (
+	BoundStartBlock      uint64 = 31268600 // The starting block height of the first segment, was produced on Aug-29-2023
+	HistorySegmentLength uint64 = 2592000  // Assume 1 block for every 3 second, 2,592,000 blocks will be produced in 90 days.
```
zzzckck, line 5: https://docs.google.com/document/d/1txmWxATz-VnZpH63SWAEsBVrEr6HMtNZtsK3ZtaV0mQ/edit
history block data has increased ~500GB in 2023.
After a second though, maybe we can set it to 8months or even 1year, then segment change will be much slower.

zzzckck:
could you provide some manual on how to use this feature?

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,202 @@ func dump(ctx *cli.Context) error {
...
+		{
+			Index:           0,
+			ReGenesisNumber: 0,
+			ReGenesisHash:   genesisHash,
+			TD:              td.Uint64(),
+		},
+	}
+	// try find finalized block in every segment boundary
+	for num := boundStartBlock; num <= target; num += historySegmentLength {
+		// align the segment start at parlia's epoch
```
zzzckck, line 774: why epoch aligned?

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,202 @@ func dump(ctx *cli.Context) error {
...
+	td := rawdb.ReadTd(db, genesisHash, 0)
+	chainConfig, engine, headerChain, err := simpleHeaderChain(db, genesisHash)
+	if err != nil {
+		return err
+	}
+	latest := headerChain.CurrentHeader()
+	if _, ok := engine.(consensus.PoSA); !ok {
+		return errors.New("cannot generate history segment because consensus engine is not PoSA")
+	}
+	if !chainConfig.IsLuban(latest.Number) {
```
buddh0, line 731: IsPlato？
from plato, GetFinalizedHeader works

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,202 @@ func dump(ctx *cli.Context) error {
...
+		return errors.New("luban hard fork is not enabled , cannot generate history segment")
+	}
+
+	var (
+		boundStartBlock      = params.BoundStartBlock
+		historySegmentLength = params.HistorySegmentLength
+	)
+	switch genesisHash {
+	case params.BSCGenesisHash, params.ChapelGenesisHash, params.RialtoGenesisHash:
```
buddh0, line 740: to test locally, should support flag RialtoHash
https://github.com/bnb-chain/bsc/commit/a6befb5078425c148eb74943586e4b438264af78

core/headerchain.go
```diff
@@ -134,6 +134,17 @@ func (hc *HeaderChain) getFinalizedNumber(header *types.Header) uint64 {
 	return 0
 }
 
+// GetFinalizedHeader returns the highest finalized header before the specific block.
+func (hc *HeaderChain) GetFinalizedHeader(header *types.Header) *types.Header {
+	if p, ok := hc.engine.(consensus.PoSA); ok {
+		if finalizedHeader := p.GetFinalizedHeader(hc, header); finalizedHeader != nil {
```
buddh0, line 140: just `return p.GetFinalizedHeader(hc, header); `?

cmd/geth/chaincmd.go
```diff
@@ -675,8 +711,202 @@ func dump(ctx *cli.Context) error {
...
+			// if there cannot found any finalized block, just skip
+			break
+		}
+		if ft.Number.Uint64() < num {
+			// cannot find expect finality blocks, just skip
+			break
+		}
+		log.Info("found segment boundary", "startAt", ft.Number, "FinalityAt", fs.Number)
+		segment := params.HistorySegment{
+			Index:           uint64(len(segments)),
```
buddh0, line 799: uint64(len(segments))+1?

consensus/parlia/parlia.go
```diff
@@ -1512,6 +1555,20 @@ func (p *Parlia) Close() error {
 	return nil
 }
 
+// GetConsensusData load the header's last snapshot for history segment
+func (p *Parlia) GetConsensusData(chain consensus.ChainHeaderReader, header *types.Header) ([]byte, error) {
+	number := header.Number.Uint64()
+	snap, err := p.snapshot(chain, number-1, header.ParentHash, nil)
```
NathanBSC, line 1561: why exclude current header？
from interface name, It's try to get GetConsensusData from chain with header as latest block

consensus/parlia/parlia.go
```diff
@@ -751,6 +764,36 @@ func (p *Parlia) snapshot(chain consensus.ChainHeaderReader, number uint64, hash
 	return snap, err
 }
 
+func (p *Parlia) findSnapFromHistorySegment(hash common.Hash, number uint64) (*Snapshot, bool) {
+	if p.hsm == nil {
+		return nil, false
+	}
+	segment := p.hsm.CurSegment(number + 1)
```
NathanBSC, line 771: number+1? related GetConsensusData

eth/downloader/downloader.go
```diff
@@ -573,6 +595,13 @@ func (d *Downloader) syncWithPeer(p *peerConnection, hash common.Hash, td, ttd *
 				return err
 			}
 		}
+
+		// if enable history segment, force reset freezer tail
+		if d.prevSegment != nil && localHeight == 0 {
+			if err := d.blockchain.FreezerDBReset(origin, origin); err != nil {
```
buddh0, line 601: this function only called here, and tail == head,

so much code on path ` tail !=head` will never be called?

if so, we can just simplify the logic

0xbundler:
Below is the quick user guide of BEP-283 pr.
## Run Node

### Mainnet/Testnet

If you run a mainnet or testnet node, you just run your node with a new flag `--history-segment` . It will use hard code history segment to auto prune useless ancient data(except tx index) when node restart.

```bash
./bsc --config "./node/config.toml" --rpc.allow-unprotected-txs --allow-insecure-unlock --gcmode full --syncmode snap --ws --datadir "./node/data-seed" --metrics --metrics.addr 0.0.0.0 --metrics.port 6062 --pprof --pprof.port 6072 --http.corsdomain "*" --maxpeers 500 --history.transactions 0 --metrics.expensive --history-segment
```

Default segment config as below:

| Parameter | Value | Description |
| ---| ---| --- |
| BoundStartBlock | 31,268,600 | The starting block height of the first segment. |
| HistorySegmentLength | 2,592,000 | assume 1 block for every 3 second, 2,592,000 blocks will be produced in 90 days. |

There are also provide a offline prune tool, it will prune tx index too.

```bash
./bsc prune-history-segments --datadir {dataDir}
```

It is also compatible with `pruneancient` & `prune-block` feature.

#### Snap Sync

If you start snap sync from genesis, the node will only download latest 2 segments blocks.

image

### Private Network

If you run a private network, you could export & import your custom history segment file.

*   export history segment config, specific segment `boundstart` block, and `segmentlen` .

```bash
./bsc export-segment --datadir {dataDir} --history-segment.output ./history_segments.json --history-segment.boundstart 31268530 --history-segment.segmentlen 1000000
```

*   run node with custom history segment, with flag `--history-segment --history-segment.custom ./history_segments.json` .
*   prune ancient data.

```bash
./bsc prune-history-segments --datadir {dataDir} --history-segment.custom ./history_segments.json
```

> Attention: history segment need `Luban` hardfork.

## Storage Profile

| Type | DIsk Usage | Storage Size |
| ---| ---| --- |
| Normal Node | 2.2 T | 2.41 T |
| Enable History Segment Prune<br> | 837 G, \-63% | 1001.25 G, \-60% |

consensus/parlia/parlia.go
0xbundler: Yes, the segment's consensus data is the previous block snapshot, so it can provide segment start block validation in parlia.

cmd/geth/chaincmd.go
0xbundler: When parlia verifies the header, it needs a snapshot produced from the epoch block header. So it is better to align the segment starting at parlia's epoch, the parlia always could find that info.

eth/backend.go
0xbundler: It is a very simple way to handle the tx index, it prevents the indexer from regenerating ancient tx index, and it may cause error when ancient data has been removed.

melissamforbs (approved):

allformless:
the design is outdate,
upstream go-ethereum has support flag `--history.chain`

PR #2050, opened 2023-12-07
Introduce s2 compression for p2p

MatusKysel:
### Description

This pull request introduces the S2 compression algorithm to the BSC node. The implementation integrates the S2 algorithm as an additional compression option alongside the existing methods. This enhancement aims to improve data compression efficiency, potentially reducing storage requirements and increasing data transmission speeds for BSC node operations.

### Rationale

The current BSC node utilizes traditional compression techniques, which, while effective, can be further optimized. The S2 compression algorithm is renowned for its high compression ratio and speed, especially in scenarios involving large datasets. By integrating S2, we aim to:

- Enhance Compression Efficiency: Achieve better compression ratios, which can lead to significant storage savings.
- Increase Speed: Improve the speed of data processing and transmission, crucial for high-throughput blockchain environments.
- Maintain Compatibility: Ensure backward compatibility with existing nodes, allowing for a smooth transition and adoption.

### Documentation

S2 vs Snappy performance https://github.com/klauspost/compress/tree/master/s2#performance 

### Changes

Notable changes: 
* S2 Algorithm Integration: Added the S2 compression algorithm as a part of the BSC node's data processing module.

### DIFF
--- a/p2p/peer.go
+++ b/p2p/peer.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/p2p/enr"
+	"github.com/ethereum/go-ethereum/p2p/rlpx"
 	"github.com/ethereum/go-ethereum/rlp"
 	"golang.org/x/exp/slices"
 )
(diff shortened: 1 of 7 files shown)


weiihann:
Overall looks good to me. It'll be even better if we can have some benchmarking comparisons between Snappy and S2 in Geth itself, since the comparison results might differ due to the difference in data format.

p2p/transport.go
```diff
@@ -181,3 +181,10 @@ func readProtocolHandshake(rw MsgReader) (*protoHandshake, error) {
 	}
 	return &hs, nil
 }
+
+func Min(a, b uint64) uint64 {
```
Mister-EA, line 185: Don't think there is a need to export this.

p2p/peer.go
```diff
@@ -39,12 +40,10 @@ var (
 )
 
 const (
-	baseProtocolVersion    = 5
+	baseProtocolVersion    = rlpx.S2ProtocolVersion
```
Mister-EA, line 43: Can this be configured via a flag?

PR #2067, opened 2023-12-14
cmd/geth: add check func to validate state scheme

sysvm:
### Description

Provide explicit for which state scheme geth can use to start or init. And add check function to validate provided state scheme.

### Rationale

Users may add `StateScheme = "path"` in `config.toml` and want to start geth without persistent state in disk, but geth is started in `hash` mode. For solving these cases, add a function, obey the following rules:

1. If config isn't provided, write hash mode to config by default, so in current function, config is nonempty.
2. If persistent state and cli is empty, use config param.
3. If persistent state is empty, provide CLI flag and config, choose CLI to return.
4. If persistent state is nonempty and CLI isn't provided, persistent state should be equal to config.
5. If all three items are provided: if any two of the three are not equal, return error.

### Example

1. `nohup ./geth --config config.toml --cache 8000 --rpc.allow-unprotected-txs --history.transactions 0 >> nohup.out 2>&1&` and add `StateScheme = "path"` in `config.toml`, geth is started in `path` mode.

### Changes

Notable changes: 
* N/A

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -51,7 +51,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/kzg4844"
 	"github.com/ethereum/go-ethereum/eth"
-
 	"github.com/ethereum/go-ethereum/eth/downloader"
 	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/eth/filters"
@@ -1941,7 +1940,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	}
 	// Parse state scheme, abort the process if it's not compatible.
 	chaindb := tryMakeReadOnlyDatabase(ctx, stack)
-	scheme, err := ParseStateScheme(ctx, chaindb)
+	scheme, err := ResolveStateScheme(ctx, cfg.StateScheme, chaindb)
 	chaindb.Close()
 	if err != nil {
 		Fatalf("%v", err)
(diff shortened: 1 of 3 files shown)


joeylichang (approved):

joeylichang:
LGTM

cmd/utils/flags.go
```diff
@@ -2487,6 +2486,49 @@ func MakeConsolePreloads(ctx *cli.Context) []string {
...
+// 3. If persistent state is empty, provide CLI flag and config, choose CLI to return.
+// 4. If persistent state is nonempty and CLI isn't provided, persistent state should be equal to config.
+// 5. If all three items are provided: if any two of the three are not equal, return error.
+func ResolveStateScheme(ctx *cli.Context, stateSchemeCfg string, disk ethdb.Database) (string, error) {
+	stored := rawdb.ReadStateScheme(disk)
+	if stored == "" {
+		// there is no persistent state data in disk db(e.g. geth init)
+		if !ctx.IsSet(StateSchemeFlag.Name) {
+			log.Info("State scheme set by config", "scheme", stateSchemeCfg)
+			return stateSchemeCfg, nil
```
flywukong, line 2502: need judge  if the stateSchemeCfg is not setted

cmd/utils/flags.go
```diff
@@ -2487,6 +2486,49 @@ func MakeConsolePreloads(ctx *cli.Context) []string {
...
+		// if both CLI flag and config are set, choose CLI
+		scheme := ctx.String(StateSchemeFlag.Name)
+		if !ValidateStateScheme(scheme) {
+			return "", fmt.Errorf("invalid state scheme param in CLI: %s", scheme)
+		}
+		log.Info("State scheme set by CLI", "scheme", scheme)
+		return scheme, nil
+	}
+	if !ctx.IsSet(StateSchemeFlag.Name) {
+		if stored != stateSchemeCfg {
```
flywukong, line 2513: should not return error  if stateSchemeCfg  is empty

cmd/utils/flags.go
```diff
@@ -2487,6 +2486,49 @@ func MakeConsolePreloads(ctx *cli.Context) []string {
...
+			return "", fmt.Errorf("incompatible state scheme, stored: %s, config: %s", stored, stateSchemeCfg)
```
flywukong, line 2514: should not return error if stateSchemeCfg is empty

cmd/utils/flags.go
sysvm: `stateSchemeCfg` cannot be empty in `ResolveStateScheme` function, because there is default value in caller function `loadBaseConfig`.

cmd/utils/flags.go
sysvm: ditto. `stateSchemeCfg` cannot be empty in `ResolveStateScheme` function.

flywukong (dismissed):

RenRick (approved):
LGTM

flywukong (approved):
LGTM

PR #2070, opened 2023-12-15
cmd, eth: add --disablebscprotocol flag

weiihann:
### Description
This PR adds a new flag to disable the BSC subprotocol.

### Rationale
When setting up a local private network, users may want to disable the BSC subprotocol for development purposes.

### Example
Usage:
```
./geth --disablebscprotocol
```

### Changes
- new `--disablebscprotocol` flag

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -111,6 +111,11 @@ var (
 		Usage:    "Enable trust protocol",
 		Category: flags.FastNodeCategory,
 	}
+	DisableBscProtocolFlag = &cli.BoolFlag{
+		Name:     "disablebscprotocol",
+		Usage:    "Disable bsc protocol",
+		Category: flags.FastFinalityCategory,
+	}
 	PipeCommitFlag = &cli.BoolFlag{
 		Name:     "pipecommit",
 		Usage:    "Enable MPT pipeline commit, it will improve syncing performance. It is an experimental feature(default is false)",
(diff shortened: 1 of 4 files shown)


buddh0:
if all middle nodes disable bscprotocol, votes will be not reachable

as suggested, I remove flat `disablebscprotocol` previously:
https://github.com/bnb-chain/bsc/pull/1805/files

eth/ethconfig/config.go
```diff
@@ -115,6 +115,7 @@ type Config struct {
 	DirectBroadcast     bool
 	DisableSnapProtocol bool //Whether disable snap protocol
 	EnableTrustProtocol bool //Whether enable trust protocol
+	DisableBscProtocol  bool //Whether disable bsc protocol
```
buddh0, line 118: when  define of `Config` changed, gen_config.go should be regenerated

image

cmd/utils/flags.go
```diff
@@ -1920,6 +1925,12 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 	if ctx.IsSet(EnableTrustProtocolFlag.Name) {
 		cfg.EnableTrustProtocol = ctx.IsSet(EnableTrustProtocolFlag.Name)
 	}
+	if ctx.IsSet(DisableBscProtocolFlag.Name) {
+		if ctx.Bool(MiningEnabledFlag.Name) {
```
buddh0, line 1929: check `cfg.DisableBscProtocol` is better after DisableBscProtocol is assigned
otherwise if `--disablebscprotocol false`
here will result wrong behavior

zzzckck:
hold the PR for now, as it may have some side effects, could break the FastFinality feature if many nodes use this flag improperly. And the benefit could be limitted.

PR #2074, opened 2023-12-15
new faucet client

MatusKysel:
### Description

In this update to the faucet client, the default functionality of the faucet binary has been altered. Previously, the faucet binary utilized the Light Ethereum Subprotocol (LES) for connecting to the BSC network. However, this update changes this behavior, making the faucet connect to an already running BSC node via WebSocket (WS) protocol. This modification is aimed to support faucet after LES deprecation. 

### Rationale

The motivation behind these changes includes:

- **Resource Efficiency**: By connecting to an existing node, we reduce the resource load on the system running the faucet, as it no longer needs to maintain a full LES client.
- **Enhanced Flexibility**: This approach allows for greater flexibility in network setups, as the faucet can connect to any accessible node on the network, not just those running LES.

### Example

After implementing these changes, the faucet can be connected to a running node using a command similar to:

```
 faucet --ws <node_ws_endpoint>
```

### Changes

Notable changes in this update include:

- **Faucet Binary Modification**: Altered the default connection protocol of the faucet binary from LES to WebSocket.
- **New Command-Line Options**: Added new CLI options to specify the WebSocket endpoint of the BSC node.

### DIFF
--- a/cmd/faucet/faucet.go
+++ b/cmd/faucet/faucet.go
@@ -335,7 +335,11 @@ func (f *faucet) apiHandler(w http.ResponseWriter, r *http.Request) {
 	f.lock.RLock()
 	reqs := f.reqs
 	f.lock.RUnlock()
-	peers, _ := f.client.PeerCount(context.Background())
+	ctx, _ := context.WithTimeout(context.Background(), time.Second)
+	peers, err := f.client.PeerCount(ctx)
+	if err != nil {
+		log.Warn("Failed to get peerCount on init", "err", err)
+	}
 	if err = send(wsconn, map[string]interface{}{
 		"funds":    new(big.Int).Div(balance, ether),
 		"funded":   nonce,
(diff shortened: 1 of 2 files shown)


weiihann (dismissed):
LGTM

zzzckck (approved):

cmd/faucet/faucet.go
```diff
@@ -248,65 +227,18 @@ type wsConn struct {
...
-			DiscoveryV5:      true,
-			ListenAddr:       fmt.Sprintf(":%d", port),
-			MaxPeers:         25,
-			BootstrapNodesV5: enodes,
-		},
-	})
-	if err != nil {
-		return nil, err
-	}
+func newFaucet(genesis *core.Genesis, url *string, ks *keystore.KeyStore, index []byte, bep2eInfos map[string]bep2eInfo) (*faucet, error) {
```
bnb-alexlucaci, line 230: why does the url needs to be a string pointer instead of string?

zzzckck (approved):
LGTM

weiihann (approved):

PR #2083, opened 2023-12-18
triedb/pathdb: fix async node buffer diskroot mismatches when journaling

sysvm:
### Description

When running `TestJournal` UT in `trie/triedb/pathdb`, there is a certain probability that this error will occur: `Invalid state, err: state 0xdda0214f6c0c6ea546cdbf3db74a31f8337a01b41908b3b8ea27a8fde0011294 is not available`. 

The reason of this error is that the journal loaded from disk is wrong, the error message of `loadJournal` function: `unmatched journal want c3bd21026750db0009399463f5947e1663c7057cb4aaabeb0b39ed68dedd4d0b got 6d574af1033459ae676316a1262a5d89edb7bbd515ecd253b4319a3161093153`.

### Rationale

When geth calls `Journal` function to commit entire diff hierarchy to disk into a single journal entry, the flush process of `asyncnodebuffer` maybe doesn't end, so we should stop, wait it and journal.

### Example

N/A

### Changes

Notable changes: 
* N/A

### DIFF
--- a/trie/triedb/pathdb/asyncnodebuffer.go
+++ b/trie/triedb/pathdb/asyncnodebuffer.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/VictoriaMetrics/fastcache"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -18,14 +19,14 @@ import (
 
 var _ trienodebuffer = &asyncnodebuffer{}
 
-// asyncnodebuffer implement trienodebuffer interface, and aysnc the nodecache
+// asyncnodebuffer implement trienodebuffer interface, and async the nodecache
 // to disk.
 type asyncnodebuffer struct {
 	mux          sync.RWMutex
 	current      *nodecache
 	background   *nodecache
-	stopFlushing uint64
-	flushing     uint64
+	isFlushing   atomic.Bool
+	stopFlushing atomic.Bool
 }
 
 // newAsyncNodeBuffer initializes the async node buffer with the provided nodes.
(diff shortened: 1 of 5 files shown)


trie/triedb/pathdb/asyncnodebuffer.go
```diff
@@ -18,12 +18,13 @@ import (
...
 var _ trienodebuffer = &asyncnodebuffer{}
 
-// asyncnodebuffer implement trienodebuffer interface, and aysnc the nodecache
+// asyncnodebuffer implement trienodebuffer interface, and async the nodecache
 // to disk.
 type asyncnodebuffer struct {
 	mux        sync.RWMutex
 	current    *nodecache
 	background *nodecache
+	isFlushing atomic.Bool
```
joeylichang, line 27: I think need a `stop` flag to stop flush when call waitAndStopFlush

sysvm: yes, fixed

trie/triedb/pathdb/asyncnodebuffer.go
```diff
@@ -149,26 +148,36 @@ func (a *asyncnodebuffer) flush(db ethdb.KeyValueStore, clean *fastcache.Cache,
...
 		}
 	}(id)
 	return nil
 }
 
+func (a *asyncnodebuffer) waitAndStopFlushing() {
+	a.stopFlushing.Store(true)
+	if a.isFlushing.Load() {
+		time.Sleep(time.Second)
+		log.Info("waiting background memory table flushed into disk")
```
joeylichang, line 170: log.Warn is better？

sysvm: fixed

joeylichang (approved):

buddh0 (approved):

PR #2107, opened 2023-12-25
cmd/geth: fix parse state scheme

sysvm:
### Description

Provide explicit for which state scheme geth can use to start or init. And add check function to validate provided state scheme.

### Rationale

Users can use `CLI` or `config.toml` to specify using which mode to start. And if you use snapshot(PBSS or HBSS) to start, there is meta info in db. Therefore, there are three items containing state scheme: `diskdb`, `config.toml`, `CLI`. Now valid state schemes are `hash` and `path`. Here are the rules:

1. There are no default value in `config.toml` and `CLI`, so if you don't set, it's empty string.
2. If users provide any two config items, they should be equal.
3. If all of three are not provided, geth use `hash` to run. If all of three are provided, they all should be equal.

### Example

N/A

### Changes

Notable changes: 
* N/A


cmd/utils/flags.go
```diff
@@ -2485,96 +2485,13 @@ func MakeConsolePreloads(ctx *cli.Context) []string {
...
-	if stored == "" || scheme == stored {
-		log.Info("State scheme set by user", "scheme", scheme)
-		return scheme, nil
-	}
-	return "", fmt.Errorf("incompatible state scheme, stored: %s, provided: %s", stored, scheme)
-}
-
 // MakeTrieDatabase constructs a trie database based on the configured scheme.
 func MakeTrieDatabase(ctx *cli.Context, disk ethdb.Database, preimage bool, readOnly bool) *trie.Database {
+	log.Info("enter MakeTrieDatabase")
```
flywukong, line 2490: remove the log

sysvm: fixed

flywukong (approved):

RenRick (approved):
LGTM

Fynnss:
LGTM

Fynnss (approved):
LGTM

Issue #2091, opened 2023-12-19
v1.3.7,Fatal: Could not open database: MemTableSize (4.0GB) must be < 4.0GB


niocm:
#### System information

Geth
Version: 1.3.7
Git Commit: f28b98a9947b52816714f90326b2545b08606b1a
Git Commit Date: 20231219
Architecture: amd64
Go Version: go1.20.2
Operating System: linux

#### Expected behaviour
Running below command：
`geth --config ./config.toml --datadir ./node  --cache 51200 --rpc.allow-unprotected-txs --history.transactions=0 --syncmode=full --tries-verify-mode=local --pruneancient --db.engine=pebble --rpc.txfeecap 0 --rpc.gascap 0`

#### Actual behaviour

```
root@Ubuntu-2204-jammy-amd64-base /jiedian/bsc # geth --config ./config.toml --datadir ./node  --cache 86016 --rpc.allow-unprotected-txs --history.transactions=0 --syncmode=full --tries-verify-mode=local --pruneancient --db.engine=pebble --rpc.txfeecap 0 --rpc.gascap 0
INFO [12-19|15:25:12.298] Starting Geth on BSC mainnet...
INFO [12-19|15:25:12.299] Maximum peer count                       ETH=200 LES=0 total=200
INFO [12-19|15:25:12.300] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [12-19|15:25:12.300] Using pebble as db engine
Fatal: Could not open database: MemTableSize (4.0GB) must be < 4.0GB
```

#### Steps to reproduce the behaviour

When submitting logs: please submit them as text and not screenshots.

serfiraz21:
remove `--cache` flag or reduce to under 4gib

flywukong:
Please lower you cache setting -cache to run . The issue is cause by the upstream pebble. we will  have a temp fix on it.

PR #2120, opened 2023-12-28
fix:  fix pebble mem table size boundary error

flywukong:
### Description
If mem table size  exceeds 4G, the memtable will be set to 4G in bsc code .  But Pebble has not handled this boundary well , it require less than 4G but not equal 4G . (https://github.com/cockroachdb/pebble/pull/3008)
 
### Rationale

solve issue(https://github.com/bnb-chain/bsc/issues/2091)

### Example

### Changes

Notable changes: 
* ...


Fynnss (approved):
LGTM

PR #2123, opened 2023-12-29
p2p, eth: improve logs

weiihann:
### Description
This PR improves user experience by modifying logs output, mainly on:
1. Do not show `Current status` for regular nodes, only used for bootnodes
2. Change log level for `Expired request does not exist` from `Error` to `Trace`

### Rationale
For the 1st log change, `Current status` log is meant for bootnodes to monitor the discovery status. For regular nodes, the information is not useful.

For the 2nd log change, when `*queue.Revoke()` and `*queue.expire()` happen concurrently, the resulting pending pool will no longer contain the peer request. Hence, `Expired request does not exist` will be shown. It's not a useful log for an average user, so its log level can be changed to `Trace`.


zzzckck (approved):
LGTM

Issue #2131, opened 2024-01-03
BSC version 1.3.7 got panic when syncing using pbss snapshot geth-pbss-pebble-20231217.tar.lz4


Tronglx:
```go
Panic:
github.com/ethereum/go-ethereum/core/rawdb.(*ResettableFreezer).AncientRange(0xcfee40?, {0x28e4c02?, 0xc0149c2828?}, 0xc0149c2928?, 0x248d3a0?, 0xc014d08060?)
	/home/runner/work/bsc/bsc/core/rawdb/freezer_resettable.go:126 +0x5c
github.com/ethereum/go-ethereum/core/rawdb.ReadStateHistoryMetaList(...)
	/home/runner/work/bsc/bsc/core/rawdb/accessors_state.go:180
github.com/ethereum/go-ethereum/trie/triedb/pathdb.checkHistories(0x0, 0x7e83553c7d1bf918?, 0x5028b97545f8df6c?, 0xc0154a7750)
	/home/runner/work/bsc/bsc/trie/triedb/pathdb/history.go:548 +0x85
github.com/ethereum/go-ethereum/trie/triedb/pathdb.(*Database).Recoverable(0xc0113db900, {0x70, 0x16, 0x98, 0xfd, 0xca, 0xe7, 0x2d, 0xa0, 0x18, ...})
	/home/runner/work/bsc/bsc/trie/triedb/pathdb/database.go:363 +0x205
github.com/ethereum/go-ethereum/trie.(*Database).Recoverable(0x7fdafcede2e8?, {0x70, 0x16, 0x98, 0xfd, 0xca, 0xe7, 0x2d, 0xa0, 0x18, ...})
	/home/runner/work/bsc/bsc/trie/database.go:320 +0x45
github.com/ethereum/go-ethereum/core.NewBlockChain({0x33776f8?, 0xc0014b4cc0}, 0x0?, 0x7ffffffe805afca8?, 0x0?, {0x3365fe0?, 0xc001392a00?}, {{0x0, 0x0}, 0x0, ...}, ...)
	/home/runner/work/bsc/bsc/core/blockchain.go:403 +0x14b0
github.com/ethereum/go-ethereum/eth.New(0xc0001860e0, 0xc001701000)
	/home/runner/work/bsc/bsc/eth/backend.go:252 +0x170f
github.com/ethereum/go-ethereum/cmd/utils.RegisterEthService(0x0?, 0xc001701000)
[...]

Issue #2110, opened 2023-12-26
can't start bsc - Fatal: Could not open database: open /srv/bsc/geth/chaindata/ancient/chain/bodies.cidx: no such file or directory


rmlcu:
#### System information

bsc-node:~$ sudo /srv/bsc/bsc-geth version
Geth
Version: 1.3.6
Git Commit: 5ee77bbe8bba2f1e9b7bc9368347fd531891ec7e
Git Commit Date: 20231214
Architecture: amd64
Go Version: go1.20.11
Operating System: linux
GOPATH=
GOROOT=/opt/hostedtoolcache/go/1.20.11/x64

bsc-node:~$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian

bsc-node:~$ sudo cat /etc/systemd/system/bsc-geth.service
[Unit]
Description=bsc geth go client
After=syslog.target network-online.target

[Service]
User=geth
Group=geth
#Type=simple
ExecStart=/srv/bsc/bsc-geth --config=/srv/bsc/config.toml --datadir=/srv/bsc --nat=extip:1.2.3.4 \
        --history.transactions=5000000 --syncmode=full --tries-verify-mode=none --diffblock=5000 \
        --cache=18000 --rpc.allow-unprotected-txs --db.engine=pebble
#KillMode=process
KillSignal=SIGINT
TimeoutStopSec=90
Restart=on-failure
RestartSec=10s

[Install]
WantedBy=multi-user.target

#### Expected behaviour

node successfully starting

#### Actual behaviour

node is failing to start:
```
-- Boot 6f5a36d8fb0a45ff894b0834f496879c --
Dec 26 17:01:46 bsc-node systemd[1]: Started bsc geth go client.
Dec 26 17:01:47 bsc-node bsc-geth[28939]: INFO [12-26|17:01:47.135] Starting Geth on BSC mainnet...
Dec 26 17:01:47 bsc-node bsc-geth[28939]: WARN [12-26|17:01:47.136] Config field is deprecated and won't have an effect name=ethconfig.Config.EWASMInterpreter
[...]

PR #2135, opened 2024-01-05
[WIP]cmd, core: resolve scheme from a read-write database

sysvm:
### Description

This pull request solves four pull requests: [#2110](https://github.com/bnb-chain/bsc/issues/2110), [#2131](https://github.com/bnb-chain/bsc/issues/2131),  [#2141](https://github.com/bnb-chain/bsc/issues/2141), [#2146](https://github.com/bnb-chain/bsc/issues/2146)

1. The first question: resolve state scheme from a read-write database:

This issue can occur when unclean shutdown happens and freezer tables need to be repair. In database read-only mode, the freezer will only be validated but not repair and terminates the node abnormally.
This pull request moves the scheme check into the Ethereum constructor where the read-write database is already initialized. This approach can avoid opening read-only database in very early stage, to prevent this error totally.

2. The second question: ancient directory should be compatible with hash and path mode:

Current geth doesn't deal with chain freezer dir in a good way. Therefore, this pr refactors `resolveChainFreezerDir` to fix in pruned and non-pruned mode.

### Rationale

These will influence how should users use `pruenancient` mode that block data could be replaced in which.

### Example

N/A

### Changes

Notable changes: 
* N/A

### DIFF
--- a/cmd/geth/chaincmd.go
+++ b/cmd/geth/chaincmd.go
@@ -655,7 +655,7 @@ func dump(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	triedb := utils.MakeTrieDatabase(ctx, db, true, false) // always enable preimage lookup
+	triedb := utils.MakeTrieDatabase(ctx, db, true, true) // always enable preimage lookup
 	defer triedb.Close()
 
 	state, err := state.New(root, state.NewDatabaseWithNodeDB(db, triedb), nil)
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -516,6 +516,10 @@ func Open(o OpenOptions) (ethdb.Database, error) {
 	if err != nil {
 		return nil, err
 	}
+	if ReadAncientType(kvdb) == PruneFreezerType {
+		log.Info("db is pruned, forcefully set pruneancient flag for db")
+		o.PruneAncientData = true
+	}
 	if len(o.AncientsDirectory) == 0 {
 		return kvdb, nil
 	}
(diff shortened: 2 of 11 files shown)


Fynnss (dismissed):
LGTM

cmd/geth/chaincmd.go
```diff
@@ -655,7 +655,7 @@ func dump(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	triedb := utils.MakeTrieDatabase(ctx, db, true, false) // always enable preimage lookup
```
Fynnss, line 658: Why not readonly mode? It's dangerous.

sysvm: Original code is not read-only mode. I change it into read-only mode.

core/rawdb/database.go
```diff
@@ -219,12 +219,26 @@ func NewFreezerDb(db ethdb.KeyValueStore, frz, namespace string, readonly bool,
...
 	// - chain freezer is not initialized
 	// - chain freezer exists in legacy location (root ancient folder)
 	freezer := path.Join(ancient, chainFreezerName)
+	state := path.Join(ancient, stateFreezerName)
+	if pruned {
+		if common.FileExist(freezer) {
+			return freezer
+		}
+		if common.FileExist(state) {
+			return freezer
```
zzzckck, line 234: not state?

buddh0:
some PR are from go-ethereum
so I think, we can format this PR:
PR from go-ethereum use cherry-pick and solve conflict
then apply our commit

zzzckck:
LGTM, would approve once updated to keep the origin commit record.

PR #2144, opened 2024-01-09
p2p: reset disconnect set with magic enode ID

weiihann:
### Description
Related to #2140.

This PR adds a small feature to reset the disconnect enode set by removing a peer (i.e. via `admin_removePeer`) using a hardcoded enode ID.

### Rationale
Previously, the only way to reset the disconnect enode set was to restart the node. This may not be very inconvenient. Hence, we need a mechanism to reset the set without restarting the node.

By sending a hardcoded "magic" enode ID as a parameter to `admin_removePeer`, the set will be reset. The hardcoded "magic" enode ID is referenced from the test cases.

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -74,6 +74,10 @@ var (
 	errServerStopped       = errors.New("server stopped")
 	errEncHandshakeError   = errors.New("rlpx enc error")
 	errProtoHandshakeError = errors.New("rlpx proto error")
+
+	// magicEnodeID is a special enode ID that can be used to disconnect all peers
+	// enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439
+	magicEnodeID = enode.ID{52, 49, 195, 147, 158, 30, 226, 166, 52, 94, 151, 106, 130, 52, 249, 135, 1, 82, 214, 72, 121, 243, 11, 194, 114, 160, 116, 246, 133, 158, 117, 232}
 )
 
 // Config holds Server options.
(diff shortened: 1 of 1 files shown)


p2p/server.go
```diff
@@ -367,6 +368,16 @@ func (srv *Server) RemovePeer(node *enode.Node) {
...
+		srv.disconnectEnodeSetMu.Lock()
+		srv.disconnectEnodeSet = make(map[enode.ID]struct{})
+		srv.disconnectEnodeSetMu.Unlock()
+		srv.log.Debug("Reset disconnect enode set")
+		return
+	}
+
 	// Disconnect the peer on the main loop.
 	srv.doPeerOp(func(peers map[enode.ID]*Peer) {
 		srv.dialsched.removeStatic(node)
```
zzzckck, line 383: we can move the clean up operation inside of this function, it will be executed in the server's main loop, so it will not have the concurrent access issue, so need the extra lock

zzzckck: besides, why not use Magic EnodeID? I think it could be much easier for maintenance.

weiihann: My initial thought is that the ID is guaranteed to be unique because a node can't connect to itself as a peer. But this introduces an additional where a node operator has to call `admin_nodeInfo` or store the enode ID somewhere. Yeah, not too good for maintenance.

Already amended to use a magic enode ID instead.

p2p/server.go
```diff
@@ -74,6 +74,10 @@ var (
 	errServerStopped       = errors.New("server stopped")
 	errEncHandshakeError   = errors.New("rlpx enc error")
 	errProtoHandshakeError = errors.New("rlpx proto error")
+
+	// magicEnodeID is a special enode ID that can be used to disconnect all peers
+	// enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439
```
zzzckck, line 79: why not just: "enode://0000.....0000"?

weiihann: I tried and it doesn't work. The reason is because it must be a valid secp256k1 public key. Specifically, "0000.....0000" is not a valid point in the secp256k1 curve, so it's not a valid enode.

`enode://1dd9d65c4552b5eb43d5ad55a2ee3f56c6cbc1c64a5c8d659f51fcd51bace24351232b8d7821617d2b29b54b81cdefb9b3e9c37d7fd5f63270bcc9e1a6f6a439` is chosen because it's being used in the test cases, so it's convenient to use it as the magic enode ID.

zzzckck (approved):

Issue #2143, opened 2024-01-08
Value of gasUsed is different on "eth_getTransactionReceipt" and "debug_traceTransaction"?


Aracki:
I'm running several binance nodes (bsc 1.3.7 version), but I discovered differences for `gasUsed` value between these two methods.

Transaction with this problem: `0x7eba4edc7c1806d6ee1691d43513838931de5c94f9da56ec865721b402f775b0`.

```
curl --request POST \
  --url http://node-bsc:8545 \
  --header 'Content-Type: application/json' \
  --data '{
	"jsonrpc": "2.0",
	"id": 1,
	"method": "eth_getTransactionReceipt",
	"params": [
		"0x7eba4edc7c1806d6ee1691d43513838931de5c94f9da56ec865721b402f775b0"
	]
}'
```
returns:
```
"gasUsed":"0x1afd3"
```

although, this one:
```
curl --request POST \
  --url http://node-bsc:8545 \
  --header 'Content-Type: application/json' \
  --data '{
	"jsonrpc": "2.0",
	"id": 123,
	"method": "debug_traceTransaction",
	"params": [
		"0x7eba4edc7c1806d6ee1691d43513838931de5c94f9da56ec865721b402f775b0",
		{"tracer": "callTracer"}
	]
}'
```
returns:
```
"gasUsed":"0x412c3"
```

Here are some additional flags used for running nodes:

```
            "--networkid=56",
            "--syncmode=full",
            "--gcmode=archive",
            "--history.transactions=0",
            "--rpc.allow-unprotected-txs",
            "--cache.preimages",    
```

deepcrazy:
Hi there, Can you provide your full geth startup command too? Also, please share the logs by setting log level to `debug`.

qybdyx:
It's know issue of hertzfix. There is no good solution for it, just ignore it.

Aracki:
@deepcrazy Unfortunately, I've lost logs older than 30 days. And block `33851236` happened before ~40days. 

@qybdyx I have to ignore it, but I'm more interested whether the unwind would help potentially? Or downloading some old archive snapshot?

zzzckck:
> @deepcrazy Unfortunately, I've lost logs older than 30 days. And block `33851236` happened before ~40days.
> 
> @qybdyx I have to ignore it, but I'm more interested whether the unwind would help potentially? Or downloading some old archive snapshot?

[...]

PR #2147, opened 2024-01-10
eth/tracer/api.go: the patch for trace rpc about hertz fix

qybdyx:
### Description

Trace transaction will fail in block [33851236](https://bscscan.com/block/33851236)

### Rationale

Related to #2143. Block 33851236 will have different execution result. This corner case was fixed by [HertzFix](https://forum.bnbchain.org/t/about-the-hertzfix/2400), but trace history block still have the problem. This pr will fix it.

### Example

### Changes

Notable changes: 
* eth/tracer/api.go: add a patch for corner case.

### DIFF
--- a/eth/tracers/api.go
+++ b/eth/tracers/api.go
@@ -968,6 +968,9 @@ func (api *API) traceTx(ctx context.Context, message *core.Message, txctx *Conte
 	if config == nil {
 		config = &TraceConfig{}
 	}
+	if res, _ := patchForSpecialCase(txctx.TxHash, config); res != nil {
+		return res, nil
+	}
 	// Default tracer is the struct logger
 	tracer = logger.NewStructLogger(config.Config)
 	if config.Tracer != nil {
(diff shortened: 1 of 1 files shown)


eth/tracers/api.go
```diff
@@ -1075,3 +1078,94 @@ func overrideConfig(original *params.ChainConfig, override *params.ChainConfig)
...
+		TypeString: vm.STATICCALL.String(),
+	}
+
+	var tracerConfig callTracerConfig
+	if config == nil {
+		return nil, nil
+	}
+	if config.Tracer == nil || *config.Tracer != "callTracer" {
```
zzzckck, line 1146: how about combine to one line:
```
if config == nil || config.Tracer == nil || *config.Tracer != "callTracer" {
```

zzzckck:
no strong demand for it and it could make the code dirty, so we'd rather not to fix this corner case RPC call.

PR #2151, opened 2024-01-10
feat: enable NoDial should still dial static nodes

### LINKED ISSUE #2149: feat: enable NoDial should still dial static nodes
# Rationale
In some cases, we would not want to dial any external peers but still want to dial the preconfigured static nodes. In this case, the config file would look something like:

```
...
[Node.P2P]
NoDial = true
StaticNodes = [...]
``` 

In the current implementation, setting `NoDial` to true will by default not dial any peers at all, including the static nodes. This should be modified.

weiihann:
### Description
Resolves #2149

### Changes
Previously, when `NoDial` is set to true, then the maximum dial peers is always going to be 0. Hence, the node is never going to dial static nodes.

With this code change, when `NoDial` is set to true, the maximum dial peers will be set to the number of static peers. With this, the dial scheduler knows that there are available slots to dial the static peers.

One concern regarding this design is that, if the node doesn't dial the static nodes successfully, will it then dial the other peers, since there are available dial slots?

Fortunately, the answer is no. Based on the code [here](https://github.com/bnb-chain/bsc/blob/master/p2p/dial.go#L405-L414), it's guaranteed to return the number of static nodes. Therefore, there will be [no available slots](https://github.com/bnb-chain/bsc/blob/master/p2p/dial.go#L238-L244) to dial other peers.

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -169,7 +169,7 @@ type Config struct {
 	// is used to dial outbound peer connections.
 	Dialer NodeDialer `toml:"-"`
 
-	// If NoDial is true, the server will not dial any peers.
+	// If NoDial is true, the server will not dial any peers, except for pre-configured static nodes.
 	NoDial bool `toml:",omitempty"`
 
 	// If EnableMsgEvents is set then the server will emit PeerEvents
@@ -676,7 +676,10 @@ func (srv *Server) SetFilter(f forkid.Filter) {
 }
 
 func (srv *Server) maxDialedConns() (limit int) {
-	if srv.NoDial || srv.MaxPeers == 0 {
+	if srv.NoDial {
+		return len(srv.StaticNodes) + len(srv.VerifyNodes)
+	}
+	if srv.MaxPeers == 0 {
 		return 0
 	}
 	if srv.DialRatio == 0 {


weiihann:
> what about `VerifyNodes` and `TrustedNodes`are we ok we are not connecting to those?

I think we can include VerifyNodes, since the node is going to connect with the VerifyNodes when started right away, exactly the same as StaticNodes. For TrustedNodes, I don't think it's necessary because the scenario here targets `NoDial = true`, so TrustedNodes won't be connected anyways.

du5:
I think your needs should be solved using `--nodiscover=true` instead of modifying NoDial
```toml
[Node.P2P]
NoDiscovery = true
```

```
--nodiscover                   (default: false)
      Disables the peer discovery mechanism (manual peer addition)
```

weiihann:
> I think your needs should be solved using `--nodiscover=true` instead of modifying NoDial
> 
> ```toml
> [Node.P2P]
> NoDiscovery = true
> ```
> 
> ```
> --nodiscover                   (default: false)
>       Disables the peer discovery mechanism (manual peer addition)
> ```
> 
> However, your changes can indeed meet your needs, but I think nodiscover is more suitable for your needs. It can manually manage, add and delete

Not exactly. `NoDiscovery` disables the UDP port, which means that other nodes cannot discover our node at all. With this PR, enabling `NoDial` and `StaticNodes` would still allow other peers to discover and connect to our node, but our node won't be the one initiating the connection request (i.e. dialing), except for the static nodes.

zzzckck (approved):
LGTM, I think it is reasonable that user could only want to establish connection with StaticNodes.

PR #818, opened 2022-03-28
add sharedStorage to the prefetcher of miner

flywukong:
### Description
Since the miner has prefetched data , this PR make the miner main process and the prefetch process share a shared_pool with the stateDB.  By this way , the prefetch data that can be directly fished to the L1.5 layer and make miner prefetch faster

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes: 
* add each change in a bullet point here
* ...

PR #2172, opened 2024-01-19
improve: remove sharedpool from miner

zzzckck:
### Description
SharedPool was introduced since [v1.1.9](https://github.com/bnb-chain/bsc/releases/tag/v1.1.9), but it has a potential bug and is no longer supported since [hertzFix](https://forum.bnbchain.org/t/about-the-hertzfix/2400).

This PR is to remove the sharedPool code in miner module, which was introduced in this PR https://github.com/bnb-chain/bsc/pull/818 and is useless now.

It has no impact to the current logic, but just code improve

### Rationale
NA

### Example
NA

### Changes
NA

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1022,11 +1022,6 @@ func (bc *BlockChain) SnapSyncCommitHead(hash common.Hash) error {
 	return nil
 }
 
-// StateAtWithSharedPool returns a new mutable state based on a particular point in time with sharedStorage
-func (bc *BlockChain) StateAtWithSharedPool(root common.Hash) (*state.StateDB, error) {
-	return state.NewWithSharedPool(root, bc.stateCache, bc.snaps)
-}
-
 // Reset purges the entire blockchain, restoring it to its genesis state.
 func (bc *BlockChain) Reset() error {
 	return bc.ResetWithGenesisBlock(bc.genesisBlock)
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -1022,11 +1022,6 @@ func (bc *BlockChain) SnapSyncCommitHead(hash common.Hash) error {
 	return nil
 }
 
-// StateAtWithSharedPool returns a new mutable state based on a particular point in time with sharedStorage
-func (bc *BlockChain) StateAtWithSharedPool(root common.Hash) (*state.StateDB, error) {
-	return state.NewWithSharedPool(root, bc.stateCache, bc.snaps)
```
Fynnss, line 1027: Does the `NewWithSharedPool` used in the `InsertChain` function need to be removed?

https://github.com/bnb-chain/bsc/blob/f28b98a9947b52816714f90326b2545b08606b1a/core/blockchain.go#L2013

zzzckck: > Does the `NewWithSharedPool` used in the `InsertChain` function need to be removed?
> 
> https://github.com/bnb-chain/bsc/blob/f28b98a9947b52816714f90326b2545b08606b1a/core/blockchain.go#L2013

I am afraid not, `InsertChain` needs make sure backward compatibility, like the `hertzFix` case, so we can not remove the sharedPool logic from `InsertChain`

buddh0 (approved):

zjubfd (approved):

Fynnss (approved):

PR #2181, opened 2024-01-25
core: use finalized block as the chain freeze indicator

jingjunLi:
### Description

This pull request uses the finalized block tag as the indicator for chain freezing.

Originally, the latest 90K blocks will be retained in the key-value store by default, which are regarded as "non-finalized" in the context of Proof-of-work. However, in Proof-of-stake, we do have the concrete finalized block tag from the consensus client. Therefore, it makes more sense to directly rely on the Finalized for chain freezing.

### Rationale

By deploying it on a full node, we can see there are only 23 blocks left in the key-value store.
image

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/cmd/geth/pruneblock_test.go
+++ b/cmd/geth/pruneblock_test.go
@@ -164,10 +164,10 @@ func BlockchainCreator(t *testing.T, chaindbPath, AncientPath string, blockRemai
 
 	// Force run a freeze cycle
 	type freezer interface {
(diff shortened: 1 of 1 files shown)


Fynnss:
Change the `freezerRecheckInterval` to 15s or less? 

Currently it's 1min and will generate 20 blocks during the interval, plus 2-3 un-finalized blocks.

PR #2183, opened 2024-01-25
p2p: resolved deadlock on p2p server shutdown

MatusKysel:
### Description
This PR removes forceful stop of p2p server that was caused by deadlock on channels. In `protoTracker()` we should be waiting for signals from all `handlerDoneCh`, but with return on `stopCh` we returned and protocol handler got stuck on sending signal to `handlerDoneCh`.  This changes removes this deadlock and also part of the code from p2p server that forcefully killed the server.

### DIFF
--- a/p2p/server.go
+++ b/p2p/server.go
@@ -445,7 +445,18 @@ func (srv *Server) Stop() {
 	}
 	close(srv.quit)
 	srv.lock.Unlock()
-	srv.loopWG.Wait()
+
+	stopChan := make(chan struct{})
+	go func() {
+		srv.loopWG.Wait()
+		close(stopChan)
+	}()
+
+	select {
+	case <-stopChan:
+	case <-time.After(defaultDialTimeout): // we should use defaultDialTimeout as we can dial just before the shutdown
+		srv.log.Warn("stop p2p server timeout, forcing stop")
+	}
 }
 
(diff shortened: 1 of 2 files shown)


p2p/server.go
```diff
@@ -448,18 +445,7 @@ func (srv *Server) Stop() {
...
-		srv.loopWG.Wait()
-		close(stopChan)
-	}()
-
-	select {
-	case <-stopChan:
-	case <-time.After(stopTimeout):
-		srv.log.Warn("stop p2p server timeout, forcing stop")
-	}
+	srv.loopWG.Wait()
```
zzzckck, line 448: Would is possible that there is other deadlocks, so maybe we still need this timeout?

zzzckck:
Could you also post the deadlock callstack?

p2p/server.go
```diff
@@ -448,18 +445,7 @@ func (srv *Server) Stop() {
 	}
 	close(srv.quit)
 	srv.lock.Unlock()
-
-	stopChan := make(chan struct{})
```
GalaIO, line 452: The previous code tries to skip `srv.loopWG.Wait()` if `stopTimeout` reach, maybe this is to solve issues before?
Is it ok just remove above `case <-cs.handler.stopCh:`, and keep this logic unchanged?

zzzckck (approved):

GalaIO (approved):

PR #2190, opened 2024-01-29
txpool/legacypool: deprecate already known error

weiihann:
### Description
This PR deprecates the usage of `ErrAlreadyKnown` in the `legacypool` package. Existing dependencies on this error type should use `txpool.ErrAlreadyKnown` instead.

### Rationale
Originally, upon triggering `legacypool.ErrAlreadyKnown`, this error will be propagated back to the initial caller functions.
For example, in [this code](https://github.com/bnb-chain/bsc/blob/master/eth/fetcher/tx_fetcher.go#L323), the expected behavior is that the metrics will be bumped if `txpool.ErrAlreadyKnown` function is used. However, `legacypool.ErrAlreadyKnown` and `txpool.ErrAlreadyKnown` are 2 different error types with the same error messages. Hence, the case of checking if the error is of `txpool.ErrAlreadyKnown` will not be triggered so it will go to the default case instead.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -57,10 +57,6 @@ const (
 )
 
 var (
-	// ErrAlreadyKnown is returned if the transactions is already contained
-	// within the pool.
-	ErrAlreadyKnown = errors.New("already known")
-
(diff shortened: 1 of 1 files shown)


zzzckck (approved):
LGTM

PR #2199, opened 2024-02-02
fix: add check for block number

Pythonberg1997:
### Description

This pr is to fix a bug within precompile contract `verifyDoubleSignEvidence`

### Rationale

An arbitrary block number that is bigger than 2^256 could overwrite the address in the returned bytes.

### Example

### Changes

Notable changes: 
* add check for block number in `verifyDoubleSignEvidence`

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -1399,7 +1399,8 @@ func (c *verifyDoubleSignEvidence) RequiredGas(input []byte) uint64 {
 }
 
 var (
-	extraSeal = 65
+	extraSeal  = 65
+	maxUnit256 = new(big.Int).SetUint64(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
 )
 
 type DoubleSignEvidence struct {
@@ -1432,6 +1433,9 @@ func (c *verifyDoubleSignEvidence) Run(input []byte) ([]byte, error) {
 	}
 
 	// basic check
+	if header1.Number.Cmp(maxUnit256) == 1 || header2.Number.Cmp(maxUnit256) == 1 {
+		return nil, ErrExecutionReverted
+	}
 	if header1.Number.Cmp(header2.Number) != 0 {
 		return nil, ErrExecutionReverted
 	}


core/vm/contracts.go
```diff
@@ -1432,6 +1433,9 @@ func (c *verifyDoubleSignEvidence) Run(input []byte) ([]byte, error) {
 	}
 
 	// basic check
+	if header1.Number.Cmp(maxUnit256) == 1 || header2.Number.Cmp(maxUnit256) == 1 {
```
j75689, line 1436: `header2.Number.Cmp(maxUnit256) > 1`?

OleksiyBiliy (approved):
[Uploading zkmerkle_20231101.zip…]()

j75689 (approved):

core/vm/contracts.go
```diff
@@ -1399,7 +1399,8 @@ func (c *verifyDoubleSignEvidence) RequiredGas(input []byte) uint64 {
 }
 
 var (
-	extraSeal = 65
+	extraSeal  = 65
+	maxUnit256 = new(big.Int).SetUint64(0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)
```
zjubfd, line 1403: `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`, but the function here is `SetUint64`.

I will suggest just set the max  uint64 not uint256

j75689: try this, new(big.Int).SetBytes([]byte{
		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	}).Bytes()

core/vm/contracts.go
j75689, line 1436: or you can compare length of number here, `header1.Number.Cmp(maxUnit256) == 1 || len(header1.Number.Bytes()) >32`

core/vm/contracts.go
```diff
@@ -1432,6 +1433,9 @@ func (c *verifyDoubleSignEvidence) Run(input []byte) ([]byte, error) {
 	}
 
 	// basic check
+	if header1.Number.Cmp(maxUnit256) == 1 || header2.Number.Cmp(maxUnit256) == 1 {
+		return nil, ErrExecutionReverted
+	}
 	if header1.Number.Cmp(header2.Number) != 0 {
 		return nil, ErrExecutionReverted
```
zjubfd, line 1440: `copy(returnBz[52-len(evidenceHeightBz):], evidenceHeightBz)` is not a good code style,  why not start from index 20 directly but `52-len(evidenceHeightBz)`

core/vm/contracts.go
Pythonberg1997: Done

core/vm/contracts.go
Pythonberg1997: Done

core/vm/contracts.go
Pythonberg1997: Because we need zero-padding

zjubfd (approved):

core/vm/contracts.go
```diff
@@ -1399,7 +1399,8 @@ func (c *verifyDoubleSignEvidence) RequiredGas(input []byte) uint64 {
 }
 
 var (
-	extraSeal = 65
+	extraSeal          = 65
```
j75689, line 1402: `extraSeal` can be const.

core/vm/contracts.go
```diff
@@ -1399,7 +1399,8 @@ func (c *verifyDoubleSignEvidence) RequiredGas(input []byte) uint64 {
...
+	errInvalidEvidence = errors.New("invalid double sign evidence")
```
j75689, line 1403: move to `errors.go`

PR #2202, opened 2024-02-04
vote: lazy init vote manager's signer;

GalaIO:
### Description

vote: lazy init vote manager's signer;

### Rationale

If the node not config to mine

### Example

If the node does not set `--mine`, then the vote singer should not be initialized directly, but when the mine is enabled through flags or RPC, then the specified account is initialized for voting.

This change may facilitate configuration optimization and avoid configuration complexity.

### Changes

Notable changes: 
* vote: lazy init vote manager's signer;

### DIFF
--- a/core/vote/vote_manager.go
+++ b/core/vote/vote_manager.go
@@ -43,25 +43,24 @@ type VoteManager struct {
 	journal *VoteJournal
 
 	engine consensus.PoSA
+
+	// signer config
+	blsPasswordPath string
+	blsWalletPath   string
+	lock            sync.RWMutex
 }
 
 func NewVoteManager(eth Backend, chain *core.BlockChain, pool *VotePool, journalPath, blsPasswordPath, blsWalletPath string, engine consensus.PoSA) (*VoteManager, error) {
 	voteManager := &VoteManager{
-		eth:         eth,
-		chain:       chain,
-		chainHeadCh: make(chan core.ChainHeadEvent, chainHeadChanSize),
-		syncVoteCh:  make(chan core.NewVoteEvent, voteBufferForPut),
-		pool:        pool,
-		engine:      engine,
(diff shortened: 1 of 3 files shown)


NathanBSC:
I prefer devops to delete `VoteEnable` in config.toml
and just use `--mine` and `--vote` to control whether to mine or vote

NathanBSC:
when there are some issues with bls wallet
and operator use ipc to `StartMining`
then the node will still to produce blocks, that means  vote fails silently

0xbundler:
> when there are some issues with bls wallet and operator use ipc to `StartMining` then the node will still to produce blocks, that means vote fails silently
> 
It will likely cause an error code, and cannot start mine.

GalaIO:
> when there are some issues with bls wallet and operator use ipc to `StartMining` then the node will still to produce blocks, that means vote fails silently
> 
It will likely cause an error code, and cannot start mine.

NathanBSC:
> > when there are some issues with bls wallet and operator use ipc to `StartMining` then the node will still to produce blocks, that means vote fails silently
> 
> It will likely cause an error code, and cannot start mine.

OK
the difference left is that then: 
if --vote added or VoteEnable set to true and bls wallet have issue, 
do we need the node to fail at the start time? 
**I think, it do.**

zzzckck:
> > > when there are some issues with bls wallet and operator use ipc to `StartMining` then the node will still to produce blocks, that means vote fails silently
> > 
> > 
> > It will likely cause an error code, and cannot start mine.
> 
> OK the difference left is that then: if --vote added or VoteEnable set to true and bls wallet have issue, do we need the node to fail at the start time? **I think, it do.**

I agree, if `VoteEnable = true` is set specifically, it is better to not start when it failed to load the BLS key.

PR #2225, opened 2024-02-20
types: fix header nil withdraw decode bug;

GalaIO:
### Description

types: fix header nil withdraw decode bug;

### Rationale

When `BlobGasUsed` is set but `WithdrawalsHash` is left nil, the error occurs `rlp:' is reported in deserialization. input string too short for common.Hash, decoding into (types.Header).WithdrawalsHash`.

Everything was fine after repair.

### Changes

Notable changes: 
* types: fix header nil withdraw decode bug;

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -108,14 +108,16 @@ type Header struct {
 	// BaseFee was added by EIP-1559 and is ignored in legacy headers.
 	BaseFee *big.Int `json:"baseFeePerGas" rlp:"optional"`
 
-	// WithdrawalsHash was added by EIP-4895 and is ignored in legacy headers.
-	WithdrawalsHash *common.Hash `json:"withdrawalsRoot" rlp:"optional"`
-
 	// BlobGasUsed was added by EIP-4844 and is ignored in legacy headers.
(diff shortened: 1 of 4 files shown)


core/types/block.go
```diff
@@ -108,15 +108,16 @@ type Header struct {
...
-	WithdrawalsHash *common.Hash `json:"withdrawalsRoot" rlp:"optional"`
-
 	// BlobGasUsed was added by EIP-4844 and is ignored in legacy headers.
 	BlobGasUsed *uint64 `json:"blobGasUsed" rlp:"optional"`
 
 	// ExcessBlobGas was added by EIP-4844 and is ignored in legacy headers.
 	ExcessBlobGas *uint64 `json:"excessBlobGas" rlp:"optional"`
 
+	// Leave fields not used by BSC last to avoid being serialized by RLP
+	// WithdrawalsHash was added by EIP-4895 and is ignored in legacy headers.
```
emailtovamos, line 118: We should mention in the comment here that WithdrawalsHash is always `nil` for bsc/parlia to avoid confusion in future.

emailtovamos (approved):

PR #2228, opened 2024-02-21
core: rephrase TriesInMemory log

weiihann:
### Description
This PR rephrases the log message of `TriesInMemory` when it differs from the default value.

### Rationale
The original intention is to inform users that they must use the same `TriesInMemory` value to prune data. However, with the introduction of PBSS, offline pruning is no longer required. Therefore, this log message should only be applied when users are running with hash-based state scheme.


zzzckck (approved):
LGTM

Fynnss (approved):

PR #2229, opened 2024-02-21
feat: support blob storage & miscs;

GalaIO:
### Description

The PR mainly implements blob storage for BSC blobtx feature. It uses KVDB & freezerDB to support blob CRUD, and implement blob expiry, and block DA verification.

### Rationale

We can use `--override.cancun` to enable cancun fork.

### Changes

Notable changes: 
* blob: support config blob extra reserve;
* rawdb: support to CRUD blobs;
* freezer: support to freeze block blobs;
* chainconfig: use cancun fork for BSC;
* feat: fill WithdrawalsHash when BSC enable cancun fork;
...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1593,9 +1593,13 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 		// if enable cancun, it needs to write blobs too
 		if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
-			blobs, _ := bc.receivedBlobsCache.Load(block.Hash())
-			rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), blobs.(types.BlobTxSidecars))
-			bc.receivedBlobsCache.Delete(block.Hash())
+			blobs, exist := bc.receivedBlobsCache.Load(block.Hash())
+			if exist {
+				rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), blobs.(types.BlobTxSidecars))
+			} else {
+				rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), nil)
+				bc.receivedBlobsCache.Delete(block.Hash())
+			}
 		}
 		rawdb.WritePreimages(blockBatch, state.Preimages())
 		if err := blockBatch.Write(); err != nil {
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1358,7 +1358,8 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 		if first.NumberU64() == 1 {
 			if frozen, _ := bc.db.Ancients(); frozen == 0 {
 				td := bc.genesisBlock.Difficulty()
-				writeSize, err := rawdb.WriteAncientBlocks(bc.db, []*types.Block{bc.genesisBlock}, []types.Receipts{nil}, td)
+				// TODO(GalaIO): when sync the history block, it needs store blobs too.
+				writeSize, err := rawdb.WriteAncientBlocks(bc.db, []*types.Block{bc.genesisBlock}, []types.Receipts{nil}, td, nil)
 				if err != nil {
 					log.Error("Error writing genesis to ancients", "err", err)
 					return 0, err
@@ -1376,7 +1377,8 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 
 		// Write all chain data to ancients.
 		td := bc.GetTd(first.Hash(), first.NumberU64())
-		writeSize, err := rawdb.WriteAncientBlocks(bc.db, blockChain, receiptChain, td)
+		// TODO(GalaIO): when sync the history block, it needs store blobs too.
+		writeSize, err := rawdb.WriteAncientBlocks(bc.db, blockChain, receiptChain, td, nil)
 		if err != nil {
 			log.Error("Error importing chain data to ancients", "err", err)
 			return 0, err
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -285,9 +285,6 @@ type BlockChain struct {
 	// Cache for the blocks that failed to pass MPT root verification
 	badBlockCache *lru.Cache[common.Hash, time.Time]
 
-	// blobs
-	receivedBlobsCache sync.Map // it saves received blobs for validation & storage
-
 	// trusted diff layers
 	diffLayerCache             *exlru.Cache                          // Cache for the diffLayers
 	diffLayerChanCache         *exlru.Cache                          // Cache for the difflayer channel
@@ -651,11 +648,6 @@ func (bc *BlockChain) cacheDiffLayer(diffLayer *types.DiffLayer, diffLayerCh cha
 
 func (bc *BlockChain) cacheBlock(hash common.Hash, block *types.Block) {
 	bc.blockCache.Add(hash, block)
-	// try cache blob too
-	blob, ok := bc.receivedBlobsCache.Load(hash)
-	if ok {
-		bc.blobsCache.Add(hash, blob.(types.BlobTxSidecars))
-	}
 }
 
 // empty returns an indicator whether the blockchain is empty.
@@ -1545,9 +1537,7 @@ func (bc *BlockChain) writeBlockWithoutState(block *types.Block, td *big.Int) (e
 	rawdb.WriteBlock(batch, block)
 	// if enable cancun, it needs to write blobs too
 	if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
-		blobs, _ := bc.receivedBlobsCache.Load(block.Hash())
-		rawdb.WriteBlobs(batch, block.Hash(), block.NumberU64(), blobs.(types.BlobTxSidecars))
-		bc.receivedBlobsCache.Delete(block.Hash())
+		rawdb.WriteBlobs(batch, block.Hash(), block.NumberU64(), block.Blobs())
 	}
 	if err := batch.Write(); err != nil {
 		log.Crit("Failed to write block into disk", "err", err)
@@ -1593,13 +1583,7 @@ func (bc *BlockChain) writeBlockWithState(block *types.Block, receipts []*types.
 		rawdb.WriteReceipts(blockBatch, block.Hash(), block.NumberU64(), receipts)
 		// if enable cancun, it needs to write blobs too
 		if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
-			blobs, exist := bc.receivedBlobsCache.Load(block.Hash())
-			if exist {
-				rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), blobs.(types.BlobTxSidecars))
-			} else {
-				rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), nil)
-				bc.receivedBlobsCache.Delete(block.Hash())
-			}
+			rawdb.WriteBlobs(blockBatch, block.Hash(), block.NumberU64(), block.Blobs())
 		}
 		rawdb.WritePreimages(blockBatch, state.Preimages())
 		if err := blockBatch.Write(); err != nil {
@@ -2026,11 +2010,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		// TODO(GalaIO): move IsDataAvailable combine into verifyHeaders?
 		if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
 			if posa, ok := bc.engine.(consensus.PoSA); ok {
-				blobs, exist := bc.receivedBlobsCache.Load(block.Hash())
-				if !exist {
-					return it.index, fmt.Errorf("cannot find the target block's blob info, block: %v, hash: %v", block.NumberU64(), block.Hash())
-				}
-				if err = posa.IsDataAvailable(bc, block, blobs.(types.BlobTxSidecars)); err != nil {
+				if err = posa.IsDataAvailable(bc, block); err != nil {
 					return it.index, err
 				}
 			}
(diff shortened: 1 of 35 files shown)


emailtovamos:
Wouldn't this function also require modification to consider and save sidecars? 
`func (bc *BlockChain) WriteBlockAndSetHead(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB, emitHeadEvent bool) (status WriteStatus, err error)`

GalaIO:
> Wouldn't this function also require modification to consider and save sidecars? `func (bc *BlockChain) WriteBlockAndSetHead(block *types.Block, receipts []*types.Receipt, logs []*types.Log, state *state.StateDB, emitHeadEvent bool) (status WriteStatus, err error)`

Yes, I will leave some TODO for it. This pr is aim to create the basic CRUD for blobs, we can fix them later.

core/rawdb/chain_freezer.go
```diff
@@ -276,6 +277,11 @@ func (f *chainFreezer) freezeRange(nfdb *nofreezedb, number, limit uint64) (hash
 			if len(td) == 0 {
 				return fmt.Errorf("total difficulty missing, can't freeze block %d", number)
 			}
+			// TODO(GalaIO): blobs may nil before cancun fork
+			blobs := ReadBlobsRLP(nfdb, hash, number)
+			//if len(blobs) == 0 {
```
emailtovamos, line 282: Can't we check if it is cancun fork or not and have this check? Or will it be part of your TODO mentioned above?

core/rawdb/schema.go
```diff
@@ -203,6 +205,11 @@ func blockReceiptsKey(number uint64, hash common.Hash) []byte {
 	return append(append(blockReceiptsPrefix, encodeBlockNumber(number)...), hash.Bytes()...)
 }
 
+// blockBlobsKey = BlockBlobsPrefix + num (uint64 big endian) + hash
+func blockBlobsKey(number uint64, hash common.Hash) []byte {
```
emailtovamos, line 209: Can we write a description here mentioning what exactly are `number` and `hash`?

core/rawdb/accessors_chain_test.go
```diff
@@ -663,6 +745,38 @@ func makeTestReceipts(n int, nPerBlock int) []types.Receipts {
...
+		return nil
+	}
+	ret := make(types.BlobTxSidecars, n)
+	for i := 0; i < n; i++ {
+		blobs := make([]kzg4844.Blob, nPerTx)
+		commitments := make([]kzg4844.Commitment, nPerTx)
+		proofs := make([]kzg4844.Proof, nPerTx)
+		for i := 0; i < nPerTx; i++ {
+			io.ReadFull(rand2.Reader, blobs[i][:])
+			io.ReadFull(rand2.Reader, commitments[i][:])
```
emailtovamos, line 759: Shouldn't commitment be generated from blob instead of being random? Or am I missing something?

core/rawdb/accessors_chain.go
```diff
@@ -801,15 +801,70 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+func ReadBlobsRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValue {
+	var data []byte
+	db.ReadAncients(func(reader ethdb.AncientReaderOp) error {
+		// Check if the data is in ancients
+		if isCanon(reader, number, hash) {
+			data, _ = reader.Ancient(ChainFreezerBlobTable, number)
+			return nil
+		}
+		// If not, try reading from leveldb
+		data, _ = db.Get(blockBlobsKey(number, hash))
```
emailtovamos, line 827: Why no error check?

core/rawdb/accessors_chain.go
```diff
@@ -801,15 +801,70 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+	if err := rlp.DecodeBytes(data, &ret); err != nil {
+		log.Error("Invalid blob array RLP", "hash", hash, "err", err)
+		return nil
+	}
+	return ret
+}
+
+// WriteBlobs stores all the transaction blobs belonging to a block.
+// It could input nil for empty blobs.
+func WriteBlobs(db ethdb.KeyValueWriter, hash common.Hash, number uint64, blobs types.BlobTxSidecars) {
```
emailtovamos, line 849: Who will call this function? Can it be called from Parlia right after a block is saved to store sidecars?

core/rawdb/accessors_chain.go
GalaIO: oh, here is a geth common method to handle db get, you can check other read functions.

core/blockchain.go
```diff
@@ -277,12 +278,16 @@ type BlockChain struct {
...
 	txLookupCache *lru.Cache[common.Hash, txLookup]
+	blobsCache    *lru.Cache[common.Hash, types.BlobTxSidecars]
 
 	// future blocks are blocks added for later processing
 	futureBlocks *lru.Cache[common.Hash, *types.Block]
 	// Cache for the blocks that failed to pass MPT root verification
 	badBlockCache *lru.Cache[common.Hash, time.Time]
 
+	// blobs
+	receivedBlobsCache sync.Map // it saves received blobs for validation & storage
```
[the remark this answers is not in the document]
emailtovamos: Also I don't see receivedBlobsCache adding a sidecar anywhere? Will it be done in a later PR?

core/blockchain.go
```diff
@@ -1377,8 +1376,15 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 
 		// Write all chain data to ancients.
 		td := bc.GetTd(first.Hash(), first.NumberU64())
+		// TODO(GalaIO): if you just back fill blob, you need check blob ancient and reset it
```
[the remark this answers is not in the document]
GalaIO: Oh, I will opt for it to be simpler. Because freezerdb is composed of closely arranged 2GB files, continuous and uninterrupted elements need to be passed in, and the order needs to be guaranteed for each insert, so the blob table needs to be additionally initialized here.

core/rawdb/chain_freezer.go
```diff
@@ -251,17 +252,22 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 	}
 }
 
-func (f *chainFreezer) freezeRange(nfdb *nofreezedb, number, limit uint64) (hashes []common.Hash, err error) {
+func (f *chainFreezer) freezeRange(nfdb *nofreezedb, number, limit uint64, chainCfg *params.ChainConfig) (hashes []common.Hash, err error) {
```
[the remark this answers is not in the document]
GalaIO: I think it is ok, but do you mean the function needs to be the below?
```go
freezeRange(nfdb *nofreezedb, number, limit uint64, isCacncun func(num *big.Int, time uint64) bool) (hashes []common.Hash, err error)
```

core/blockchain.go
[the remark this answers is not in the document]
GalaIO: Yes, adding a sidecar needs another PR to implement, I just add writeBlobs logic. LRU cache is a cache that will drop cache in some scenarios, so it is not suitable for here.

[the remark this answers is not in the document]
GalaIO: Here I think it is more compatible to add sidecar cache to associate sidecar fetching, verification, and saving. Of course, other solutions can also be discussed.

core/rawdb/accessors_chain.go
```diff
@@ -809,6 +823,98 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+// WriteAncientBlocksWithBlobs writes entire block data into ancient store and returns the total written size.
+// Attention: The caller must set blobs after cancun
+func WriteAncientBlocksWithBlobs(db ethdb.AncientStore, blocks []*types.Block, receipts []types.Receipts, td *big.Int, blobs []types.BlobTxSidecars) (int64, error) {
+	if len(blocks) == 0 {
+		return 0, nil
+	}
+
+	// do some sanity check
+	if len(blocks) != len(blobs) {
+		return 0, fmt.Errorf("the blobs len is different with blobks, %v:%v", len(blobs), len(blocks))
```
zzzckck, line 835: typo: blobks

core/rawdb/freezer.go
```diff
@@ -167,6 +181,20 @@ func NewFreezer(datadir string, namespace string, readonly bool, offset uint64,
...
+	if readonly {
+		f, err := newTable(datadir, name, readMeter, writeMeter, sizeGauge, maxTableSize, disableSnappy, false)
+		if err != nil {
+			return nil, err
+		}
+		if err = f.Close(); err != nil {
+			return nil, err
+		}
+	}
+	return newTable(datadir, name, readMeter, writeMeter, sizeGauge, maxTableSize, disableSnappy, readonly)
```
zzzckck, line 195: seems like, we don't need the `if readonly ` check, can simply call:
f, err := newTable(datadir, name, readMeter, writeMeter, sizeGauge, maxTableSize, disableSnappy, false) ?

core/blockchain.go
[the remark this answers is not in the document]
emailtovamos: Ok, so the fetcher will have a new 'inserter' for sidecars/blobs and that will use this to store.

core/rawdb/freezer.go
GalaIO: There is an error for opening new table in read-only mode, the code takes a trick to open as write mode and close first, and open as read-only later.

core/blockchain.go
```diff
@@ -2034,6 +2021,20 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		if parent == nil {
 			parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 		}
+
+		// check blob data available first
+		// TODO(GalaIO): move IsDataAvailable combine into verifyHeaders?
+		if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
+			if posa, ok := bc.engine.(consensus.PoSA); ok {
+				blobs, exist := bc.receivedBlobsCache.Load(block.Hash())
+				if !exist {
```
emailtovamos, line 2030: Are we going to save empty blobs as well?

core/blockchain.go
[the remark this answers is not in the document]
emailtovamos: > LRU cache is a cache that will drop cache in some scenarios, so it is not suitable for here.

Ok, in that case, why not use `sync.Map` for `blobsCache` as well?

core/rawdb/accessors_chain.go
```diff
@@ -809,6 +823,98 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+		return nil
+	}
+	return ret
+}
+
+// WriteBlobs stores all the transaction blobs belonging to a block.
+// It could input nil for empty blobs.
+func WriteBlobs(db ethdb.KeyValueWriter, hash common.Hash, number uint64, blobs types.BlobTxSidecars) {
+	data, err := rlp.EncodeToBytes(blobs)
+	if err != nil {
```
emailtovamos, line 902: Since this function can have errors, we should have error return for `WriteBlobs()` so that we can act accordingly wherever it is used

core/rawdb/accessors_chain.go
```diff
@@ -809,6 +823,98 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+	// Store the flattened receipt slice
+	if err := db.Put(blockBlobsKey(number, hash), data); err != nil {
+		log.Crit("Failed to store block blobs", "err", err)
+	}
+}
+
+// DeleteBlobs removes all blob data associated with a block hash.
+func DeleteBlobs(db ethdb.KeyValueWriter, hash common.Hash, number uint64) {
+	if err := db.Delete(blockBlobsKey(number, hash)); err != nil {
+		log.Crit("Failed to delete block blobs", "err", err)
```
emailtovamos, line 914: Same as above, we should return the error.

emailtovamos (changes requested):

core/blockchain.go
GalaIO: Yes, empty blobs means not blobs, same as empty transactions or receipts.

core/blockchain.go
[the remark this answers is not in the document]
GalaIO: blobsCache is a query cache for query blobs, it allows cache loss.

core/rawdb/accessors_chain.go
GalaIO: This is the bad DB interface of geth, most DB function are not return err, but the `log.Crit` code will cause `os.Exit(1)`.

core/types/block.go
```diff
@@ -232,6 +232,9 @@ type Block struct {
 	// inter-peer block relay.
 	ReceivedAt   time.Time
 	ReceivedFrom interface{}
+
+	// blobs provides DA check
+	blobs BlobTxSidecars
```
emailtovamos, line 237: Weren't sidecars supposed to be together with block only for propagation purposes? Why it got added inside the block itself?

GalaIO: The block struct may be a good glue struct, we can set blobs into it, and transfer them to the related component. There are other extra fields in here to record metadata.

We also need another extra blob with the block for p2p messages.

emailtovamos: Just wondering if there's any process that is done with blocks passing apart from broadcasting. If there is any such process then it will slow down the work due to block becoming very heavy due to blobs. 

> We also need another extra blob with the block for p2p messages.

If we are including sidecar inside block definition itself then we won't need to make any changes for the p2p messages and the current mechanism only will need to check for Cancun or not and act accordingly.

GalaIO: Ok, we can discuss more about the definition. The block could transfer to `InsertChain`, `IsDataAvailbility` for checking block and blob, block is a glue struct, it contains header, block body, blobs, and other metadata. So we can remove receovedBlobCache too.

Block is passed by pointer, so it will cost little in most scenarios. It is only be used in `NewBlockMsg`, but only header, body will be encoded.

```go
func (b *Block) EncodeRLP(w io.Writer) error {
	return rlp.Encode(w, &extblock{
		Header:      b.header,
		Txs:         b.transactions,
		Uncles:      b.uncles,
		Withdrawals: b.withdrawals,
	})
}
```

emailtovamos: > It is only be used in NewBlockMsg, but only header, body will be encoded.

If we encode the sidecars as well then we may not even need to modify `BlockWithExtraData` and have this definition of block everywhere except when it comes to storing. As storing of sidecars and blocks will be separate. What do you think?

emailtovamos (approved):

PR #2234, opened 2024-02-23
cmd/utils: disable snap protocol for fast node

weiihann:
### Description
This PR adds a logic check such that if the node is fast node (i.e. `--tries-verify-mode` is not `local`), then it will disable the snap protocol by default.

### Rationale
When serving snap data (i.e. `AccountRange`, `StorageRange`, `TrieNodes`), access to a complete state trie is necessary. However, for fast nodes, since they only store the snapshot, they can't service these requests. Disabling this snap protocol by default may help reduce network calls and processing time for the nodes in general.

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -1953,6 +1953,11 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		if cfg.TriesVerifyMode.NeedRemoteVerify() {
 			cfg.EnableTrustProtocol = true
 		}
+		// A node without trie is not able to provide snap data, so it should disable snap protocol.
+		if cfg.TriesVerifyMode != core.LocalVerify {
+			log.Info("Automatically disables snap protocol due to verify mode", "mode", cfg.TriesVerifyMode)
+			cfg.DisableSnapProtocol = true
+		}
 
 		if cfg.SyncMode == downloader.SnapSync && cfg.TriesVerifyMode.NoTries() {
 			log.Warn("Only local TriesVerifyMode can support snap sync, resetting to full sync", "mode", cfg.TriesVerifyMode)


zzzckck (approved):
LGTM

zlacfzy (approved):

zlacfzy (dismissed):

zzzckck (approved):

cmd/utils/flags.go
```diff
@@ -1887,7 +1887,7 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 		cfg.DirectBroadcast = ctx.Bool(DirectBroadcastFlag.Name)
 	}
 	if ctx.IsSet(DisableSnapProtocolFlag.Name) {
-		cfg.DisableSnapProtocol = ctx.Bool(DisableSnapProtocolFlag.Name)
+		cfg.DisableSnapProtocol = cfg.DisableSnapProtocol || ctx.Bool(DisableSnapProtocolFlag.Name)
```
buddh0, line 1890: does the flag not have the prority now?

cmd/utils/flags.go
```diff
@@ -1954,6 +1954,12 @@ func SetEthConfig(ctx *cli.Context, stack *node.Node, cfg *ethconfig.Config) {
 			cfg.EnableTrustProtocol = true
 		}
 
+    // A node without trie is not able to provide snap data, so it should disable snap protocol.
```
buddh0, line 1957: if disable
fast node can do snap sync?

cmd/utils/flags.go
weiihann: The snap protocol will be disabled as long as one of these 2 conditions is satisfied:

1. `tries-verify-mode` != `local`
2. `--disablesnapprotocol` enabled

On that note, the additional logic check `cfg.DisableSnapProtocol` is not needed because then the `DisableSnapProtocol` in the config will supersede the one specified by the flag. I got confused with the ordering, reverted the changes.

cmd/utils/flags.go
weiihann: If you meant when a node starts with a fast node, which does full sync, then later on becomes a full node and does snap sync again, will it work? I'm not entirely sure, my intuition is that it doesn't work.

zzzckck (approved):

buddh0 (approved):

PR #2247, opened 2024-03-01
all: use errors.New() when there is no param rather than fmt.Errorf()

yasyzb:
### Description

There are many fmt.Errorf() with no param.

### Rationale

Parameterless errors should be using errors.New instead of fmt.Errorf.

### Example

```
errors.New("invalid hex data")
```

### Changes

Notable changes: 
* replace fmt.Errorf which has no param with errors.New


zzzckck (approved):

zlacfzy (approved):

zzzckck:
thx

PR #2260, opened 2024-03-07
blockchian: support to import block with blob & blobGasFee;

GalaIO:
### Description

The PR supports import block with blob, adds chasing head for DA checking, and supports side chain reimport.

It also modify some parameters for BSC 4844.

### Rationale

no

### Changes

Notable changes: 
* blob: implement min&max gas price logic;
* blob: change some key parameters;
* blockchian: support import side chain;
* blobpool: reject the banned address;
* blockchain: add chasing head for DA check;
* ...


core/data_availability.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/ethereum/go-ethereum/common/gopool"
+	"github.com/ethereum/go-ethereum/consensus"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto/kzg4844"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+)
+
+// validateBlobSidecar it is same as validateBlobSidecar in core/txpool/validation.go
+func validateBlobSidecar(hashes []common.Hash, sidecar *types.BlobTxSidecar) error {
```
emailtovamos, line 19: If it is the same, can we make the other one public and not have to write this function?

params/protocol_params.go
```diff
@@ -184,8 +184,8 @@ const (
 	BlobTxTargetBlobGasPerBlock = 3 * BlobTxBlobGasPerBlob // Target consumable blob gas for data blobs per block (for 1559-like pricing)
 	MaxBlobGasPerBlock          = 6 * BlobTxBlobGasPerBlob // Maximum consumable blob gas for data blobs per block
 
-	BlobReserveThreshold      = 18 * (24 * 3600) / 3 // it keeps blob data available for 18 days in local.
-	BlobExtraReserveThreshold = 1 * (24 * 3600) / 3  // it adds more time for expired blobs for some request cases, like expiry blob when remote peer is syncing, default 1 day.
+	MinBlocksForBlobRequests           uint64 = 4096 * 32 * 12 / 3  // it keeps blob data available for ~18.2 days in local.
```
emailtovamos, line 187: Can you write in the comment what do 4096, 32 and 12 represent for easier understanding?

core/blockchain.go
```diff
@@ -2006,15 +2030,6 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 			parent = bc.GetHeader(block.ParentHash(), block.NumberU64()-1)
 		}
 
-		// check blob data available first
-		// TODO(GalaIO): move IsDataAvailable combine into verifyHeaders?
-		if bc.chainConfig.IsCancun(block.Number(), block.Time()) {
-			if posa, ok := bc.engine.(consensus.PoSA); ok {
```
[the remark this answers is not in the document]
GalaIO: > Why is this removed?

I added `CheckDataAvailableInBatch` at 1869 line, it could check DA concurrently.

core/data_availability.go
GalaIO: I tried before, but it caused an import Cyclic dependence error. Copy the code is ok, and we can modify it independently.

params/protocol_params.go
GalaIO: ok, it ref the latest BEP changes.

emailtovamos (approved):

PR #2281, opened 2024-03-13
syncing: support to sync blobs and check DA;

GalaIO:
### Description

syncing: support to sync blobs and check DA;

### Changes

Notable changes: 
* execution: add blob gas fee reward to system;
* syncing: support blob syncing & DA checking;
* ...

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -475,3 +475,16 @@ func (d *Decimal) UnmarshalJSON(input []byte) error {
 		return err
 	}
 }
+
+// IsNil uses reflect to distinguish data that is truly nil.
+func IsNil(i interface{}) bool {
+	if i == nil {
+		return true
+	}
+
+	switch reflect.TypeOf(i).Kind() {
+	case reflect.Chan, reflect.Func, reflect.Map, reflect.Pointer, reflect.UnsafePointer, reflect.Interface, reflect.Slice:
+		return reflect.ValueOf(i).IsNil()
+	}
+	return false
+}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -279,7 +279,7 @@ type BlockChain struct {
 	receiptsCache *lru.Cache[common.Hash, []*types.Receipt]
 	blockCache    *lru.Cache[common.Hash, *types.Block]
 	txLookupCache *lru.Cache[common.Hash, txLookup]
-	blobsCache    *lru.Cache[common.Hash, types.BlobTxSidecars]
+	sidecarsCache *lru.Cache[common.Hash, types.BlobTxSidecars]
 
 	// future blocks are blocks added for later processing
 	futureBlocks *lru.Cache[common.Hash, *types.Block]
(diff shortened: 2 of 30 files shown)


core/blockchain.go
```diff
@@ -1336,6 +1336,14 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 		}
 	}
 
+	// check DA after cancun
+	lastBlk := blockChain[len(blockChain)-1]
+	if bc.chainConfig.Parlia != nil && !bc.chainConfig.IsCancun(lastBlk.Number(), lastBlk.Time()) {
```
emailtovamos, line 1341: If it is supposed to be after cancun, shouldn't it be without the `!` ?

core/rawdb/accessors_chain.go
```diff
@@ -823,6 +823,80 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+
+	// if there has pre-cancun and post-cancun blocks, write them separately
+	var (
+		writeSize  int64
+		tdSum      = new(big.Int).Set(td)
+		stReceipts []*types.ReceiptForStorage
+	)
+
+	// write pre-cancun blocks
+	if cancunIndex > 0 {
```
buddh0, line 851: no need to check

core/rawdb/accessors_chain.go
```diff
@@ -823,6 +823,80 @@ func WriteAncientBlocks(db ethdb.AncientWriter, blocks []*types.Block, receipts
...
+		preBlocks := blocks[:cancunIndex]
+		preReceipts := receipts[:cancunIndex]
+		size, err := db.ModifyAncients(func(op ethdb.AncientWriteOp) error {
+			for i, block := range preBlocks {
```
buddh0, line 855: 854~870 call `WriteAncientBlocks` instead?

core/state_transition.go
```diff
@@ -465,6 +465,13 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	// consensus engine is parlia
 	if st.evm.ChainConfig().Parlia != nil {
 		st.state.AddBalance(consensus.SystemAddress, fee)
+		// add extra blob fee reward
+		if rules.IsCancun {
+			blobFee := new(big.Int).SetUint64(st.blobGasUsed())
```
buddh0, line 470: use uint256.NewInt instead?

core/types/block.go
```diff
@@ -458,6 +458,12 @@ func (b *Block) Blobs() BlobTxSidecars {
 	return b.blobs
 }
 
+func (b *Block) CleanBlobs() {
+	if len(b.blobs) > 0 {
+		b.blobs = BlobTxSidecars{}
```
buddh0, line 463: b.blobs = nil

core/rawdb/accessors_chain.go
GalaIO: Oh, here is for sharing the TD.

buddh0 (approved):

common/types.go
```diff
@@ -475,3 +475,16 @@ func (d *Decimal) UnmarshalJSON(input []byte) error {
 		return err
 	}
 }
+
+// IsNil uses reflect to distinguish data that is truly nil.
```
emailtovamos, line 479: Is there any case where the data isn't truly nil? Just to know why this function was required.

emailtovamos (approved):

common/types.go
GalaIO: Some UTs send a `(evm.ChainContext)(nil)` to `eip4844.CalcBlobFee(*header.ExcessBlobGas, config)`, it will cause panic, if we just use `config == nil` to check nil.

PR #2286, opened 2024-03-14
Implement eth_getBlobSidecars && eth_getBlobSidecarByTxHash

zlacfzy:
### Description

Implemet two method eth_getBlobSidecars && eth_getBlobSidecarByTxHash

### Rationale

support get BlobSidecars

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/types/tx_blob.go
+++ b/core/types/tx_blob.go
@@ -65,9 +65,9 @@ func (s BlobTxSidecars) EncodeIndex(i int, w *bytes.Buffer) {
 
 // BlobTxSidecar contains the blobs of a blob transaction.
 type BlobTxSidecar struct {
-	Blobs       []kzg4844.Blob       // Blobs needed by the blob pool
-	Commitments []kzg4844.Commitment // Commitments needed by the blob pool
-	Proofs      []kzg4844.Proof      // Proofs needed by the blob pool
+	Blobs       []kzg4844.Blob       `json:"blobs"`       // Blobs needed by the blob pool
+	Commitments []kzg4844.Commitment `json:"commitments"` // Commitments needed by the blob pool
+	Proofs      []kzg4844.Proof      `json:"proofs"`      // Proofs needed by the blob pool
 }
 
 // BlobHashes computes the blob hashes of the given blobs.
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
 	return result, nil
 }
 
+func (s *BlockChainAPI) GetBlobSidecars(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]map[string]interface{}, error) {
+	block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
+		return nil, nil
+	}
+	blobSidecars, err := s.b.GetBlobSidecars(ctx, block.Hash())
+	if err != nil {
+		return nil, err
+	}
+	result := make([]map[string]interface{}, len(blobSidecars))
+	blobIndex := 0
+	for txIndex, tx := range block.Transactions() {
+		if tx.Type() == types.BlobTxType {
+			result[blobIndex] = marshalBlobSidecar(blobSidecars[blobIndex], block.Hash(), block.NumberU64(), tx, txIndex)
+			blobIndex++
+		}
+		if blobIndex >= len(blobSidecars) {
+			return nil, nil
+		}
+	}
+	return result, nil
+}
+
+func (s *BlockChainAPI) GetBlobSidecarByTxHash(ctx context.Context, hash common.Hash) (map[string]interface{}, error) {
+	tx, blockHash, blockNumber, Index := rawdb.ReadTransaction(s.b.ChainDb(), hash)
+	if tx == nil {
+		return nil, nil
+	}
+	block, err := s.b.BlockByHash(ctx, blockHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
+		return nil, nil
+	}
+	blobSidecars, err := s.b.GetBlobSidecars(ctx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+	blobIndex := 0
+	for txIndex, tx := range block.Transactions() {
+		if txIndex > int(Index) {
+			break
+		}
+		if tx.Type() == types.BlobTxType {
+			blobIndex++
+		}
+	}
+	result := marshalBlobSidecar(blobSidecars[blobIndex], blockHash, blockNumber, block.Transaction(hash), int(Index))
+	return result, nil
+}
+
 // OverrideAccount indicates the overriding fields of account during the execution
 // of a message call.
 // Note, state and stateDiff can't be specified at the same time. If state is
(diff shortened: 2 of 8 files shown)


internal/ethapi/api.go
```diff
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
 	return result, nil
 }
 
+func (s *BlockChainAPI) GetBlobSidecars(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]map[string]interface{}, error) {
+	block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
```
NathanBSC, line 1017: per?

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
...
+	}
+	result := make([]map[string]interface{}, len(blobSidecars))
+	blobIndex := 0
+	for txIndex, tx := range block.Transactions() {
+		if tx.Type() == types.BlobTxType {
+			result[blobIndex] = marshalBlobSidecar(blobSidecars[blobIndex], block.Hash(), block.NumberU64(), tx, txIndex)
+			blobIndex++
+		}
+		if blobIndex >= len(blobSidecars) {
+			return nil, nil
```
NathanBSC, line 1032: return nil, nil --> break

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
...
+	if err != nil {
+		return nil, err
+	}
+	blobIndex := 0
+	for txIndex, tx := range block.Transactions() {
+		if txIndex > int(Index) {
+			break
+		}
+		if tx.Type() == types.BlobTxType {
+			blobIndex++
```
NathanBSC, line 1059: if only one blobtx in the block
so blobIndex will 1, then toblobSidecars[blobIndex]? panic then

internal/ethapi/api.go
```diff
@@ -2105,6 +2158,17 @@ func marshalReceipt(receipt *types.Receipt, blockHash common.Hash, blockNumber u
 	return fields
 }
 
+func marshalBlobSidecar(blobSidecar *types.BlobTxSidecar, blockHash common.Hash, blockNumber uint64, tx *types.Transaction, txIndex int) map[string]interface{} {
+	fields := map[string]interface{}{
+		"blockHash":        blockHash,
+		"blockNumber":      hexutil.Uint64(blockNumber),
+		"transactionHash":  tx.Hash(),
```
NathanBSC, line 2165: keep same with other interface
transactionHash-->txHash
transactionIndex-->txIndex

core/types/tx_blob.go
```diff
@@ -65,9 +65,9 @@ func (s BlobTxSidecars) EncodeIndex(i int, w *bytes.Buffer) {
 
 // BlobTxSidecar contains the blobs of a blob transaction.
 type BlobTxSidecar struct {
-	Blobs       []kzg4844.Blob       // Blobs needed by the blob pool
-	Commitments []kzg4844.Commitment // Commitments needed by the blob pool
-	Proofs      []kzg4844.Proof      // Proofs needed by the blob pool
+	Blobs       []kzg4844.Blob       `json:"blobs"`       // Blobs needed by the blob pool
```
NathanBSC, line 68: when marshal
need to transfer to hex? otherwise may not suitable for human to see

not quite sure, just check

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
 	return result, nil
 }
 
+func (s *BlockChainAPI) GetBlobSidecars(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) ([]map[string]interface{}, error) {
+	block, err := s.b.BlockByNumberOrHash(ctx, blockNrOrHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
+		return nil, nil
```
NathanBSC, line 1018: return nil, nil-->return nil, err
don't swallow the error

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,59 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
...
+func (s *BlockChainAPI) GetBlobSidecarByTxHash(ctx context.Context, hash common.Hash) (map[string]interface{}, error) {
+	tx, blockHash, blockNumber, Index := rawdb.ReadTransaction(s.b.ChainDb(), hash)
+	if tx == nil {
+		return nil, nil
+	}
+	block, err := s.b.BlockByHash(ctx, blockHash)
+	if block == nil || err != nil {
+		// When the block doesn't exist, the RPC method should return JSON null
+		// as per specification.
+		return nil, nil
```
NathanBSC, line 1047: return nil, nil-->return nil, err
don't swallow the error

internal/ethapi/api_test.go
```diff
@@ -563,6 +563,15 @@ func (b testBackend) GetReceipts(ctx context.Context, hash common.Hash) (types.R
...
+
+func (b testBackend) GetBlobSidecars(ctx context.Context, hash common.Hash) (types.BlobTxSidecars, error) {
+	header, err := b.HeaderByHash(ctx, hash)
+	if header == nil || err != nil {
+		return nil, err
+	}
+	blobSidecars := rawdb.ReadRawBlobs(b.db, hash, header.Number.Uint64())
+	return blobSidecars, nil
+}
 func (b testBackend) GetTd(ctx context.Context, hash common.Hash) *big.Int {
```
buddh0, line 575: can we add some unit tests for these two apis?

internal/ethapi/api.go
zzzckck: it is ok to follow the definition of `GetBlockReceipts()`, when failed to get Block, just return nil, nil

internal/ethapi/api.go
zzzckck: `if blobIndex >= len(blobSidecars) {}`
it should never happen, otherwise there could be a bug, so better to print an error message and return nil, err

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,60 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
...
+	if err != nil || len(blobSidecars) == 0 {
+		return nil, nil
+	}
+	blobIndex := -1
+	for txIndex, tx := range block.Transactions() {
+		if tx.Type() == types.BlobTxType {
+			if txIndex <= int(Index) {
+				blobIndex++
+			} else {
+				return nil, fmt.Errorf("blobSidecars length mismatch %d", len(blobSidecars))
```
zzzckck, line 1059: should break here, not return? Say if `Index == 0` and there are 2+ blobTx in this block.

internal/ethapi/api.go
```diff
@@ -1010,6 +1010,60 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
...
+			}
+		}
+	}
+	result := marshalBlobSidecar(blobSidecars[blobIndex], blockHash, blockNumber, block.Transaction(hash), int(Index))
```
zzzckck, line 1063: we don't need call `block.Transaction(hash)` here, as we already have the tx at the beginning of the function, just rename it with txTarget should be ok.

zzzckck (approved):

core/types/tx_blob.go
zlacfzy: yes, need to transfer to hex. Refer to this: https://ethereum.github.io/beacon-APIs/#/Beacon/getBlobSidecars

buddh0 (approved):

PR #2292, opened 2024-03-15
da: support pruneancient nodes;

GalaIO:
### Description

da: support pruneancient nodes;

### Changes

Notable changes: 
* da: support pruneancient nodes;
* ...

### DIFF
--- a/core/data_availability.go
+++ b/core/data_availability.go
@@ -64,6 +64,13 @@ func IsDataAvailable(chain consensus.ChainHeaderReader, block *types.Block) (err
 		return nil
 	}
 
+	sidecars := block.Sidecars()
+	// there has a very special case that nodes only keep 90000 recent blocks/blobs when it enables --pruneancient flag.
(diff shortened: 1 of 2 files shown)


core/data_availability.go
```diff
@@ -64,6 +64,13 @@ func IsDataAvailable(chain consensus.ChainHeaderReader, block *types.Block) (err
 		return nil
 	}
 
+	sidecars := block.Sidecars()
+	// there has a very special case that nodes only keep 90000 recent blocks/blobs when it enables --pruneancient flag.
```
emailtovamos, line 68: so if `pruneancient` is set, the blocks older than 90000 will also be deleted? Along with blobs.

NathanBSC, line 68: need check flag here？

GalaIO:
Oh, this PR needs to be closed, because pruneancient nodes cannot provide ancient blocks/blobs for syncing, so it don't need the special logic for DA checking.

PR #2302, opened 2024-03-18
cmd/geth, core/rawdb: add dbDeleteTrieState

weiihann:
### Description
This PR adds a feature to convert full node to fast node by introducing a new CLI tool called `delete-trie-state`. The tool will delete all PBSS key-value pairs and ancient state stores from the database.

### Rationale
This is particularly useful when users want to switch from full node to fast node by deleting the trie state from the database, reducing disk space occupied.

### Example
Usage: `./geth --datadir db delete-trie-state`
```
...
INFO [03-18|16:06:39.783] Opened ancient database                  database=/datadir_test/geth/chaindata/ancient/chain readonly=false frozen=909,979
INFO [03-18|16:06:43.759] Deleted trie state                       count=1,683,851 elapsed=3.971s
INFO [03-18|16:06:43.759] Removing ancient state database          path=/datadir_test/geth/chaindata/ancient/state
INFO [03-18|16:06:43.862] State database successfully deleted      path=/datadir_test/geth/chaindata/ancient/state elapsed=103.202ms
```

### Changes
Also added an additional `NoTries` logic check in `SetupGenesisBlockWithOverride` to prevent committing the genesis block again. For custom networks, this will result in an error because `!triedb.Initialized(header.Root)` returns true due to an empty trie database, so we need to skip committing if it's a fast node.

### DIFF
--- a/cmd/geth/dbcmd.go
+++ b/cmd/geth/dbcmd.go
@@ -76,6 +76,7 @@ Remove blockchain and state databases`,
 			dbCompactCmd,
 			dbGetCmd,
 			dbDeleteCmd,
+			dbDeleteTrieStateCmd,
 			dbInspectTrieCmd,
 			dbPutCmd,
 			dbGetSlotsCmd,
@@ -205,6 +206,15 @@ corruption if it is aborted during execution'!`,
 		}, utils.NetworkFlags, utils.DatabaseFlags),
 		Description: `This command deletes the specified database key from the database.
 WARNING: This is a low-level operation which may cause database corruption!`,
+	}
+	dbDeleteTrieStateCmd = &cli.Command{
+		Action: dbDeleteTrieState,
+		Name:   "delete-trie-state",
+		Usage:  "Delete all trie state key-value pairs from the database and the ancient state. Does not support hash-based state scheme.",
+		Flags: flags.Merge([]cli.Flag{
+			utils.SyncModeFlag,
+		}, utils.NetworkFlags, utils.DatabaseFlags),
+		Description: `This command deletes all trie state key-value pairs from the database and the ancient state.`,
 	}
 	dbPutCmd = &cli.Command{
 		Action:    dbPut,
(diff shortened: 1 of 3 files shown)


core/rawdb/database.go
```diff
@@ -887,6 +887,61 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
...
+
+	prefixKeys := map[string]func([]byte) bool{
+		string(trieNodeAccountPrefix): IsAccountTrieNode,
+		string(trieNodeStoragePrefix): IsStorageTrieNode,
+		string(stateIDPrefix):         func(key []byte) bool { return len(key) == len(stateIDPrefix)+common.HashLength },
+	}
+
+	for prefix, isValid := range prefixKeys {
+		it = db.NewIterator([]byte(prefix), nil)
+
```
flywukong, line 908: defer  it.Release()

cmd/geth/dbcmd.go
```diff
@@ -810,6 +820,74 @@ func dbDelete(ctx *cli.Context) error {
...
+				}
+				if !fileInfo.IsDir() {
+					os.Remove(path)
+				}
+			}
+			return nil
+		})
+		return err
+	} else {
+		err = rawdb.DeleteTrieState(db)
```
flywukong, line 855: it is better to delete the kv data in state by API `if db.StateStore() != nil  { 
  1)err = rawdb.DeleteTrieState(db.db.StateStore())
   2)...remove the ancient state
} else {
 err = rawdb.DeleteTrieState(db)
}`

core/rawdb/database.go
weiihann: We shouldn't do `defer it.Release()` as it may lead to memory leak issues. At the end of each iteration, I added `it.Release()`. But I forgot to add it when `batch.Write` returns an err. Already added it in the new changes.

cmd/geth/dbcmd.go
weiihann: Changed.

cmd/geth/dbcmd.go
```diff
@@ -810,6 +820,61 @@ func dbDelete(ctx *cli.Context) error {
...
+	}
+
+	stack, config := makeConfigNode(ctx)
+	defer stack.Close()
+
+	db := utils.MakeChainDatabase(ctx, stack, false, false)
+	defer db.Close()
+
+	var err error
+	if db.StateStore() != nil { // if separate trie db exists, delete all files in the db folder
```
Fynnss, line 836: called rawdb.DeleteTrieState(db) directly and check whether the stateStore is nil in DeleteTrieState function.

cmd/geth/dbcmd.go
weiihann: No need to revert KV data one by one if it's separate db. Reverted.

zzzckck (approved):

flywukong (approved):
LGTM

PR #2316, opened 2024-03-21
fix: fix some validation issues & ancient reset;

GalaIO:
### Description

fix: fix some validation issues & ancient reset;

### Changes

Notable changes: 
* fix: fix some validation issues & ancient reset;
* ...

### DIFF
--- a/core/rawdb/freezer.go
+++ b/core/rawdb/freezer.go
@@ -344,8 +344,13 @@ func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
 	}
 	for kind, table := range f.tables {
 		err := table.truncateHead(items - f.offset)
-		// if truncate below tail, try reset to it
 		if err == errTruncationBelowTail {
+			// This often happens in chain rewinds, but the blob table is special.
+			// It has the same head, but a different tail from other tables (like bodies, receipts).
+			// So if the chain is rewound to head below the blob's tail, it needs to reset again.
+			if kind != ChainFreezerBlobSidecarTable {
+				return 0, err
+			}
 			nt, err := table.resetItems(items-f.offset, items-f.offset)
 			if err != nil {
 				return 0, err
@@ -480,8 +485,13 @@ func (f *Freezer) repair() error {
 			continue
 		}
 		err := table.truncateHead(head)
-		// if truncate below tail, try reset to it
 		if err == errTruncationBelowTail {
+			// This often happens in chain rewinds, but the blob table is special.
+			// It has the same head, but a different tail from other tables (like bodies, receipts).
+			// So if the chain is rewound to head below the blob's tail, it needs to reset again.
+			if kind != ChainFreezerBlobSidecarTable {
+				return err
+			}
 			nt, err := table.resetItems(head, head)
 			if err != nil {
 				return err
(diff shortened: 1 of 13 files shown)


core/rawdb/freezer_table.go
```diff
@@ -406,7 +408,7 @@ func (t *freezerTable) truncateHead(items uint64) error {
 		return nil
 	}
 	if items < t.itemHidden.Load() {
-		return fmt.Errorf("truncation below tail, name: %v, tail: %v, head: %v, truncate to: %v", t.name, t.itemHidden.Load(), t.items.Load(), items)
+		return errTruncationBelowTail
```
buddh0, line 411: lost some useful info after modifying?

core/rawdb/freezer.go
```diff
@@ -342,8 +342,16 @@ func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
 	if oitems <= items {
 		return oitems, nil
 	}
-	for _, table := range f.tables {
-		if err := table.truncateHead(items - f.offset); err != nil {
+	for kind, table := range f.tables {
+		err := table.truncateHead(items - f.offset)
+		// if truncate below tail, try reset to it
+		if err == errTruncationBelowTail {
+			nt, err := table.resetItems(items-f.offset, items-f.offset)
```
emailtovamos, line 349: This resetting was the only place with the bug or was there any other place as well?

GalaIO: It will be triggered when the chain rewinds to genesis, it will truncate the head till genesis, but the blob table is special, it has the same head, but a different tail from other tables (like bodies, receipts).

So if the chain is rewound to head below the blob's tail, it needs to reset again.

core/rawdb/freezer.go
```diff
@@ -342,8 +342,16 @@ func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
...
+			if err != nil {
+				return 0, err
+			}
+			f.tables[kind] = nt
+		} else {
 			return 0, err
```
zzzckck, line 355: should be:?
else if err != nil {
  return 0, err
}

core/rawdb/freezer.go
```diff
@@ -469,7 +477,15 @@ func (f *Freezer) repair() error {
...
-		if err := table.truncateHead(head); err != nil {
+		err := table.truncateHead(head)
+		// if truncate below tail, try reset to it
+		if err == errTruncationBelowTail {
+			nt, err := table.resetItems(head, head)
+			if err != nil {
+				return err
+			}
+			f.tables[kind] = nt
+		} else {
```
zzzckck, line 488: same:
else if err != nil {
  return err
}

zzzckck (approved):

core/rawdb/freezer.go
```diff
@@ -342,8 +342,18 @@ func (f *Freezer) TruncateHead(items uint64) (uint64, error) {
 	if oitems <= items {
 		return oitems, nil
 	}
-	for _, table := range f.tables {
-		if err := table.truncateHead(items - f.offset); err != nil {
+	for kind, table := range f.tables {
+		err := table.truncateHead(items - f.offset)
+		// if truncate below tail, try reset to it
+		if err == errTruncationBelowTail {
```
buddh0, line 348: can we do resetItems for sidecars
other table keep as before？
It's a more safe way

buddh0 (approved):

emailtovamos (approved):

PR #2322, opened 2024-03-22
chore: fix typos

xiaoxianBoy:
Fix: minor typo


crypto/secp256k1/libsecp256k1/src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -24,6 +24,9 @@
 import com.google.common.base.Preconditions;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.bitcoin.NativeSecp256k1Util.AssertFailException;
```
zzzckck, line 28: why import this package?

Pjrich1313: Please don't import Eugene is they only one allowed to make changes on encoder for my main server please don't make me block you

PR #2333, opened 2024-03-25
remove code that will not be executed

sellskin:
### Description

The program will exit directly after Fatal, the subsequent code will not be executed.

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck (approved):

emailtovamos (approved):

weiihann:
Please fix the checks

PR #2359, opened 2024-04-01
triedb: do not open state freezer under notries

weiihann:
### Description
This PR fixes a bug on v1.4.x BSC version where a pruned fast node cannot be initialized properly.

### Rationale
State freezer is introduced in PBSS. When a full node is converted to a fast node (i.e. via prune state), state freezer is no longer needed. However, the current implementation still opens a state freezer even though it's a fast node, which immediately throws an error and causes the node to not be able to start at all. This fix adds a rule such that if the configuration states that `NoTries = true`, then the node doesn't need to open the state freezer.

### DIFF
--- a/triedb/database.go
+++ b/triedb/database.go
@@ -116,6 +116,9 @@ func NewDatabase(diskdb ethdb.Database, config *Config) *Database {
 			config.HashDB = hashdb.Defaults
 		}
 	}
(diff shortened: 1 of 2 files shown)


Fynnss (approved):
LGTM

zzzckck (approved):

PR #2363, opened 2024-04-02
 feat: greedy merge tx in bid

irrun:
### Description

1. enable greedy merge tx in bid
2. fix several bugs

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -506,8 +506,8 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		builder     = bidRuntime.bid.Builder
 
 		bidTxs   = bidRuntime.bid.Txs
-		bidLen   = len(bidTxs)
-		payBidTx = bidTxs[bidLen-1]
+		bidTxLen = len(bidTxs)
+		payBidTx = bidTxs[bidTxLen-1]
 
 		err     error
 		success bool
@@ -582,7 +582,7 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		default:
 		}
 
-		if bidRuntime.env.tcount == bidLen-1 {
+		if bidRuntime.env.tcount == bidTxLen-1 {
 			break
 		}
 
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -621,7 +621,10 @@ func (b *bidSimulator) reportIssue(bidRuntime *BidRuntime, err error) {
 			BidHash:   bidRuntime.bid.Hash(),
 			Message:   err.Error(),
 		})
-		log.Error("BidSimulator: failed to report issue", "builder", bidRuntime.bid.Builder, "err", err)
+
+		if err != nil {
+			log.Error("BidSimulator: failed to report issue", "builder", bidRuntime.bid.Builder, "err", err)
+		}
 	}
 }
 
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -615,11 +615,13 @@ func (b *bidSimulator) reportIssue(bidRuntime *BidRuntime, err error) {
 
 	cli := b.builders[bidRuntime.bid.Builder]
 	if cli != nil {
-		cli.ReportIssue(context.Background(), &types.BidIssue{
+		err = cli.ReportIssue(context.Background(), &types.BidIssue{
 			Validator: bidRuntime.env.header.Coinbase,
 			Builder:   bidRuntime.bid.Builder,
+			BidHash:   bidRuntime.bid.Hash(),
 			Message:   err.Error(),
 		})
+		log.Error("BidSimulator: failed to report issue", "builder", bidRuntime.bid.Builder, "err", err)
 	}
 }
 
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -311,6 +311,8 @@ func (b *bidSimulator) newBidLoop() {
 
 	// commit aborts in-flight bid execution with given signal and resubmits a new one.
 	commit := func(reason int32, bidRuntime *BidRuntime) {
+		log.Debug("BidSimulator: start", "bidHash", bidRuntime.bid.Hash().Hex())
+
 		// if the left time is not enough to do simulation, return
 		var simDuration time.Duration
 		if lastBid := b.GetBestBid(bidRuntime.bid.ParentHash); lastBid != nil && lastBid.duration != 0 {
@@ -318,6 +320,7 @@ func (b *bidSimulator) newBidLoop() {
 		}
 
 		if time.Until(b.bidMustBefore(bidRuntime.bid.ParentHash)) <= simDuration*leftOverTimeRate/leftOverTimeScale {
+			log.Debug("BidSimulator: abort commit, not enough time to simulate", "bidHash", bidRuntime.bid.Hash().Hex())
 			return
 		}
 
@@ -349,6 +352,7 @@ func (b *bidSimulator) newBidLoop() {
 
 			if expectedValidatorReward.Cmp(big.NewInt(0)) < 0 {
 				// damage self profit, ignore
+				log.Debug("BidSimulator: invalid bid, validator reward is less than 0, ignore", "bidHash", newBid.Hash().Hex())
 				continue
 			}
 
@@ -378,6 +382,7 @@ func (b *bidSimulator) newBidLoop() {
 					continue
 				}
 
+				log.Debug("BidSimulator: lower reward, ignore", "bidHash", newBid.Hash().Hex())
 				continue
 			}
 
@@ -389,6 +394,7 @@ func (b *bidSimulator) newBidLoop() {
 				continue
 			}
 
+			log.Debug("BidSimulator: lower reward, ignore", "bidHash", newBid.Hash().Hex())
 		case <-b.exitCh:
 			return
 		}
@@ -535,7 +541,7 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 
 		if err != nil {
 			logCtx = append(logCtx, "err", err)
-			log.Debug("bid simulation failed", logCtx...)
+			log.Info("BidSimulator: simulation failed", logCtx...)
 
 			go b.reportIssue(bidRuntime, err)
 		}
@@ -606,7 +612,7 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 	if b.config.GreedyMergeTx {
 		delay := b.engine.Delay(b.chain, bidRuntime.env.header, &b.delayLeftOver)
 		if delay != nil && *delay > 0 {
-			log.Debug("BidSimulator: GreedyMergeTx stopTimer", "block", bidRuntime.env.header.Number,
+			log.Debug("BidSimulator: greedy merge tx stopTimer", "block", bidRuntime.env.header.Number,
 				"header time", time.Until(time.Unix(int64(bidRuntime.env.header.Time), 0)),
 				"commit delay", *delay, "DelayLeftOver", b.delayLeftOver)
 
@@ -618,9 +624,8 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 			}
 
 			fillErr := b.bidWorker.fillTransactions(interruptCh, bidRuntime.env, stopTimer, bidTxsSet)
-			if fillErr != nil {
-				log.Debug("BidSimulator: GreedyMergeTx fillTransactions", "block", bidRuntime.env.header.Number, "err", fillErr)
-			}
+			log.Info("BidSimulator: greedy merge tx fill transactions", "block", bidRuntime.env.header.Number,
+				"tx count", bidRuntime.env.tcount-bidTxLen+1, "err", fillErr)
 
 			// recalculate the packed reward
 			bidRuntime.packReward(b.config.ValidatorCommission)
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -11,6 +11,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	mapset "github.com/deckarep/golang-set/v2"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/bidutil"
 	"github.com/ethereum/go-ethereum/consensus"
@@ -61,9 +63,10 @@ var (
 	}
 )
 
-type WorkPreparer interface {
+type bidWorker interface {
 	prepareWork(params *generateParams) (*environment, error)
 	etherbase() common.Address
+	fillTransactions(interruptCh chan int32, env *environment, stopTimer *time.Timer, bidTxs mapset.Set[common.Hash]) (err error)
 }
 
 // simBidReq is the request for simulating a bid
@@ -79,7 +82,8 @@ type bidSimulator struct {
 	delayLeftOver time.Duration
 	chain         *core.BlockChain
 	chainConfig   *params.ChainConfig
-	workPreparer  WorkPreparer
+	engine        consensus.Engine
+	bidWorker     bidWorker
 
 	running atomic.Bool // controlled by miner
 	exitCh  chan struct{}
@@ -112,16 +116,18 @@ type bidSimulator struct {
 func newBidSimulator(
 	config *MevConfig,
 	delayLeftOver time.Duration,
-	chainConfig *params.ChainConfig,
 	chain *core.BlockChain,
-	workPreparer WorkPreparer,
+	chainConfig *params.ChainConfig,
+	engine consensus.Engine,
+	bidWorker bidWorker,
 ) *bidSimulator {
 	b := &bidSimulator{
 		config:        config,
 		delayLeftOver: delayLeftOver,
-		chainConfig:   chainConfig,
 		chain:         chain,
-		workPreparer:  workPreparer,
+		chainConfig:   chainConfig,
+		engine:        engine,
+		bidWorker:     bidWorker,
 		exitCh:        make(chan struct{}),
 		chainHeadCh:   make(chan core.ChainHeadEvent, chainHeadChanSize),
 		builders:      make(map[common.Address]*builderclient.Client),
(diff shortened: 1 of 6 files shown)


miner/bid_simulator.go
```diff
@@ -593,6 +610,34 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		return
 	}
 
+	// fill transactions from mempool
+	if b.config.AppendToBid {
```
zjubfd, line 614: AppendToBid seems not a good name

miner/bid_simulator.go
```diff
@@ -692,5 +742,261 @@ func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *para
 		env.receipts = append(env.receipts, receipt)
 	}
 
+	r.env.tcount++
+
 	return nil
 }
+
+// fillTransactions fill pending transactions from the txpool into the bid.
+func (r *BidRuntime) fillTransactions(
```
zjubfd, line 751: It that possible that we dont have to copy so many code?

miner/bid_simulator.go
```diff
@@ -692,5 +742,261 @@ func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *para
...
+
+	if len(remotePlainTxs) > 0 || len(remoteBlobTxs) > 0 {
+		plainTxs := newTransactionsByPriceAndNonce(env.signer, remotePlainTxs, env.header.BaseFee)
+		blobTxs := newTransactionsByPriceAndNonce(env.signer, remoteBlobTxs, env.header.BaseFee)
+
+		r.commitTransactions(interruptCh, stopTimer, prefetcher, vmConfig, chain, chainConfig, plainTxs, blobTxs)
+	}
+}
+
+func (r *BidRuntime) commitTransactions(
```
zjubfd, line 777: so is this part.

miner/bid_simulator.go
```diff
@@ -692,5 +742,261 @@ func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *para
...
+		default:
+			// Transaction is regarded as invalid, drop all consecutive transactions from
+			// the same sender because of `nonce-too-high` clause.
+			log.Debug("BidSimulator: Transaction failed, account skipped", "hash", ltx.Hash, "err", err)
+			txs.Pop()
+		}
+	}
+}
+
+func (b *bidSimulator) pendingTxs(
```
zjubfd, line 900: so is this part.

miner/bid_simulator.go
irrun: how about enlargeBid?

miner/bid_simulator.go
irrun: fixed

miner/bid_simulator.go
irrun: fixed

miner/bid_simulator.go
irrun: fixed

miner/bid_simulator.go
zjubfd: I dont have good idea either... What about `LocalTxMerge`, `GreedyMergeTx`?

go.sum
```diff
@@ -405,6 +405,8 @@ github.com/elastic/gosigar v0.14.2/go.mod h1:iXRIGg2tLnu7LBdpqzyQfGDEidKCfWcCMS0
 github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
 github.com/emicklei/dot v0.11.0/go.mod h1:DeV7GvQtIw4h2u73RKBkkFdvVAz0D9fzeJrgPW6gy/s=
 github.com/emicklei/go-restful v0.0.0-20170410110728-ff4f55a20633/go.mod h1:otzb+WCGbkyDHkqmQmT5YD2WR4BBwUdeQoFo8l/7tVs=
+github.com/emirpasic/gods v1.18.1 h1:FXtiHYKDGKCW2KzwZKx0iC0PQmdlorYgdFG9jPXJ1Bc=
```
zjubfd, line 408: Please dont add go.mod and go.sum it if is not necessary for this PR.

miner/bid_simulator.go
```diff
@@ -34,6 +36,8 @@ const (
 	leftOverTimeRate = 11
 	// leftOverTimeScale is the scale of left over time to simulate a bid
 	leftOverTimeScale = 10
+
+	PayBidTxGas uint64 = 21000 // The gas reserved for pay bid tx
```
zjubfd, line 40: can you check the code of builder, I think it is 25000.

miner/bid_simulator.go
```diff
@@ -64,6 +68,7 @@ var (
 type WorkPreparer interface {
 	prepareWork(params *generateParams) (*environment, error)
 	etherbase() common.Address
+	fillTransactions(interruptCh chan int32, env *environment, stopTimer *time.Timer, bidTxs *hashset.Set) (err error)
```
zjubfd, line 71: I think it is not proper to call it `WorkPreparer` any longer as it has the `fillTransactions` interface.

miner/bid_simulator.go
```diff
@@ -500,8 +506,13 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		blockNumber = bidRuntime.bid.BlockNumber
 		parentHash  = bidRuntime.bid.ParentHash
 		builder     = bidRuntime.bid.Builder
-		err         error
-		success     bool
+
+		bidTxs   = bidRuntime.bid.Txs
+		bidLen   = len(bidTxs)
+		payBidTx = bidTxs[bidLen-1]
```
zjubfd, line 512: I am a bit confused here, the API level allow the no payBidTx, however here it assume that it must have the payBidTx.  We must maintain consistency.

miner/bid_simulator.go
```diff
@@ -572,17 +583,16 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		default:
 		}
 
-		// Start executing the transaction
-		bidRuntime.env.state.SetTxContext(tx.Hash(), bidRuntime.env.tcount)
```
zjubfd, line 576: why remove this? not necessary?

miner/bid_simulator.go
```diff
@@ -572,17 +583,16 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+		}
 
 		err = bidRuntime.commitTransaction(b.chain, b.chainConfig, tx)
 		if err != nil {
 			log.Error("BidSimulator: failed to commit tx", "bidHash", bidRuntime.bid.Hash(), "tx", tx.Hash(), "err", err)
 			err = fmt.Errorf("invalid tx in bid, %v", err)
 			return
 		}
-
-		bidRuntime.env.tcount++
```
zjubfd, line 585: why remove this?

miner/bid_simulator.go
```diff
@@ -593,6 +603,41 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+	if b.config.EnlargeBid {
+		delay := b.engine.Delay(b.chain, bidRuntime.env.header, &b.delayLeftOver)
+		if delay != nil || *delay > 0 {
+			log.Debug("BidSimulator: stopTimer", "block", bidRuntime.env.header.Number,
+				"header time", time.Until(time.Unix(int64(bidRuntime.env.header.Time), 0)),
+				"commit delay", *delay, "DelayLeftOver", b.delayLeftOver)
+
+			stopTimer := time.NewTimer(0)
+			defer stopTimer.Stop()
+			<-stopTimer.C // discard the initial tick
```
zjubfd, line 616: why not `stopTimer := time.NewTimer(*delay)`?

miner/bid_simulator.go
```diff
@@ -593,6 +603,41 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+			stopTimer.Reset(*delay)
+
+			bidTxsSet := hashset.New()
+			for _, tx := range bidRuntime.bid.Txs {
+				bidTxsSet.Add(tx.Hash())
+			}
+
+			bidRuntime.env.gasPool.SubGas(PayBidTxGas)
+
+			_ = b.workPreparer.fillTransactions(interruptCh, bidRuntime.env, stopTimer, bidTxsSet)
```
zjubfd, line 627: shall we ignore the error here?

miner/bid_simulator.go
```diff
@@ -593,6 +603,41 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+			<-stopTimer.C // discard the initial tick
+
+			stopTimer.Reset(*delay)
+
+			bidTxsSet := hashset.New()
+			for _, tx := range bidRuntime.bid.Txs {
+				bidTxsSet.Add(tx.Hash())
+			}
+
+			bidRuntime.env.gasPool.SubGas(PayBidTxGas)
```
zjubfd, line 625: The side effect is that, even `SubGas` here and the gasPool is possible go to be zero eventually, the `bidRuntime.commitTransaction(b.chain, b.chainConfig, payBidTx)` will fail too.

miner/worker.go
```diff
@@ -1080,6 +1081,32 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
...
+		filterBidTxs := func(commonTxs map[common.Address][]*txpool.LazyTransaction) {
+			defer wg.Done()
+			for acc, txs := range commonTxs {
+				newTxs := make([]*txpool.LazyTransaction, 0)
+				for _, tx := range txs {
+					if !bidTxs.Contains(tx.Hash) {
+						newTxs = append(newTxs, tx)
+					}
+				}
+				commonTxs[acc] = newTxs
```
zjubfd, line 1098: I think in most cases the txhash will not conflict, we dont have to do `commonTxs[acc] = newTxs` for each account. Just assign only if necessary?

miner/worker.go
```diff
@@ -1080,6 +1081,32 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
...
+		var wg sync.WaitGroup
+		wg.Add(4)
+
+		filterBidTxs := func(commonTxs map[common.Address][]*txpool.LazyTransaction) {
+			defer wg.Done()
+			for acc, txs := range commonTxs {
+				newTxs := make([]*txpool.LazyTransaction, 0)
+				for _, tx := range txs {
+					if !bidTxs.Contains(tx.Hash) {
+						newTxs = append(newTxs, tx)
```
zjubfd, line 1095: you can not simply remove the iteam in the slice..

What does tx conflict means:
if tx hash is same, it means the nonce of the tx is same,  then all the tx before the conflict tx from a same account is invalid. So once a conflict tx find, we need remove all the txs that is queued before this tx.

miner/worker.go
```diff
@@ -1080,6 +1081,32 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
...
+					}
+				}
+				commonTxs[acc] = newTxs
+			}
+		}
+
+		go filterBidTxs(localPlainTxs)
```
zjubfd, line 1102: I dont think there is a need to start 4 Go routine:
1. Blob Txs is quite few on the network;
2. local Txs is quite few or none for a validator;

so only `filterBidTxs(remotePlainTxs)` will cost most of the time.

miner/bid_simulator.go
irrun: ok, GreedyMergeTx

miner/bid_simulator.go
irrun: fixed

miner/bid_simulator.go
irrun: bidWorker now

miner/bid_simulator.go
irrun: the API level does not allow no payBidTx for now

miner/bid_simulator.go
irrun: fixed

miner/bid_simulator.go
irrun: errors may happen here: 1)interrupt by better bid;2)out of gas;3)timeout for commit more tx; all can be ignored; 
added log.

miner/bid_simulator.go
irrun: fixed, add minGas in the params of fillTransactions

miner/worker.go
irrun: ok

miner/worker.go
irrun: fixed

miner/worker.go
irrun: fixed

go.sum
irrun: removed

miner/bid_simulator.go
```diff
@@ -593,6 +601,38 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+				"commit delay", *delay, "DelayLeftOver", b.delayLeftOver)
+
+			stopTimer := time.NewTimer(*delay)
+
+			bidTxsSet := mapset.NewSet[common.Hash]()
+			for _, tx := range bidRuntime.bid.Txs {
+				bidTxsSet.Add(tx.Hash())
+			}
+
+			fillErr := b.bidWorker.fillTransactions(interruptCh, bidRuntime.env, stopTimer, bidTxsSet, params.TxGas+params.PayBidTxGasLimit)
```
zjubfd, line 619: both `params.TxGas` and `params.PayBidTxGasLimit` are constant, so I think no need pass it as a parameter, it is a long way to go.  

And the patch does not fix the problem:
```
		if env.gasPool.Gas() < minGas {
			log.Trace("Not enough gas for further transactions", "have", env.gasPool, "want", minGas)
			signal = commitInterruptOutOfGas
			break
		}
```
It is still possible that `env.gasPool.Gas()` is smaller than 25000

miner/worker.go
zjubfd: emm, the current implementation get too many duplicated code, pease do optimization.

miner/worker.go
```diff
@@ -1080,6 +1082,65 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
 			localBlobTxs[account] = txs
 		}
 	}
+
+	if bidTxs != nil {
+		for acc, txs := range localPlainTxs {
+			for i := len(txs) - 1; i >= 0; i-- {
+				if bidTxs.Contains(txs[i].Hash) {
+					txs = txs[i+1:]
+					break
```
zjubfd, line 1091: we get a problem, shall we break here?

You know it is possible that account A can have two tx1 and tx2 in the `bidTxs`, if the logic here match the tx1, then it will still have the tx with same nonce with tx2 in the pool, so I think we may should not use txhash to detect conflict, but use the nonce of an account to detect the conflict.

miner/worker.go
irrun: fixed

miner/worker.go
```diff
@@ -1080,6 +1081,30 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
...
+				for i := len(txs) - 1; i >= 0; i-- {
+					if bidTxs.Contains(txs[i].Hash) {
+						txs = txs[i+1:]
+						break
+					}
+				}
+				if len(txs) == 0 {
+					delete(commonTxs, acc)
+				} else {
+					commonTxs[acc] = txs
```
zjubfd, line 1097: if no conflict at all, why do we bother to do assign again?

miner/bid_simulator.go
```diff
@@ -500,8 +504,13 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		blockNumber = bidRuntime.bid.BlockNumber
 		parentHash  = bidRuntime.bid.ParentHash
 		builder     = bidRuntime.bid.Builder
-		err         error
-		success     bool
+
+		bidTxs   = bidRuntime.bid.Txs
+		bidLen   = len(bidTxs)
```
zjubfd, line 509: `bidLen` ==> `bidTxLen`

miner/worker.go
reneecok: cause txs is local value, it's not point to map value

reneecok: fixed

miner/bid_simulator.go
reneecok: ok

zjubfd (approved):

miner/bid_simulator.go
```diff
@@ -593,6 +608,38 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+		delay := b.engine.Delay(b.chain, bidRuntime.env.header, &b.delayLeftOver)
+		if delay != nil && *delay > 0 {
+			log.Debug("BidSimulator: greedy merge tx stopTimer", "block", bidRuntime.env.header.Number,
+				"header time", time.Until(time.Unix(int64(bidRuntime.env.header.Time), 0)),
+				"commit delay", *delay, "DelayLeftOver", b.delayLeftOver)
+
+			stopTimer := time.NewTimer(*delay)
+
+			bidTxsSet := mapset.NewSet[common.Hash]()
+			for _, tx := range bidRuntime.bid.Txs {
```
zzzckck, line 622: better to eliminate this for loop.
We can pass `bidRuntime.bid.Txs` into `fillTransactions(..., bidRuntime.bid.Txs)` directly.
```
// in fillTransactions(...)
for _, bigTx range bidRuntime.bid.Txs {
   // check and remove from pendingPlainTxs & pendingBlobTxs
}
```

miner/worker.go
```diff
@@ -1080,6 +1081,29 @@ func (w *worker) fillTransactions(interruptCh chan int32, env *environment, stop
 			localBlobTxs[account] = txs
 		}
 	}
+
```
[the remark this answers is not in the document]
irrun: ok,fixed

miner/bid_simulator.go
irrun: hashset makes it simple

zzzckck (approved):

unclezoro (approved):

Issue #2331, opened 2024-03-23
Fast node returns wrong state siliently


wtdcode:
#### System information

Geth version: `geth_linux version 1.3.9-5e74ea65-20240220`
OS & Version: Linux

Command line:

```
             --syncmode full --datadir /work/data --cache 8000 --rpc.allow-unprotected-txs
             --history.transactions 0
             --ws --ws.addr 0.0.0.0 --ws.api net,web3,eth --ws.port 8556 --ws.origins *
             --http --http.addr 0.0.0.0
             --http.api net,web3,eth
             --http.port 8555
             --http.vhosts *
             --db.engine=pebble
             --verbosity 4
             --tries-verify-mode=none
```

#### Expected behaviour

Return `missing trie root` or some other error.

#### Actual behaviour

Return 0

#### Steps to reproduce the behaviour

```
curl -H 'Content-Type: application/json'  -X POST --data '{"jsonrpc":"2.0", "method": "eth_getStorageAt", "params": ["0x00020F14443e103A0B72F1f2DFDaac07DDAa6Df5", "0x0", "0x237f199"], "id": 1}' http://127.0.0.1:8555
{"jsonrpc":"2.0","id":1,"result":"0x0000000000000000000000000000000000000000000000000002704c6f721947"}
```

After `0x237f199` falls 128 blocks behind, executing the same command.

```
curl -H 'Content-Type: application/json'  -X POST --data '{"jsonrpc":"2.0", "method": "eth_getStorageAt", "params": ["0x00020F14443e103A0B72F1f2DFDaac07DDAa6Df5", "0x0", "0x237f199"], "id": 1}' http://127.0.0.1:8555
{"jsonrpc":"2.0","id":1,"result":"0x0000000000000000000000000000000000000000000000000000000000000000"}
```

The state silently changes. I would expect something like `missing trie node`

zzzckck:
noted, thank you for your feedback, we will fix it.

weiihann:
The reason why `0x00...00` is returned is that the snap layer specified has expired (i.e. after 128 blocks), so the program will read from an empty trie since it's a fast node. Attempting to retrieve a value from an empty trie will return `0x00...00` by default instead of an error.

So the state doesn't change, it's just that the specified snap layer is missing.

zzzckck:
[...]

PR #2369, opened 2024-04-03
core: stateDb has no trie and no snap return err

weiihann:
### Description
This PR adds a check to `Blockchain.StateAt()` such that if the node is a fast node (i.e. NoTrie = true) and the snap layer cannot be found, then it will return an error. 

### Rationale
Refer to #2331.

By default, a fast node will have an empty trie for any state. Retrieving a value from an empty trie will give `0x00...00`. It may be confusing for users as `0x00...00` could also be the actual value. Hence, it's more useful to throw an error when the state is not available. 

### Example
Command
```
curl -H 'Content-Type: application/json'  -X POST --data '{"jsonrpc":"2.0", "method": "eth_getStorageAt", "params": ["0x00020F14443e103A0B72F1f2DFDaac07DDAa6Df5", "0x0", "0x237f199"], "id": 1}' http://127.0.0.1:8545
```

Output
```
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"state is not available"}}
```

### Changes
- Add `StateDB.GetSnap()`

### DIFF
--- a/core/blockchain_reader.go
+++ b/core/blockchain_reader.go
@@ -396,7 +396,20 @@ func (bc *BlockChain) State() (*state.StateDB, error) {
 
 // StateAt returns a new mutable state based on a particular point in time.
 func (bc *BlockChain) StateAt(root common.Hash) (*state.StateDB, error) {
-	return state.New(root, bc.stateCache, bc.snaps)
+	stateDb, err := state.New(root, bc.stateCache, bc.snaps)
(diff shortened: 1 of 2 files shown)


setunapo (approved):

zzzckck (approved):

flywukong (approved):
LGTM

PR #2383, opened 2024-04-09
blobtx: fix some found issues;

GalaIO:
### Description

blobtx: fix some found issues;

### Rationale

The newly added freezer table(BlobTable) relies on `FreezerEnv`, so add some code to detect that scenario.

### Changes

Notable changes: 
* freezer: fix some env missing issues;
* logs: downgrade some info level logs;
* da: add some DA benchmark;
* ...

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -308,6 +308,9 @@ func getBlobExtraReserveFromEnv(env *ethdb.FreezerEnv) uint64 {
 }
 
 func (f *chainFreezer) freezeRangeWithBlobs(nfdb *nofreezedb, number, limit uint64) (hashes []common.Hash, err error) {
+	defer func() {
+		log.Debug("freezeRangeWithBlobs", "from", number, "to", limit, "err", err)
+	}()
 	lastHash := ReadCanonicalHash(nfdb, limit)
 	if lastHash == (common.Hash{}) {
 		return nil, fmt.Errorf("canonical hash missing, can't freeze block %d", limit)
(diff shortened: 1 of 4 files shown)


core/rawdb/chain_freezer.go
```diff
@@ -413,6 +429,21 @@ func (f *chainFreezer) SetupFreezerEnv(env *ethdb.FreezerEnv) error {
...
+func (f *chainFreezer) checkFreezerEnv() error {
+	_, exist := f.freezeEnv.Load().(*ethdb.FreezerEnv)
+	if exist {
+		return nil
+	}
+	blobFrozen, err := f.TableAncients(ChainFreezerBlobSidecarTable)
+	if err != nil {
+		return err
+	}
+	if blobFrozen > 0 {
```
emailtovamos, line 441: How does this condition means freezer env is missing? A blob can still freeze by default right?

core/rawdb/chain_freezer.go
```diff
@@ -289,9 +308,6 @@ func getBlobExtraReserveFromEnv(env *ethdb.FreezerEnv) uint64 {
 }
 
 func (f *chainFreezer) freezeRangeWithBlobs(nfdb *nofreezedb, number, limit uint64) (hashes []common.Hash, err error) {
-	defer func() {
-		log.Info("freezeRangeWithBlobs", "from", number, "to", limit, "err", err)
```
buddh0, line 293: we can change it to debug level, so we can we found issue when do test

buddh0 (approved):

core/rawdb/chain_freezer.go
GalaIO: yes, this condition means that env must be guaranteed to exist when the blob table already has data.

emailtovamos (approved):

zzzckck (approved):

PR #2395, opened 2024-04-15
fix: trieJournal format compatible old db format

jingjunLi:
### Description

If trieJournal is stored in the database, it retains the old journal format; if trieJournal writes to the journal file using the new journal format;

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -92,7 +92,7 @@ type layer interface {
 	// journal commits an entire diff hierarchy to disk into a single journal entry.
 	// This is meant to be used during shutdown to persist the layer without
 	// flattening everything down (bad for reorgs).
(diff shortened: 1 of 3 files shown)


RenRick (dismissed):
LGTM

Fynnss (dismissed):
LGTM

NathanBSC:
can we avoid to change the interface of function loadDiffLayer, loadDiskLayer and journal?
we can already get JournalFilePath by `db.config.JournalFilePath`

NathanBSC:
why a field `length` is need?
rlp decode will not use it, I also can't see other usage

NathanBSC (approved):

buddh0 (approved):

zzzckck (approved):

PR #2404, opened 2024-04-18
cmd, p2p: filter peers by regex on name

weiihann:
### Description
This PR adds a new feature to filter peers by their names based on a set of customized regex patterns.

### Rationale
Some outdated versions of nodes running on BSC may keep connecting to a healthy set of peers and only consume information, but not return any useful data. Such nodes should be filtered out to save network bandwidth.

### Example
In the config file:
```
...
[Node.P2P]
PeerFilterPatterns = ["(?!)geth"]
...
```

An example of a false regexp pattern error message:
```
t=2024-04-18T06:02:10+0000 lvl=error msg="Failed to compile peer filter patterns" err="error parsing regexp: missing closing ): `abc(def`"

```

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -145,6 +145,7 @@ var (
 		// utils.MinerNewPayloadTimeout,
 		utils.NATFlag,
 		utils.NoDiscoverFlag,
+		utils.PeerFilterPatternsFlag,
 		utils.DiscoveryV4Flag,
 		utils.DiscoveryV5Flag,
 		utils.InstanceFlag,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -882,6 +882,11 @@ var (
 		Usage:    "Disables the peer discovery mechanism (manual peer addition)",
 		Category: flags.NetworkingCategory,
 	}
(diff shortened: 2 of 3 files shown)


zzzckck (approved):

wellgamer789:
> "may keep connecting to a healthy set of peers and only consume information, but not return any useful data."

I may be wrong, but i think what you are fighting are "fake nodes" and they can easily bypass this "check".

zzzckck:
> > "may keep connecting to a healthy set of peers and only consume information, but not return any useful data."
> 
> I may be wrong, but i think what you are fighting are "fake nodes" and they can easily bypass this "check".

yes, it is just optional feature, can not solve the case you mentioned. But could be useful in some cases.

kaber2:
Obsolete peers should be filtered based on the fork ID. This seems really useless.

zzzckck:
to filter out specified nodes, suppose client name is "alice"
```
[Node.P2P]
PeerFilterPatterns = ["^alice.*"]
```
This will reject connection from client with name "alice*"

setunapo:
https://www.jyshare.com/front-end/854/?optionGlobl=global

PR #2406, opened 2024-04-18
feat: adaptive for loading journal file or journal kv during loadJournal

jingjunLi:
### Description

During loading, If it is not necessary to judge based on --journalfile, handle it based on the existence of journalkv or journalfile;
* If trieJournal exists in the database, load it from the database first;
* If the journal file exists, load it from the journal file;
* If neither exists, treat it as a kv;

During persistence:
Handle it based on --journalfile, without considering whether it was journalkv or journalfile before;

Verification cases:

* Start and shut down using the kv method; Verify (type: journal kv);
* Start by specifying --journalfile to use journal file, and verify if it can start (type: journal kv), then shut down to save (type: journal file); Verify block tracking;
* Start using the kv method again (type: journal file); Verify block tracking; then shut down to save (type: journal kv); Verify block tracking again;
* Start by specifying --journalfile;

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/triedb/pathdb/database.go
+++ b/triedb/pathdb/database.go
@@ -65,6 +65,14 @@ const (
 	DefaultBatchRedundancyRate = 1.1
 )
 
+type JournalType int
+
+const (
+	JournalNone JournalType = iota
+	JournalKVType
+	JournalFileType
+)
+
 // layer is the interface implemented by all state layers which includes some
 // public methods and some additional methods for internal usage.
 type layer interface {
@@ -92,7 +100,7 @@ type layer interface {
 	// journal commits an entire diff hierarchy to disk into a single journal entry.
 	// This is meant to be used during shutdown to persist the layer without
 	// flattening everything down (bad for reorgs).
-	journal(w io.Writer, journalFile bool) error
+	journal(w io.Writer, journalType JournalType) error
 }
 
 // Config contains the settings for database.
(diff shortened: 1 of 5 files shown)


triedb/pathdb/database.go
```diff
@@ -65,6 +65,14 @@ const (
 	DefaultBatchRedundancyRate = 1.1
 )
 
+type JournalType int
+
+const (
+	JournalNone JournalType = iota
```
Fynnss, line 71: Seems an unused type?

triedb/pathdb/journal.go
```diff
@@ -193,7 +195,7 @@ func newJournalReader(file string, db ethdb.Database) (JournalReader, error) {
 // loadJournal tries to parse the layer journal from the disk.
 func (db *Database) loadJournal(diskRoot common.Hash) (layer, error) {
 	start := time.Now()
-	reader, err := newJournalReader(db.config.JournalFilePath, db.diskdb)
+	reader, err := newJournalReader(db.config.JournalFilePath, db.diskdb, db.DetermineJournalTypeByStorage())
```
Fynnss, line 198: If the journalFile flag is specified, but the file doesn't exist, won't an error be reported?

I think it should return an error, rather than use kv type by default as the PR do.

Fynnss (approved):
LGTM

Issue #2414, opened 2024-04-22
use errors.New to replace fmt.Errorf with no parameters will much better.


ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better.

PR #2415, opened 2024-04-22
chore: use errors.New to replace fmt.Errorf with no parameters (#2414)

ChengenH:
use errors.New to replace fmt.Errorf with no parameters.


zzzckck (approved):

zzzckck:
cmd/era is mainly for Ethereum consensus layer, check out: https://ethresear.ch/t/era-archival-files-for-block-and-consensus-data/13526
BSC actually does not use era right now, so will not accept this PR, thx for your contribution.

PR #2419, opened 2024-04-23
metrics: add doublesign counter

emailtovamos:
### Description

double sign counter metric

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -48,8 +48,9 @@ import (
 )
 
 const (
-	inMemorySnapshots  = 256  // Number of recent snapshots to keep in memory
-	inMemorySignatures = 4096 // Number of recent block signatures to keep in memory
+	inMemorySnapshots  = 256   // Number of recent snapshots to keep in memory
+	inMemorySignatures = 4096  // Number of recent block signatures to keep in memory
+	inMemoryHeaders    = 86400 // Number of recent headers to keep in memory for double sign detection
 
 	checkpointInterval = 1024        // Number of blocks after which to save the snapshot to the database
 	defaultEpochLength = uint64(100) // Default number of blocks of checkpoint to update validatorSet from contract
@@ -267,7 +268,7 @@ func New(
 	if err != nil {
 		panic(err)
 	}
-	recentHeaders, err := lru.NewARC(inMemorySignatures)
+	recentHeaders, err := lru.NewARC(inMemoryHeaders)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


consensus/parlia/parlia.go
```diff
@@ -2011,3 +2031,17 @@ func applyMessage(
 	}
 	return msg.Gas() - returnGas, err
 }
+
+// proposalKey build a key which is a combination of the block number and the proposer address.
+func proposalKey(header types.Header) string {
+
+	slotKey := uintToString(header.Number.Uint64())
+	proposerIndexKey := uintToString(header.Coinbase.Big().Uint64())
```
zzzckck, line 2039: `header.Coinbase.String()`?

consensus/parlia/parlia.go
```diff
@@ -2011,3 +2031,17 @@ func applyMessage(
...
+
+	return slotKey + ":" + proposerIndexKey
+}
+
+// Turns a uint64 value to a string representation.
+func uintToString(val uint64) string {
```
zzzckck, line 2045: seems no need

consensus/parlia/parlia.go
```diff
@@ -2011,3 +2030,12 @@ func applyMessage(
 	}
 	return msg.Gas() - returnGas, err
 }
+
+// proposalKey build a key which is a combination of the block number and the proposer address.
+func proposalKey(header types.Header) string {
```
zzzckck, line 2035: double sign is not for sign at same block height, but sign on same block, e.g. with same ParentHash.
it should be: 
```
return header.ParentHash.String()+header.Coinbase.String()
```

consensus/parlia/parlia.go
```diff
@@ -263,6 +267,10 @@ func New(
 	if err != nil {
 		panic(err)
 	}
+	recentHeaders, err := lru.NewARC(inMemorySignatures)
```
zzzckck, line 270: inMemorySignatures = 256, is too small, double sign would trace back to recent 86400 blocks.
better to set the cache size to 86400

zzzckck (approved):

galaio (approved):

consensus/parlia/parlia.go
```diff
@@ -813,6 +823,16 @@ func (p *Parlia) verifySeal(chain consensus.ChainHeaderReader, header *types.Hea
 		return errUnauthorizedValidator(signer.String())
 	}
 
+	// check for double sign & add to cache
+	key := proposalKey(*header)
+
+	value, ok := p.recentHeaders.Get(key)
+	if ok {
+		doubleSignCounter.Inc(1)
```
zzzckck, line 831: pls add a warning message here

PR #2451, opened 2024-05-13
core/forkchoice: improve stability when inturn block not generated

NathanBSC:
### Description

core/forkchoice: improve stablity when inturn block not generated

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/forkchoice.go
+++ b/core/forkchoice.go
@@ -114,7 +114,9 @@ func (f *ForkChoice) ReorgNeeded(current *types.Header, extern *types.Header) (b
 		if f.preserve != nil {
 			currentPreserve, externPreserve = f.preserve(current), f.preserve(extern)
 		}
-		reorg = !currentPreserve && (externPreserve || f.rand.Float64() < 0.5)
+		reorg = !currentPreserve && (externPreserve ||
+			extern.Time < current.Time ||
+			extern.Time == current.Time && f.rand.Float64() < 0.5)
 	}
 	return reorg, nil
 }


zzzckck (approved):

unclezoro (approved):

zzzckck (changes requested):
In case of unknown bug, postpone this PR after Tycho mainnet HF.

core/forkchoice.go
```diff
@@ -114,7 +114,9 @@ func (f *ForkChoice) ReorgNeeded(current *types.Header, extern *types.Header) (b
 		if f.preserve != nil {
 			currentPreserve, externPreserve = f.preserve(current), f.preserve(extern)
 		}
-		reorg = !currentPreserve && (externPreserve || f.rand.Float64() < 0.5)
+		reorg = !currentPreserve && (externPreserve ||
+			extern.Time < current.Time ||
+			extern.Time == current.Time && f.rand.Float64() < 0.5)
```
annielz, line 119: Come from postmortem sharing, just feel it might be better to write as 
reorg = !currentPreserve && (externPreserve ||
			extern.Time < current.Time ||
			(extern.Time == current.Time && f.rand.Float64() < 0.5)) , adding a pair of parentheses for the last && to make the logic clearer. Not sure about the context, so please correct me if I misunderstand.

buddh0: > Come from postmortem sharing, just feel it might be better to write as reorg = !currentPreserve && (externPreserve || extern.Time < current.Time || (extern.Time == current.Time && f.rand.Float64() < 0.5)) , adding a pair of parentheses for the last && to make the logic clearer. Not sure about the context, so please correct me if I misunderstand.

thx for review!

the operator `&&` has a higher prority than `||`.
If I write 117~119 into one line, I will follow your code to add a pair of parentheses.
here, I spilt 3 conditions into 3 lines, so It's also clear to me.

NathanBSC: > Come from postmortem sharing, just feel it might be better to write as reorg = !currentPreserve && (externPreserve || extern.Time < current.Time || (extern.Time == current.Time && f.rand.Float64() < 0.5)) , adding a pair of parentheses for the last && to make the logic clearer. Not sure about the context, so please correct me if I misunderstand.

thx for review!

the operator && has a higher prority than ||.
If I write 117~119 into one line, I will follow your code to add a pair of parentheses.
here, I spilt 3 conditions into 3 lines, so It's also clear to me.

Kutasy (approved):

zzzckck (approved):

PR #2455, opened 2024-05-14
cmd: fix memory leak when big dataset

Fynnss:
### Description

This pr fix memory leak when inspect trie with big dataset(more than 300M account and 100M contract)

### Rationale

Bug fix for memory leak

### Example

```
./geth db inspect-trie latest 5000 10
```
This command will inspect the latest state tree with a concurrency of 5000 trie, and finally print out the account tree and top 10 contract tree statistics.

### Changes

Notable changes: 
* add topN for print top n storage trie
* refactor for calc topN
* fix memory leak


zzzckck (approved):

RenRick (approved):
LGTM

PR #2461, opened 2024-05-15
eth/handler: check lists in body before broadcast blocks

NathanBSC:
### Description

eth/handler: check lists in body before broadcast blocks

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -320,26 +320,22 @@ func newHandler(config *handlerConfig) (*handler, error) {
 	}
 
 	broadcastBlockWithCheck := func(block *types.Block, propagate bool) {
-		// All the block fetcher activities should be disabled
-		// after the transition. Print the warning log.
-		if h.merger.PoSFinalized() {
-			log.Warn("Unexpected validation activity", "hash", block.Hash(), "number", block.Number())
-			return
-		}
-		// Reject all the PoS style headers in the first place. No matter
-		// the chain has finished the transition or not, the PoS headers
-		// should only come from the trusted consensus layer instead of
-		// p2p network.
-		if beacon, ok := h.chain.Engine().(*beacon.Beacon); ok {
-			if beacon.IsPoSHeader(block.Header()) {
-				log.Warn("unexpected post-merge header")
-				return
-			}
-		}
 		if propagate {
-			if err := core.IsDataAvailable(h.chain, block); err != nil {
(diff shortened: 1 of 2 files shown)


eth/handler.go
```diff
@@ -320,26 +320,22 @@ func newHandler(config *handlerConfig) (*handler, error) {
...
+				checkErrs <- core.ValidateListsInBody(block)
+			}()
+			go func() {
+				checkErrs <- core.IsDataAvailable(h.chain, block)
+			}()
+
+			for i := 0; i < len(checkErrs); i++ {
+				err := <-checkErrs
+				if err != nil {
+					log.Error("Propagating block with invalid sidecars", "number", block.Number(), "hash", block.Hash(), "err", err)
```
zzzckck, line 336: "Propagating block with invalid sidecars", ==> "Propagating invalid block",?

NathanBSC: accepted

zzzckck (approved):

eth/handler.go
```diff
@@ -320,26 +320,22 @@ func newHandler(config *handlerConfig) (*handler, error) {
 	}
 
 	broadcastBlockWithCheck := func(block *types.Block, propagate bool) {
-		// All the block fetcher activities should be disabled
```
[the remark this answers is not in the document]
buddh0: > may need to retain these codes? because it is used in propagate & not propagate block scenarios?

no
becaused these code have already exsited in the lamda function `validator`, it will be called before broadcastBlockWithCheck

galaio (approved):

eth/handler.go
```diff
@@ -320,26 +320,22 @@ func newHandler(config *handlerConfig) (*handler, error) {
...
-				return
-			}
-		}
 		if propagate {
-			if err := core.IsDataAvailable(h.chain, block); err != nil {
-				log.Error("Propagating block with invalid sidecars", "number", block.Number(), "hash", block.Hash(), "err", err)
-				return
+			checkErrs := make(chan error, 2)
+
+			go func() {
```
galaio, line 326: It seems the received error from the channel is random. pls ref: https://go.dev/play/p/I70UtyRwREy

galaio: maybe use `2` instead of `len(checkErrs)`.

galaio (approved):

zzzckck (approved):

PR #2467, opened 2024-05-16
internal/ethapi: add optional parameter for blobSidecars

zlacfzy:
### Description

Make the param showBlob optional, user don't need change the way to query blobs, even if they don't add the bool param it work as before.

### Rationale

### Example
The two way both works now.
```
{"jsonrpc":"2.0","method":"eth_getBlobSidecars","params":["latest", false],"id":1}]'
{"jsonrpc":"2.0","method":"eth_getBlobSidecars","params":["latest"],"id":1}]'
```

### Changes

Notable changes: 
* ...

### DIFF
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -612,7 +612,7 @@ web3._extend({
 		new web3._extend.Method({
 			name: 'getBlockReceipts',
 			call: 'eth_getBlockReceipts',
-			params: 2,
+			params: 1,
 		}),
 		new web3._extend.Method({
 			name: 'getBlobSidecars',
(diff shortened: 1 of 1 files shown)


internal/web3ext/web3ext.go
```diff
@@ -617,12 +617,12 @@ web3._extend({
 		new web3._extend.Method({
 			name: 'getBlobSidecars',
 			call: 'eth_getBlobSidecars',
-			params: 2,
+			params: 1,
```
NathanBSC, line 620: keep 2?

buddh0 (approved):

zzzckck (approved):

Zionsammy (commented):
contracts/helpers/VerifierCaller.sol

PR #2473, opened 2024-05-17
fix: limit the gas price of the mev bid

irrun:
### Description

fix: imit the gas price of the mev bid

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -78,6 +78,7 @@ type simBidReq struct {
 type bidSimulator struct {
 	config        *MevConfig
 	delayLeftOver time.Duration
+	minGasPrice   *big.Int
 	chain         *core.BlockChain
 	chainConfig   *params.ChainConfig
 	engine        consensus.Engine
@@ -114,6 +115,7 @@ type bidSimulator struct {
 func newBidSimulator(
 	config *MevConfig,
 	delayLeftOver time.Duration,
+	minGasPrice *big.Int,
 	chain *core.BlockChain,
 	chainConfig *params.ChainConfig,
 	engine consensus.Engine,
@@ -122,6 +124,7 @@ func newBidSimulator(
 	b := &bidSimulator{
 		config:        config,
 		delayLeftOver: delayLeftOver,
+		minGasPrice:   minGasPrice,
 		chain:         chain,
 		chainConfig:   chainConfig,
 		engine:        engine,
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -17,8 +17,8 @@ require (
 	github.com/btcsuite/btcd/btcec/v2 v2.3.2
 	github.com/cespare/cp v1.1.1
 	github.com/cloudflare/cloudflare-go v0.79.0
-	github.com/cometbft/cometbft v0.37.0
 	github.com/cockroachdb/pebble v1.1.0
+	github.com/cometbft/cometbft v0.37.0
```
unclezoro, line 21: reduce unnecessary changes

miner/bid_simulator.go
```diff
@@ -78,6 +78,7 @@ type simBidReq struct {
 type bidSimulator struct {
 	config        *MevConfig
 	delayLeftOver time.Duration
+	minGasPrice   *big.Int
```
unclezoro, line 81: no need for this config. we reuse the config in the miner module.

miner/bid_simulator.go
```diff
@@ -617,15 +621,35 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
-		return
+	// check if bid gas price is lower than min gas price
+	{
+		baseFee := bidRuntime.env.header.BaseFee
+		bidGasUsed := uint64(0)
+		bidGasFee := new(big.Int)
+
+		for i, receipt := range bidRuntime.env.receipts {
+			bidGasUsed += receipt.GasUsed
+			effectiveGasPrice := new(big.Int).Add(baseFee, bidRuntime.env.txs[i].EffectiveGasTipValue(baseFee))
```
unclezoro, line 641: what is the usage of `effectiveGasPrice`?

miner/bid_simulator.go
irrun: it is

miner/bid_simulator.go
irrun: for calculating txFee. If using the balance of system address, there maybe some transfer tx as tips to system.

unclezoro: yeah, that is what we want I think, sending BNB to system address is also valid.

irrun: fixed

unclezoro (approved):

zzzckck (approved):

PR #2479, opened 2024-05-20
dev: ensure consistency in BPS bundle result

Jolly23:
### Description

This commit introduces a NonRevertible hash list within the RPC to address potential inconsistencies between the builder and validator during bundle operations. The builder sends a list of transaction hashes, which must be successfully executed, to the validator. If the validator cannot accurately execute the bundle, it refrains from packing that block. This mechanism ensures that both components maintain output consistency under the same input conditions, thereby preventing financial losses for searchers due to incorrect block inclusion.

### Rationale

This happened on BloxRoute builder before. The simulation runs fine on builder but one transaction in bundle somehow couldn't pass on validator, which costs searcher $$$$$

### Example

The builder sends the list of hashes from bundle txs that is no-revertible, to ensure the validator runs the MEV bundles right.

### Changes

Notable changes: 
* builder `SendBid` with `NonRevertible`

### DIFF
--- a/core/types/bid.go
+++ b/core/types/bid.go
@@ -42,11 +42,11 @@ func (b *BidArgs) ToBid(builder common.Address, signer Signer) (*Bid, error) {
 		return nil, err
 	}
 
-	if len(b.RawBid.NonRevertible) > len(txs) {
+	if len(b.RawBid.UnRevertible) > len(txs) {
 		return nil, fmt.Errorf("expect NonRevertible no more than %d", len(txs))
 	}
-	nonRevertibleHashes := mapset.NewThreadUnsafeSetWithSize[common.Hash](len(b.RawBid.NonRevertible))
-	nonRevertibleHashes.Append(b.RawBid.NonRevertible...)
+	unRevertibleHashes := mapset.NewThreadUnsafeSetWithSize[common.Hash](len(b.RawBid.UnRevertible))
+	unRevertibleHashes.Append(b.RawBid.UnRevertible...)
 
 	if len(b.PayBidTx) != 0 {
 		var payBidTx = new(Transaction)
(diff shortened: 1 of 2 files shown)


core/types/bid.go
```diff
@@ -70,12 +79,13 @@ func (b *BidArgs) ToBid(builder common.Address, signer Signer) (*Bid, error) {
...
-	BlockNumber uint64          `json:"blockNumber"`
-	ParentHash  common.Hash     `json:"parentHash"`
-	Txs         []hexutil.Bytes `json:"txs"`
-	GasUsed     uint64          `json:"gasUsed"`
-	GasFee      *big.Int        `json:"gasFee"`
-	BuilderFee  *big.Int        `json:"builderFee"`
+	BlockNumber   uint64          `json:"blockNumber"`
+	ParentHash    common.Hash     `json:"parentHash"`
+	Txs           []hexutil.Bytes `json:"txs"`
+	NonRevertible []common.Hash   `json:"nonRevertible"`
```
irrun, line 85: prefer Revertible, for less request size, also consistent with sendBundle

miner/bid_simulator.go
```diff
@@ -727,7 +727,7 @@ func (r *BidRuntime) packReward(validatorCommission uint64) {
 	r.packedValidatorReward.Sub(r.packedValidatorReward, r.bid.BuilderFee)
 }
 
-func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *params.ChainConfig, tx *types.Transaction) error {
+func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *params.ChainConfig, tx *types.Transaction, noRevertible bool) error {
```
irrun, line 730: noRevertible => unRevertible

miner/bid_simulator.go
```diff
@@ -758,6 +758,10 @@ func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *para
 		env.state.RevertToSnapshot(snap)
 		env.gasPool.SetGas(gp)
 		return err
+	} else if noRevertible && receipt.Status == types.ReceiptStatusFailed {
+		env.state.RevertToSnapshot(snap)
+		env.gasPool.SetGas(gp)
```
irrun, line 763: I think no need to care env now, also L758&L759 can be deleted

core/types/bid.go
unclezoro: Actually,  NonRevertible is better, because tx from searcher is less I believe, if you check the MEV explorer.

miner/bid_simulator.go
unclezoro: I think the developer just copy line 758-759, if no need we should delete both

unclezoro:
Please rebase the code to develop branch.

core/types/bid.go
reneecok: ok, but NonRevertible=>UnRevertible

unclezoro (approved):

irrun (approved):

PR #2482, opened 2024-05-20
BEP-341: Validators can produce consecutive blocks

NathanBSC:
### Description

[BEP-341: Validators can produce consecutive blocks](https://github.com/bnb-chain/BEPs/pull/341)
[related change in system contracts](https://github.com/bnb-chain/bsc-genesis-contract/pull/562)

### Rationale

### Example

### Changes

Notable changes: 
* ...


buddh0:
need change the name of hardfork later

consensus/parlia/parlia.go
```diff
@@ -381,11 +388,15 @@ func getValidatorBytesFromHeader(header *types.Header, chainConfig *params.Chain
 		return nil
 	}
 	num := int(header.Extra[extraVanity])
-	if num == 0 || len(header.Extra) <= extraVanity+extraSeal+num*validatorBytesLength {
-		return nil
-	}
 	start := extraVanity + validatorNumberSize
 	end := start + num*validatorBytesLength
+	extraMinLen := end + extraSeal
```
0xbundler, line 393: The extra dynamic part is so complex now, could we use rlp to encode the custom data into extra?

```golang
|---Extra Vanity---|---RLP(CustomData)---|---Extra Seal---|
|---Fixed 32 bytes---|---dynamic length---|---Fixed 65 bytes---|
```

consensus/parlia/parlia.go
```diff
@@ -884,6 +898,35 @@ func (p *Parlia) prepareValidators(header *types.Header) error {
...
+		if err != nil {
+			return err
+		}
+		if turnTerm == nil {
+			return errors.New("unexpected error when getTurnTerm")
+		}
+		header.Extra = append(header.Extra, byte(int(turnTerm.Int64())))
+		log.Debug("prepareTurnTerm", "turnTerm", turnTerm.Int64())
+	} else {
+		log.Debug("prepareTurnTerm", "turnTerm", "defaultTurnTerm")
```
0xbundler, line 923: This line cannot be reached, because the previous `IsBohr` check has returned.

consensus/parlia/parlia.go
```diff
@@ -1065,6 +1111,42 @@ func (p *Parlia) verifyValidators(header *types.Header) error {
...
+			turnTermFromContract, err := p.getTurnTerm(parent)
+			if err != nil {
+				return err
+			}
+			if turnTermFromContract != nil && uint8(turnTermFromContract.Int64()) == *turnTermFromHeader {
+				log.Debug("verifyTurnTerm", "turnTerm", turnTermFromContract.Int64())
+				return nil
+			}
+		} else {
+			if uint8(defaultTurnTerm) == *turnTermFromHeader {
```
0xbundler, line 1140: Same as before, this line cannot be reached, because the previous IsBohr check has returned.

consensus/parlia/snapshot.go
```diff
@@ -410,6 +461,24 @@ func parseValidators(header *types.Header, chainConfig *params.ChainConfig, parl
...
+
+	if len(header.Extra) <= extraVanity+extraSeal {
+		return nil, errors.New("invalid turnTerm")
+	}
+	num := int(header.Extra[extraVanity])
+	pos := extraVanity + validatorNumberSize + num*validatorBytesLength
+	if len(header.Extra) <= pos {
+		return nil, errors.New("invalid turnTerm")
+	}
+	turnterm := header.Extra[pos]
```
0xbundler, line 478: How to distinguish the empty `turnterm`? Here is the extra structure comment.

```golang
// |---Extra Vanity---|---Validators Number and Validators Bytes (or Empty)---|---Turn Term (or Empty)---|---Vote Attestation (or Empty)---|---Extra Seal---|
```

consensus/parlia/parlia.go
buddh0: > This line cannot be reached, because the previous `IsBohr` check has returned.

differernt: IsBohr(header... vs IsBohr(parent...  
It will be reached, if the epoch block is the first Bohr Block

consensus/parlia/snapshot.go
buddh0: > How to distinguish the empty `turnterm`? Here is the extra structure comment.
> 
> ```go
> // |---Extra Vanity---|---Validators Number and Validators Bytes (or Empty)---|---Turn Term (or Empty)---|---Vote Attestation (or Empty)---|---Extra Seal---|
> ```

If IsBohr(header.Number, header.Time), the `turnterm` won't be empty, otherwise will be empty

consensus/parlia/parlia.go
NathanBSC: > The extra dynamic part is so complex now, could we use rlp to encode the custom data into extra?
> 
> ```go
> |---Extra Vanity---|---RLP(CustomData)---|---Extra Seal---|
> |---Fixed 32 bytes---|---dynamic length---|---Fixed 65 bytes---|
> ```

good idea!
I think we can do this optimization in another seperate PR

consensus/parlia/parlia.go
NathanBSC: > Same as before, this line cannot be reached, because the previous IsBohr check has returned.
☝️

NathanBSC: > > Same as before, this line cannot be reached, because the previous IsBohr check has returned.
> > refer above reply from @buddh0  ☝️

consensus/parlia/snapshot.go
```diff
@@ -210,17 +218,45 @@ func (s *Snapshot) updateAttestation(header *types.Header, chainConfig *params.C
 	}
 }
 
-func (s *Snapshot) SignRecently(validator common.Address) bool {
+func (s *Snapshot) versionHistoryCheckLen() uint64 {
+	return uint64(len(s.Validators)) * s.TurnTerm
+}
+
+func (s *Snapshot) minerHistoryCheckLen() uint64 {
+	return (uint64(len(s.Validators)) / 2) * s.TurnTerm
```
galaio, line 226: Why not `uint64(len(s.Validators)/2 + 1` here?

consensus/parlia/snapshot.go
0xbundler: Ok, I get it, maybe the comment below will be more accurate.
```golang
// |---Extra Vanity---|---Validators Number & Validators Bytes & Turn Term (or Empty)---|---Vote Attestation (or Empty)---|---Extra Seal---|
```
It cannot be empty in epoch change header after `Bohr`, like validator set, otherwise it will be empty.

consensus/parlia/snapshot.go
```diff
@@ -247,10 +283,10 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 	for _, header := range headers {
 		number := header.Number.Uint64()
 		// Delete the oldest validator from the recent list to allow it signing again
-		if limit := uint64(len(snap.Validators)/2 + 1); number >= limit {
+		if limit := snap.minerHistoryCheckLen() + 1; number >= limit { // `+1` for genesis block which has no miner
```
galaio, line 286: maybe using `uint64(len(snap.Validators)/2 + 1)*TurnTerm` is correct?

consensus/parlia/snapshot.go
```diff
@@ -410,6 +461,24 @@ func parseValidators(header *types.Header, chainConfig *params.ChainConfig, parl
...
+		return nil, nil
+	}
+
+	if len(header.Extra) <= extraVanity+extraSeal {
+		return nil, errors.New("invalid turnTerm")
+	}
+	num := int(header.Extra[extraVanity])
+	pos := extraVanity + validatorNumberSize + num*validatorBytesLength
+	if len(header.Extra) <= pos {
+		return nil, errors.New("invalid turnTerm")
```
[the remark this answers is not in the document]
NathanBSC: > both errors are same, I think they should be different otherwise we can not say where it was returned

accepted

WTDuck2255:
Sorry i don't understand how to do it? Thanks

ในวันที่ พ. 5 มิ.ย. 2024 18:31 น. Matus Kysel ***@***.***>
เขียนว่า:

consensus/parlia/parlia.go
```diff
@@ -381,11 +392,15 @@ func getValidatorBytesFromHeader(header *types.Header, chainConfig *params.Chain
...
-	if num == 0 || len(header.Extra) <= extraVanity+extraSeal+num*validatorBytesLength {
-		return nil
-	}
 	start := extraVanity + validatorNumberSize
 	end := start + num*validatorBytesLength
+	extraMinLen := end + extraSeal
+	if chainConfig.IsBohr(header.Number, header.Time) {
+		extraMinLen += turnTermSize
+	}
+	if num == 0 || len(header.Extra) < extraMinLen {
```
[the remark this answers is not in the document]
buddh0: > this condition was before `<=`

not add `validatorNumberSize` before, and validatorNumberSize equal to 1

so it‘s equivalent between old and new version

consensus/parlia/parlia.go
```diff
@@ -1065,6 +1115,42 @@ func (p *Parlia) verifyValidators(header *types.Header) error {
 	return nil
 }
 
+func (p *Parlia) verifyTurnTerm(chain consensus.ChainHeaderReader, header *types.Header) error {
```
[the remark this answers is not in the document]
buddh0: > this is just a suggestion but `verifyTurnTerm` is almost same as `prepareTrunTerm` maybe we can have 3rd function that will have common code in one place e.g.
> 
> ```go
> func (p *Parlia) processTurnTerm(chain consensus.ChainHeaderReader, header *types.Header, isPrepare bool) (uint8, error) {
> 	if header.Number.Uint64()%p.config.Epoch != 0 ||
> 		!p.chainConfig.IsBohr(header.Number, header.Time) {
> 		return 0, nil
> 	}
> 
> 	parent := chain.GetHeaderByHash(header.ParentHash)
> 	if parent == nil {
> 		return 0, errors.New("parent not found")
> 	}
> 
> 	var turnTerm uint8
> 	var err error
> 
> 	if p.chainConfig.IsBohr(parent.Number, parent.Time) {
> 		turnTermFromContract, err := p.getTurnTerm(parent)
> 		if err != nil {
> 			return 0, err
> 		}
> 		if turnTermFromContract == nil {
> 			return 0, errors.New("unexpected error when getTurnTerm")
> 		}
> 		turnTerm = uint8(turnTermFromContract.Int64())
> 		log.Debug("processTurnTerm", "turnTerm", turnTermFromContract.Int64())
> 	} else {
> 		turnTerm = uint8(defaultTurnTerm)
> 		log.Debug("processTurnTerm", "turnTerm", "defaultTurnTerm")
> 	}
> 
> 	if isPrepare {
> 		header.Extra = append(header.Extra, byte(turnTerm))
> 	}
> 
> 	return turnTerm, nil
> }
> 
> func (p *Parlia) prepareTurnTerm(chain consensus.ChainHeaderReader, header *types.Header) error {
> 	_, err := p.processTurnTerm(chain, header, true)
> 	return err
> }
> 
> func (p *Parlia) verifyTurnTerm(chain consensus.ChainHeaderReader, header *types.Header) error {
> 	if header.Number.Uint64()%p.config.Epoch != 0 ||
> 		!p.chainConfig.IsBohr(header.Number, header.Time) {
> 		return nil
> 	}
> 
> 	turnTermFromHeader, err := parseTurnTerm(header, p.chainConfig, p.config)
> 	if err != nil {
> 		return err
> 	}
> 	if turnTermFromHeader != nil {
> 		turnTerm, err := p.processTurnTerm(chain, header, false)
> 		if err != nil {
> 			return err
> 		}
> 
> 		if turnTerm == *turnTermFromHeader {
> 			return nil
> 		}
> 	}
> 
> 	return errMismatchingEpochTurnTerm
> }
> ```

accepted and updated 👍
a bit different from your above code, pls help review again

NathanBSC:
rebase develop 06.12

consensus/parlia/bohrFork.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/ethereum/go-ethereum/core/systemcontracts"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/internal/ethapi"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rpc"
+)
+
+func (p *Parlia) getTurnTerm(chain consensus.ChainHeaderReader, header *types.Header) (*uint8, error) {
+	if header.Number.Uint64()%p.config.Epoch != 0 ||
```
unclezoro, line 22: I think such logic `header.Number.Uint64()%p.config.Epoch != 0` should leave to the caller of `getTurnTerm`.  `getTurnTerm` function just finish its job.

consensus/parlia/bohrFork.go
```diff
@@ -0,0 +1,96 @@
...
+		!p.chainConfig.IsBohr(header.Number, header.Time) {
+		return nil, nil
```
unclezoro, line 24: just return 1 instead of nil? so that we can still use `uint8` instead of `*uint8` to make it simple

consensus/parlia/snapshot.go
```diff
@@ -43,6 +44,7 @@ type Snapshot struct {
 
 	Number           uint64                            `json:"number"`                // Block number where the snapshot was created
 	Hash             common.Hash                       `json:"hash"`                  // Block hash where the snapshot was created
+	TurnTerm         uint64                            `json:"turn_term"`             // Term of `turn`, meaning the consecutive number of blocks a validator receives priority for block production
```
unclezoro, line 47: can it be uint8 directly?

consensus/parlia/parlia.go
```diff
@@ -54,10 +54,12 @@ const (
 
 	checkpointInterval = 1024        // Number of blocks after which to save the snapshot to the database
 	defaultEpochLength = uint64(100) // Default number of blocks of checkpoint to update validatorSet from contract
+	defaultTurnTerm    = uint64(1)   // Default consecutive number of blocks a validator receives priority for block production
```
unclezoro, line 57: can be uint8 directly

consensus/parlia/parlia.go
```diff
@@ -1933,7 +1988,7 @@ func (p *Parlia) backOffTime(snap *Snapshot, header *types.Header, val common.Ad
 			// Exclude the recently signed validators
 			temp := make([]common.Address, 0, len(validators))
 			for _, addr := range validators {
-				if _, ok := recentsMap[addr]; ok {
+				if snap.signRecentlyByCounts(addr, counts) {
```
unclezoro, line 1991: can you change the `TestImpactOfValidatorOutOfService` to simulate the impact of a validator is down and how long it will take after rejoin.

consensus/parlia/snapshot.go
unclezoro, line 47: I will suggest run a fullnode to use this version of code to sync from genesis to latest code on testnet or mainnet, to prove that it is fully compatible with history blocks.

consensus/parlia/snapshot.go
```diff
@@ -289,18 +342,22 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
...
+					log.Debug("Recents are cleared up", "blockNumber", number)
 				}
-			}
-			oldLimit = len(snap.Validators)
-			newLimit = len(newVals)
-			if newLimit < oldLimit {
-				for i := 0; i < oldLimit-newLimit; i++ {
-					delete(snap.RecentForkHashes, number-uint64(newLimit)-uint64(i))
+			} else {
+				oldLimit := len(snap.Validators)/2 + 1
```
unclezoro, line 355: seems not exact same with previous logic?

consensus/parlia/snapshot.go
```diff
@@ -289,18 +342,22 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
...
-			newLimit := len(newVals)/2 + 1
-			if newLimit < oldLimit {
-				for i := 0; i < oldLimit-newLimit; i++ {
-					delete(snap.Recents, number-uint64(newLimit)-uint64(i))
+			if chainConfig.IsBohr(header.Number, header.Time) {
+				epochClearKey := math.MaxUint64 - header.Number.Uint64()/s.config.Epoch // impossible used as a block number
+				// in a epoch, after the first validator set switch, minerHistoryCheckLen() may become larger,
+				// so the unexpected second switch will happen, don't clear up the `Recents` in this kind of scene.
+				if _, ok := snap.Recents[epochClearKey]; !ok {
+					snap.Recents = make(map[uint64]common.Address) // without this logic, there will be several off-turn blocks when do validator set switch
```
unclezoro, line 350: I dont think we should empty `Recents` when we do epoch switch

consensus/parlia/snapshot.go
NathanBSC: > I will suggest run a fullnode to use this version of code to sync from genesis to latest code on testnet or mainnet, to prove that it is fully compatible with history blocks.

accepted

consensus/parlia/bohrFork.go
NathanBSC: accepted

consensus/parlia/parlia.go
NathanBSC: accepted
the price is to add conversion when calculated with other variables

consensus/parlia/snapshot.go
NathanBSC: accepted

consensus/parlia/bohrFork.go
NathanBSC: accepted

consensus/parlia/bohrFork.go
NathanBSC: > just return 1 instead of nil? so that we can still use `uint8` instead of `*uint8` to make it simple

if change the return value type from `*uint8` to `uint8`

then in func `getTurnTerm`, will also return a meaningful value when error happens,
for example in the line 31 in func `getTurnTerm`
so return pointer is better IMO

consensus/parlia/parlia.go
NathanBSC: > can you change the `TestImpactOfValidatorOutOfService` to simulate the impact of a validator is down and how long it will take after rejoin.

accepted

consensus/parlia/snapshot.go
NathanBSC: a node is running fullsync againse chapel net, almost catch up the newest block

NathanBSC:
force push after rebase develop to solve conflicts 06.13

consensus/parlia/snapshot.go
NathanBSC: > seems not exact same with previous logic?

same, the previous line is 292, not 299

consensus/parlia/abi.go
```diff
@@ -2306,6 +2306,19 @@ const validatorSetABI = `
     ],
     "stateMutability": "view"
   },
+  {
+    "inputs": [],
+    "name": "getTurnTerm",
```
zzzckck, line 2311: what does "TurnTerm" stand for?

NathanBSC:
force push after rebase develop 06.19

NathanBSC:
force push after rebase develop 06.21

NathanBSC:
force push after rebase develop 06.27

consensus/parlia/abi.go
NathanBSC: > how about change it to `getInTurnLength`? I am not sure if there is such a word: `turnterm`😄

accepted and updated

NathanBSC:
apply logic https://github.com/bnb-chain/bsc/pull/2558, force push

melissamforbs (approved):

NathanBSC:
s := rand.NewSource(int64(snap.Number))
-->
s := rand.NewSource(int64(snap.Number / uint64(snap.TurnLength)))

galaio (approved):

cmd/extradump/extradump_test.go
zzzckck, line 1: why would it need to change the extradump test in BEP-341?

consensus/parlia/parlia.go
```diff
@@ -1441,10 +1508,19 @@ func (p *Parlia) Delay(chain consensus.ChainReader, header *types.Header, leftOv
...
-		delay = half
+	// The blocking time should be no more than half of period when snap.TurnLength == 1
+	timeForMining := time.Duration(p.config.Period) * time.Second / 2
+	if snap.TurnLength > 1 {
+		if snap.lastBlockInOneTurn(header.Number.Uint64()) {
+			// To ensure that the next validator produces and broadcasts blocks in a timely manner,
+			// set the value of timeForMining to a small amount
+			timeForMining = 500 * time.Millisecond
+		} else {
+			timeForMining = time.Duration(p.config.Period) * time.Second * 2 / 3
```
zzzckck, line 1519: why 2/3, to reserve some buffer?

consensus/parlia/snapshot.go
```diff
@@ -211,17 +218,45 @@ func (s *Snapshot) updateAttestation(header *types.Header, chainConfig *params.C
...
+	if s.Number > checkHistoryLength {
+		leftHistoryBound = s.Number - checkHistoryLength
+	}
+	counts := make(map[common.Address]uint8, len(s.Validators))
 	for seen, recent := range s.Recents {
-		if recent == validator {
-			if limit := uint64(len(s.Validators)/2 + 1); s.Number+1 < limit || seen > s.Number+1-limit {
-				return true
-			}
+		if seen <= leftHistoryBound || recent == (common.Address{}) {
```
zzzckck, line 237: `seen < leftHistoryBound`? equal to leftHistoryBound would be seen as recents;
take it for example:
```
s.TurnLength = 1
len(s.Validators) = 21
minerHistoryCheckLen() = 10
s.Number = 110
leftHistoryBound = 100
```
in this case block number 100 would be counted as one of the recent sign

zzzckck: `recent == (common.Address{})`  maybe we can add a warning log, as it is abnormal.

consensus/parlia/snapshot.go
```diff
@@ -248,10 +283,10 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 	for _, header := range headers {
 		number := header.Number.Uint64()
 		// Delete the oldest validator from the recent list to allow it signing again
-		if limit := uint64(len(snap.Validators)/2 + 1); number >= limit {
+		if limit := snap.minerHistoryCheckLen() + 1; number >= limit { // `+1` for genesis block which has no miner
```
zzzckck, line 286: `// `+1` for genesis block which has no miner`
what does it mean?

consensus/parlia/snapshot.go
```diff
@@ -262,16 +297,22 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
 		if _, ok := snap.Validators[validator]; !ok {
 			return nil, errUnauthorizedValidator(validator.String())
 		}
-		for _, recent := range snap.Recents {
-			if recent == validator {
+		if chainConfig.IsBohr(header.Number, header.Time) {
+			if snap.SignRecently(validator) {
 				return nil, errRecentlySigned
 			}
+		} else {
```
zzzckck, line 304: maybe we don't need this if/else, as `snap.SignRecently(validator)` also works for turnLength == 1?

consensus/parlia/snapshot.go
```diff
@@ -262,16 +297,22 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
...
+					return nil, errRecentlySigned
+				}
+			}
 		}
 		snap.Recents[number] = validator
 		snap.RecentForkHashes[number] = hex.EncodeToString(header.Extra[extraVanity-nextForkHashSize : extraVanity])
 		snap.updateAttestation(header, chainConfig, s.config)
 		// change validator set
-		if number > 0 && number%s.config.Epoch == uint64(len(snap.Validators)/2) {
+		if number > 0 && number%s.config.Epoch == snap.minerHistoryCheckLen() {
```
zzzckck, line 315: why not `snap.minerHistoryCheckLen() + 1` to change validator set?

consensus/parlia/snapshot.go
```diff
@@ -315,20 +367,16 @@ func (s *Snapshot) apply(headers []*types.Header, chain consensus.ChainHeaderRea
...
-				}
-			}
 			snap.Validators = newVals
 			if chainConfig.IsLuban(header.Number) {
 				validators := snap.validators()
 				for idx, val := range validators {
 					snap.Validators[val].Index = idx + 1 // offset by 1
 				}
 			}
+			for i := snap.versionHistoryCheckLen(); i < oldVersionsLen; i++ {
```
zzzckck, line 377: this `apply(...)` is not code review friendly...

consensus/parlia/snapshot.go
```diff
@@ -346,17 +394,20 @@ func (s *Snapshot) validators() []common.Address {
...
-	offset := (s.Number + 1) % uint64(len(validators))
-	return validators[offset] == validator
+	return s.inturnValidator() == validator
 }
 
 // inturnValidator returns the validator at a given block height.
 func (s *Snapshot) inturnValidator() common.Address {
 	validators := s.validators()
-	offset := (s.Number + 1) % uint64(len(validators))
+	offset := (s.Number + 1) / uint64(s.TurnLength) % uint64(len(validators))
```
zzzckck, line 410: `(s.Number + s.TurnLength) / uint64(s.TurnLength)`?

buddh0: > `(s.Number + s.TurnLength) / uint64(s.TurnLength)`?

when at height `s.Number + 1` ,  we can get snapshot with s.Number.
the function name is not quite clear.

consensus/parlia/snapshot.go
buddh0: the comment is not right. I will delete it

consensus/parlia/snapshot.go
buddh0: if snap.Recents has the right length, then the `if/else` is not need.
but snap.Recents may contain more elements than need, as happened [here](https://github.com/bnb-chain/bsc/pull/1344/files)

consensus/parlia/snapshot.go
buddh0: > `recent == (common.Address{})` maybe we can add a warning log for it, as it is abnormal.

when seen == `epochKey`
I will update the comment in codes

buddh0: > `seen < leftHistoryBound`? equal to leftHistoryBound would be seen as recents; take it for example:
> 
> ```
> s.TurnLength = 1
> len(s.Validators) = 21
> minerHistoryCheckLen() = 10
> s.Number = 110
> leftHistoryBound = 100
> ```
> 
> in this case block number 100 would be counted as one of the recent sign

no, 100 is excluded
(101,110] is the range, and the length is 10

consensus/parlia/parlia.go
buddh0: > why 2/3, to reserve some buffer?
yes.
the final value need to be decided after benchmark,
but if we set it to 3 seconds, the network will not stable

consensus/parlia/parlia.go
```diff
@@ -1441,10 +1508,19 @@ func (p *Parlia) Delay(chain consensus.ChainReader, header *types.Header, leftOv
...
-	half := time.Duration(p.config.Period) * time.Second / 2
-	if delay > half {
-		delay = half
+	// The blocking time should be no more than half of period when snap.TurnLength == 1
+	timeForMining := time.Duration(p.config.Period) * time.Second / 2
+	if snap.TurnLength > 1 {
+		if snap.lastBlockInOneTurn(header.Number.Uint64()) {
+			// To ensure that the next validator produces and broadcasts blocks in a timely manner,
+			// set the value of timeForMining to a small amount
+			timeForMining = 500 * time.Millisecond
```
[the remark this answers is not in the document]
buddh0: > why 500ms for the last block? too small?

the next in turn validator may lagging behind, so make the last block small, then the next in turn validator can catch up, then also produce a small block timely.
if the next in turn validator is not lagging behind, it's also ok, the next in turn validator have 2 seconds+ to produce a block, it  will produce a big block.

consensus/parlia/snapshot.go
buddh0: > why not `snap.minerHistoryCheckLen() + 1` to change validator set?

when turnLenght==1,  snap.minerHistoryCheckLen() == uint64(len(snap.Validators)/2), exactly the same as before.

why not `+1` here? 
for example when turnLenght ==1,
[200, 210] contains 11 validators, is a majority of validators, so can do the change at 210, not wait to 211.

consensus/parlia/snapshot.go
buddh0: we can compare the logic about `deleting RecentForkHashes`,
the effect is the same as before, but the code is much more simplified

consensus/parlia/snapshot.go
buddh0: > > `(s.Number + s.TurnLength) / uint64(s.TurnLength)`?
> 
> when at height `s.Number + 1` , we can get snapshot with s.Number. the function name is not quite clear. I will update the comments of the function

galaio (approved):

unclezoro (approved):

consensus/parlia/snapshot.go
zzzckck: (s.Number + s.TurnLength) == (s.Number + 1 + (s.TurnLength -1))

NathanBSC: > (s.Number + s.TurnLength) == (s.Number + 1 + (s.TurnLength -1))

not enable the first validator after switch go generate s.TurnLength blocks

PR #2484, opened 2024-05-21
fix: fix inspect database error

jingjunLi:
### Description

When executing inspect, an error in judgment causes block data to be repeatedly traversed, resulting in a large amount of data being marked as unaccounted.

### Rationale

In Go, directly comparing two interface values for equality is generally unsafe, especially for complex types like the database interface ethdb.Database. This is because interface comparison checks not only their values but also whether their dynamic types match.

Add the `HasSeparateBlockStore` interface to indicate whether a separate block store is enabled.

### Example

### Changes

Notable changes: 
* ...


flywukong (approved):
LGTM

sysvm (approved):
LGTM

PR #2507, opened 2024-06-06
fix: waiting for the last simulation before pick best bid

irrun:
### Description

fix: waiting for the last simulation before pick best bid

### Rationale

The final round of bid simulation is often interrupted by stopTimer and not fully packed, making it less competitive than local miner and degrading the mev experience.

So we take two steps to optimize it:
1. do not set a stop timer when the bid is packaged;
2. when the local miner fetches the best bid, wait a while for the bid that is being packaged at the time to finish job.

This is a short-term solution. It will be addressed in the long run with a more formal fix and be put into the next stable release.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -319,13 +319,9 @@ func (b *bidSimulator) newBidLoop() {
 	// commit aborts in-flight bid execution with given signal and resubmits a new one.
 	commit := func(reason int32, bidRuntime *BidRuntime) {
 		// if the left time is not enough to do simulation, return
-		var simDuration time.Duration
-		if lastBid := b.GetBestBid(bidRuntime.bid.ParentHash); lastBid != nil && lastBid.duration != 0 {
-			simDuration = lastBid.duration
-		}
-
-		if time.Until(b.bidMustBefore(bidRuntime.bid.ParentHash)) <= simDuration*leftOverTimeRate/leftOverTimeScale {
-			log.Debug("BidSimulator: abort commit, not enough time to simulate",
+		delay := b.engine.Delay(b.chain, bidRuntime.env.header, &b.delayLeftOver)
+		if delay == nil || *delay <= 0 {
+			log.Info("BidSimulator: abort commit, not enough time to simulate",
 				"builder", bidRuntime.bid.Builder, "bidHash", bidRuntime.bid.Hash().Hex())
 			return
 		}
@@ -370,6 +366,7 @@ func (b *bidSimulator) newBidLoop() {
 				expectedValidatorReward: expectedValidatorReward,
 				packedBlockReward:       big.NewInt(0),
 				packedValidatorReward:   big.NewInt(0),
+				finished:                make(chan struct{}),
 			}
 
 			simulatingBid := b.GetSimulatingBid(newBid.ParentHash)
@@ -530,7 +527,6 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 
 	// ensure simulation exited then start next simulation
 	b.SetSimulatingBid(parentHash, bidRuntime)
-	start := time.Now()
 
 	defer func(simStart time.Time) {
 		logCtx := []any{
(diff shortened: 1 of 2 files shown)


miner/bid_simulator.go
```diff
@@ -29,15 +29,6 @@ import (
...
 	maxBidPerBuilderPerBlock = 3
-
-	// leftOverTimeRate is the rate of left over time to simulate a bid
-	leftOverTimeRate = 11
-	// leftOverTimeScale is the scale of left over time to simulate a bid
-	leftOverTimeScale = 10
-)
-
-var (
-	bidSimTimer = metrics.NewRegisteredTimer("bid/sim/duration", nil)
```
unclezoro, line 40: why delete this metrics?

miner/bid_simulator.go
```diff
@@ -335,7 +325,7 @@ func (b *bidSimulator) newBidLoop() {
 		interruptCh = make(chan int32, 1)
 		select {
 		case b.simBidCh <- &simBidReq{interruptCh: interruptCh, bid: bidRuntime}:
-			log.Debug("BidSimulator: commit", "builder", bidRuntime.bid.Builder, "bidHash", bidRuntime.bid.Hash().Hex())
+			log.Info("BidSimulator: commit", "builder", bidRuntime.bid.Builder, "bidHash", bidRuntime.bid.Hash().Hex())
```
unclezoro, line 328: It is better change back to debug level before we merge it.

miner/bid_simulator.go
```diff
@@ -553,19 +542,16 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		}
 
 		b.RemoveSimulatingBid(parentHash)
-		bidSimTimer.UpdateSince(start)
 
 		if success {
-			bidRuntime.duration = time.Since(simStart)
```
unclezoro, line 559: we can keep this info

miner/bid_simulator.go
```diff
@@ -710,8 +696,6 @@ type BidRuntime struct {
 
 	packedBlockReward     *big.Int
 	packedValidatorReward *big.Int
-
-	duration time.Duration
```
unclezoro, line 714: we can keep this info

miner/worker.go
```diff
@@ -1336,10 +1337,23 @@ LOOP:
 	// when in-turn, compare with remote work.
 	from := bestWork.coinbase
 	if w.bidFetcher != nil && bestWork.header.Difficulty.Cmp(diffInTurn) == 0 {
+		waitBidTimer := time.NewTimer(w.config.DelayLeftOver / 5)
```
unclezoro, line 1340: can you check the real `DelayLeftOver` config on validator? Is it zero?

miner/worker.go
```diff
@@ -1336,10 +1337,23 @@ LOOP:
...
+	LOOP_WAIT_BID:
+		for {
+			select {
+			case <-waitBidTimer.C:
+				break LOOP_WAIT_BID
+			default:
+				if w.bidFetcher.GetSimulatingBid(bestWork.header.ParentHash) == nil {
```
unclezoro, line 1347: This is not good, the CPU is busy with doing `w.bidFetcher.GetSimulatingBid(bestWork.header.ParentHash)`, this may starve other thread.

miner/bid_simulator.go
irrun: reverted

miner/bid_simulator.go
irrun: reverted

miner/worker.go
irrun: fixed

miner/worker.go
irrun: fixed

miner/bid_simulator.go
irrun: reverted

unclezoro (approved):

zzzckck (approved):

PR #2508, opened 2024-06-06
perf: speedup pbss trienode read

will-2012:
### Description

In the native transfer test scenario, we found that the bottleneck of pbss trienode reading is the recursive query of the 128-layer difflayer. The current PR is mainly to optimize the pbss trienode reading latency.

### Rationale

In difflayer, a new cache map with node hash as key is added. The Node interface gives priority to accessing the cache. If a hit is found, it is returned directly. If it is not found, the disklayer is queried. Compared with the recursive query of the 128-layer difflayer, this cache query is O(1). The original query path is optimized, and most of them are fastpath.
The cache is added when the difflayer is added to the layertree and is released when the difflayer is merged into the disklayer.

### Performance

image

pbss+cache has better performance, and the optimization effects are as follows: 
* As a validator test scenario, 1 million accounts were randomly selected from 25 million accounts, and native transfer transactions were performed at 600qps. The difflayer cache optimization effect was significant, and the validation phase was reduced from 450ms to 100ms.
* As a test scenario for bsc fullnode, starting from block 34.1 million, the validate node was reduced from 73ms to 54ms, which accounts for a small proportion of the entire import block, so the overall optimization effect is not obvious.

### Changes

Notable changes: 
* Pbss difflayer: add hash cache.
* Pbss disklayer: adjust clean cache size.


eth/backend.go
```diff
@@ -161,12 +161,14 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	// Optimize memory distribution by reallocating surplus allowance from the
 	// dirty cache to the clean cache.
 	if config.StateScheme == rawdb.PathScheme && config.TrieDirtyCache > pathdb.MaxDirtyBufferSize/1024/1024 {
-		log.Info("Capped dirty cache size", "provided", common.StorageSize(config.TrieDirtyCache)*1024*1024, "adjusted", common.StorageSize(pathdb.MaxDirtyBufferSize))
```
fynnss, line 164: I think this log should be kept to know that the dirty cache has been adjusted.

triedb/pathdb/asyncnodebuffer.go
```diff
@@ -226,7 +226,8 @@ func (nc *nodecache) node(owner common.Hash, path []byte, hash common.Hash) (*tr
 	}
 	if n.Hash != hash {
 		dirtyFalseMeter.Mark(1)
-		log.Error("Unexpected trie node in async node buffer", "owner", owner, "path", path, "expect", hash, "got", n.Hash)
+		// difflayer cache miss maybe hit this, which is normal case, and the caller can retry it.
+		log.Debug("Unexpected trie node in async node buffer", "owner", owner, "path", path, "expect", hash, "got", n.Hash)
```
fynnss, line 230: If the data is corrupted, there is no `Unexpected trie node` log. Why not change it to `WARN`?

triedb/pathdb/difflayer.go
```diff
@@ -26,6 +26,86 @@ import (
 	"github.com/ethereum/go-ethereum/trie/triestate"
 )
 
+type HashIndex struct {
```
fynnss, line 29: `HashIndex` is too general. I think `HashNodeCache` is more suitable.

triedb/pathdb/difflayer.go
```diff
@@ -133,7 +221,32 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, hash common.Hash, dept
...
+
+	parent := dl.parent
+	for {
+		if disk, ok := parent.(*diskLayer); ok {
+			blob, err := disk.NodeByLogger(owner, path, hash, log.Debug)
+			if err != nil {
+				// This is a bad case with a very low probability. The same trienode exists
+				// in different difflayers and can be cleared from the map in advance. In
+				// this case, the 128-layer difflayer is queried again.
+				diffHashCacheSlowPathMeter.Mark(1)
```
fynnss, line 241: I think we should add ref for every node in cache to avoid this bad case.

triedb/pathdb/difflayer.go
```diff
@@ -133,7 +221,32 @@ func (dl *diffLayer) node(owner common.Hash, path []byte, hash common.Hash, dept
...
+	if n := dl.cache.Get(hash); n != nil {
+		// The query from the hash map is fastpath,
+		// avoiding recursive query of 128 difflayers.
+		diffHashCacheHitMeter.Mark(1)
+		diffHashCacheReadMeter.Mark(int64(len(n.Blob)))
+		return n.Blob, nil
+	}
+	diffHashCacheMissMeter.Mark(1)
+
+	parent := dl.parent
```
fynnss, line 233: Try add a `origin` pointer to the disklayer in every difflayer such as it in snapshot's difflayer. 

Ref to https://github.com/bnb-chain/bsc/blob/master/core/state/snapshot/difflayer.go#L102

triedb/pathdb/disklayer.go
```diff
@@ -150,9 +150,9 @@ func (dl *diskLayer) markStale() {
 	dl.stale = true
 }
 
-// Node implements the layer interface, retrieving the trie node with the
-// provided node info. No error will be returned if the node is not found.
-func (dl *diskLayer) Node(owner common.Hash, path []byte, hash common.Hash) ([]byte, error) {
+type loggerFunc func(string, ...interface{})
+
+func (dl *diskLayer) NodeByLogger(owner common.Hash, path []byte, hash common.Hash, logger loggerFunc) ([]byte, error) {
```
fynnss, line 155: `logger` is so confusing here without any comments. I think we could remove it to keep it simplify.

fynnss: And if the logic of querying 128 difflayer again above is removed, it is not necessary here too.

triedb/pathdb/difflayer.go
joeylichang: in pathdb, generate new disklayer ,it is used for handling reorg, so difflayer can not hold disklayer pointer.

fynnss: Modify the `origin` pointer in the Cap().

eth/backend.go
will-2012: Updated.

triedb/pathdb/asyncnodebuffer.go
will-2012: Updated.

triedb/pathdb/difflayer.go
will-2012: Updated.

triedb/pathdb/difflayer.go
will-2012: Updated.

triedb/pathdb/difflayer.go
will-2012: Updated.

triedb/pathdb/disklayer.go
will-2012: Updated.

sysvm (approved):
LGTM

joeylichang (approved):

fynnss (approved):
LGTM

RenRick (approved):
LGTM

PR #2525, opened 2024-06-21
fix: ensure empty withdrawals after cancun before broadcast

NathanBSC:
### Description

fix: ensure empty withdrawals after cancun before broadcast

### Rationale

no matter get a new block directly or by fetching

both need to ensure empty withdrawals after cancun before broadcast

### Example

### Changes

Notable changes: 
* ...


zlacfzy (approved):

galaio (approved):

unclezoro (changes requested):
Do you think it is necessary to change the code of `downloader` too?

NathanBSC:
> Do you think it is necessary to change the code of `downloader` too?

data from

zzzckck (approved):

unclezoro (approved):

PR #2538, opened 2024-06-25
feat: enhance bid comparison and reply bidding results && detail logs

Jolly23:
### Description

* mev_sendBid replies current bestBid reward if it's discard due to low reward (but it still cost 1 submit chance for builder).
* The thread safe transaction set is no need for GreedyMergeTx. The change will lift extensive locking mechanisms.
Add builder bid arriving and comparing log into standard log output.
* Simplified bidRuntime comparison by moving comparing function into bidRuntime.
* add bid arrive/comparsion detail logs,  helps validator records the builder bid operation, to better understand and get involve the BPS mechanisms.

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -737,7 +737,6 @@ type BidRuntime struct {
 }
 
 func newBidRuntime(newBid *types.Bid, validatorCommission uint64) (*BidRuntime, error) {
-
 	// check the block reward and validator reward of the newBid
 	expectedBlockReward := newBid.GasFee
 	expectedValidatorReward := new(big.Int).Mul(expectedBlockReward, big.NewInt(int64(validatorCommission)))
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -334,8 +334,8 @@ func (b *bidSimulator) newBidLoop() {
 		}
 	}
 
-	genReplyReason := func(betterBid *BidRuntime) error {
-		return fmt.Errorf("bid discarded, current bestBid is [blockReward: %s, validatorReward: %s]", betterBid.expectedBlockReward, betterBid.expectedValidatorReward)
+	genDiscardedReply := func(betterBid *BidRuntime) error {
+		return fmt.Errorf("bid is discarded, current bestBid is [blockReward: %s, validatorReward: %s]", betterBid.expectedBlockReward, betterBid.expectedValidatorReward)
 	}
 
 	for {
@@ -360,7 +360,7 @@ func (b *bidSimulator) newBidLoop() {
 				if bidRuntime.isExpectedBetterThan(simulatingBid) {
 					commit(commitInterruptBetterBid, bidRuntime)
 				} else {
-					replyErr = genReplyReason(simulatingBid)
+					replyErr = genDiscardedReply(simulatingBid)
 				}
 			} else {
 				// bestBid is nil means the bid is the first bid, otherwise the bid should compare with the bestBid
@@ -368,7 +368,7 @@ func (b *bidSimulator) newBidLoop() {
 					bidRuntime.isExpectedBetterThan(bestBid) {
 					commit(commitInterruptBetterBid, bidRuntime)
 				} else {
-					replyErr = genReplyReason(bestBid)
+					replyErr = genDiscardedReply(bestBid)
 				}
 			}
 
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -347,6 +347,9 @@ func (b *bidSimulator) newBidLoop() {
 
 			bidRuntime, err := newBidRuntime(newBid.bid, b.config.ValidatorCommission)
 			if err != nil {
+				if newBid.feedback != nil {
+					newBid.feedback <- err
+				}
 				continue
 			}
 
@@ -747,7 +750,7 @@ func newBidRuntime(newBid *types.Bid, validatorCommission uint64) (*BidRuntime, 
 		// damage self profit, ignore
 		log.Debug("BidSimulator: invalid bid, validator reward is less than 0, ignore",
 			"builder", newBid.Builder, "bidHash", newBid.Hash().Hex())
-		return nil, errors.New("validator reward is less than 0")
+		return nil, fmt.Errorf("validator reward is less than 0, value: %s, commissionConfig: %d", expectedValidatorReward, validatorCommission)
 	}
 
 	bidRuntime := &BidRuntime{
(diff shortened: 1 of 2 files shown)


miner/bid_simulator.go
```diff
@@ -327,67 +334,55 @@ func (b *bidSimulator) newBidLoop() {
...
-			expectedValidatorReward := new(big.Int).Mul(expectedBlockReward, big.NewInt(int64(b.config.ValidatorCommission)))
-			expectedValidatorReward.Div(expectedValidatorReward, big.NewInt(10000))
-			expectedValidatorReward.Sub(expectedValidatorReward, newBid.BuilderFee)
-
-			if expectedValidatorReward.Cmp(big.NewInt(0)) < 0 {
-				// damage self profit, ignore
-				log.Debug("BidSimulator: invalid bid, validator reward is less than 0, ignore",
-					"builder", newBid.Builder, "bidHash", newBid.Hash().Hex())
+			bidRuntime, err := newBidRuntime(newBid.bid, b.config.ValidatorCommission)
+			if err != nil {
```
irrun, line 349: here could also reply reason

miner/bid_simulator.go
```diff
@@ -442,10 +440,14 @@ func (b *bidSimulator) clearLoop() {
...
+	b.AddPending(bid.BlockNumber, bid.Builder, bid.Hash())
+
 	select {
-	case b.newBidCh <- bid:
-		b.AddPending(bid.BlockNumber, bid.Builder, bid.Hash())
-		return nil
+	case reply := <-replyCh:
+		return reply
 	case <-timer.C:
 		return types.ErrMevBusy
```
irrun, line 452: Line 445 may be blocked when newBidCh is busy, how about using two select to avoid it
```
replyCh := make(chan error)

select {
case b.newBidCh <- newBidPackage{bid: bid, feedback: replyCh}:
b.AddPending(bid.BlockNumber, bid.Builder, bid.Hash())
case <-timer.C:
return types.ErrMevBusy
}

select {
case reply := <-replyCh:
return reply
case <-timer.C:
return types.ErrMevBusy
}
```

irrun (dismissed):

irrun (dismissed):

unclezoro (changes requested):
hi, We will have hot fix release soon, as this PR is not a must to have feature, so we will hold this PR after the hot fix. It may take 2 weeks.

miner/bid_simulator.go
```diff
@@ -442,10 +440,19 @@ func (b *bidSimulator) clearLoop() {
...
+	case b.newBidCh <- newBidPackage{bid: bid, feedback: replyCh}:
 		b.AddPending(bid.BlockNumber, bid.Builder, bid.Hash())
-		return nil
+	case <-timer.C:
+		return types.ErrMevBusy
+	}
+
+	select {
+	case reply := <-replyCh:
+		return reply
```
unclezoro, line 455: because the `replyCh` get no buffer, if we return here directly, it is possible another routinue will be blocked when sending item into `replyCh` and thus cause dead lock.

miner/bid_simulator.go
```diff
@@ -327,67 +334,58 @@ func (b *bidSimulator) newBidLoop() {
...
-			expectedValidatorReward.Sub(expectedValidatorReward, newBid.BuilderFee)
-
-			if expectedValidatorReward.Cmp(big.NewInt(0)) < 0 {
-				// damage self profit, ignore
-				log.Debug("BidSimulator: invalid bid, validator reward is less than 0, ignore",
-					"builder", newBid.Builder, "bidHash", newBid.Hash().Hex())
+			bidRuntime, err := newBidRuntime(newBid.bid, b.config.ValidatorCommission)
+			if err != nil {
+				if newBid.feedback != nil {
+					newBid.feedback <- err
```
unclezoro, line 351: please ensure this will never block.

miner/bid_simulator.go
```diff
@@ -327,67 +334,58 @@ func (b *bidSimulator) newBidLoop() {
...
+
+				log.Info("[BID ARRIVED]",
+					"block", newBid.bid.BlockNumber,
+					"builder", newBid.bid.Builder,
+					"accepted", replyErr == nil,
+					"blockReward", weiToEtherStringF6(bidRuntime.expectedBlockReward),
+					"validatorReward", weiToEtherStringF6(bidRuntime.expectedValidatorReward),
+					"tx", len(newBid.bid.Txs),
+					"hash", newBid.bid.Hash().TerminalString(),
+				)
```
unclezoro, line 386: It is suggested to send a `nil` error back to `newBid.feedback` if no error happen, so that `sendBid` can return ASAP

miner/bid_simulator.go
```diff
@@ -633,7 +643,7 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 	if b.config.GreedyMergeTx {
 		delay := b.engine.Delay(b.chain, bidRuntime.env.header, &b.delayLeftOver)
 		if delay != nil && *delay > 0 {
-			bidTxsSet := mapset.NewSet[common.Hash]()
+			bidTxsSet := mapset.NewThreadUnsafeSetWithSize[common.Hash](len(bidRuntime.bid.Txs))
```
unclezoro, line 646: what is the motivation to change the hash set type?

irrun: for no concurrent operation on bidTxsSet

miner/bid_simulator.go
irrun: if no error happen, it's nil.

irrun (approved):

miner/bid_simulator.go
```diff
@@ -535,19 +544,20 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
 		if success {
 			bidRuntime.duration = time.Since(simStart)
 			bidSimTimer.UpdateSince(simStart)
+			metrics.GetOrRegisterCounter(fmt.Sprintf("bid/sim/count/%d", bidRuntime.bid.BlockNumber), nil).Inc(1)
```
zzzckck, line 547: it could have impact on metric efficiency, as each block would have registry a metric, there would be ~28,800(24*60*60/3) metrics registered per day, it is too much, it is not recommended.

miner/bid_simulator.go
zzzckck: check this note?

miner/bid_simulator.go
irrun: ok，removed

zzzckck (approved):

irrun (approved):

galaio (approved):

unclezoro (approved):

PR #2545, opened 2024-06-27
feat: add mev bid trace log

irrun:
### Description

feat: add mev bid trace log

### Rationale

### Example

### Changes

Notable changes: 
* ...


unclezoro (changes requested):
Need test on testnet.

PR #2561, opened 2024-07-03
tests: fix unstable test

MatusKysel:
### Description

Added dynamic sleep based on number of snap peers = that handlers are working.
Tested it locally with ```go test -p 1 -v -run TestSnapSyncWithBlobs -count 100```
### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck (approved):
maybe it could solve the random failure, let's give it a try.

emailtovamos (approved):

PR #2562, opened 2024-07-04
fix: online prune of ancient delete unexpected blocks

jingjunLi:
### Description

The first time you run with the --pruneancient option, a BAD BLOCK error may occur.

image

### Rationale

When pruneancient is enabled, it deletes the CanonicalHash previously stored in the ancient database. This results in the loss of some CanonicalHash entries in the range [first, frozen) during the subsequent loop.
image

 In gcKvStore, when attempting to delete CanonicalHash and Header, the ancients and numbers do not match, preventing the proper deletion.
> For example, if first in the loop is 60002, the range being traversed is [60002, 90003). Due to the deletion of CanonicalHash in the range [60002, 61298], the data in ancients is actually [61299, 90003).
In gcKvStore, the number range [60002, 88706) is being deleted, while the corresponding hash range is [61299, 90003), causing the construction of incorrect block keys and failing to delete the actual blocks.
number[60002] + hash[61299]
number[60003] + hash[61300]

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/rawdb/prunedfreezer.go
+++ b/core/rawdb/prunedfreezer.go
@@ -299,9 +299,8 @@ func (f *prunedfreezer) freeze() {
 				log.Error("Append ancient err", "number", f.frozen, "hash", hash, "err", err)
 				break
 			}
-			if hash != (common.Hash{}) {
-				ancients = append(ancients, hash)
-			}
+			// may include common.Hash{}, will be delete in gcKvStore
+			ancients = append(ancients, hash)
 		}
 		// Batch of blocks have been frozen, flush them before wiping from leveldb
 		if err := f.Sync(); err != nil {


core/rawdb/prunedfreezer.go
```diff
@@ -285,7 +285,7 @@ func (f *prunedfreezer) freeze() {
 		var (
 			start    = time.Now()
 			first    = f.frozen
-			ancients = make([]common.Hash, 0, limit-f.frozen)
+			ancients = make(map[uint64]common.Hash)
```
zzzckck, line 288: we can keep the code of initialise map's size:
```
ancients = make(map[uint64]common.Hash, limit-f.frozen)
```

zzzckck (approved):

RenRick (approved):
LGTM

joeylichang (approved):

PR #2568, opened 2024-07-08
core/vote: not vote if too late for next in turn validator

NathanBSC:
### Description

core/vote: not vote if too late for next in turn validator

### Rationale
image

Assume that at block height N, validator Va mines a large block BN, broadcasts it, and votes immediately. Due to the block's large size, other validators (such as Vb, Vc, and Vd) take a longer time to import BN, resulting in Vc and Vd voting for BN later. As a result, at block height N+1, Vb fails to collect enough votes for BN and mines and broadcasts a block BN1 without vote proof. Subsequently, backup validator Vc collects enough votes for BN, mines, and broadcasts a block BN1' with vote proof, causing other validators to experience a block reorganization (reorg).

this PR suggest not to vote if too late for next in turn validator

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vote/vote_manager.go
+++ b/core/vote/vote_manager.go
@@ -143,7 +143,7 @@ func (voteManager *VoteManager) loop() {
 			curHead := cHead.Block.Header()
 			if p, ok := voteManager.engine.(*parlia.Parlia); ok {
 				nextBlockMinedTime := time.Unix(int64((curHead.Time + p.Period())), 0)
-				timeForBroadcast := 100 * time.Millisecond
+				timeForBroadcast := 50 * time.Millisecond // enough to broadcast a vote
 				if time.Now().Add(timeForBroadcast).After(nextBlockMinedTime) {
 					log.Warn("too late to vote", "Head.Time(Second)", curHead.Time, "Now(Millisecond)", time.Now().UnixMilli())
 					continue
--- a/core/vote/vote_manager.go
+++ b/core/vote/vote_manager.go
@@ -145,7 +145,7 @@ func (voteManager *VoteManager) loop() {
 				nextBlockMinedTime := time.Unix(int64((curHead.Time + p.Period())), 0)
 				timeForBroadcast := 100 * time.Millisecond
 				if time.Now().Add(timeForBroadcast).After(nextBlockMinedTime) {
-					log.Debug("too late to vote", "Head.Time(Second)", curHead.Time, "Now(Millisecond)", time.Now().UnixMilli())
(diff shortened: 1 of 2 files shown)


zzzckck:
this may encourage validators to delay broadcast block?

melissamforbs (approved):

NathanBSC:
> this may encourage validators to delay broadcast block?

IMO, not.
if validator delay broadcast block, 
backup validator will broadcast their blocks and get votes.

core/vote/vote_manager.go
```diff
@@ -139,6 +141,15 @@ func (voteManager *VoteManager) loop() {
 			}
 
 			curHead := cHead.Block.Header()
+			if p, ok := voteManager.engine.(*parlia.Parlia); ok {
+				nextInTurnBlockTime := time.Unix(int64((curHead.Time + p.Period())), 0)
```
zzzckck, line 145: what if the next in-turn validator has recently signed?

core/vote/vote_manager.go
```diff
@@ -139,6 +141,15 @@ func (voteManager *VoteManager) loop() {
 			}
 
 			curHead := cHead.Block.Header()
+			if p, ok := voteManager.engine.(*parlia.Parlia); ok {
+				nextBlockMinedTime := time.Unix(int64((curHead.Time + p.Period())), 0)
+				timeForBroadcast := 100 * time.Millisecond
+				if time.Now().Add(timeForBroadcast).After(nextBlockMinedTime) {
+					log.Debug("too late to vote", "Head.Time(Second)", curHead.Time, "Now(Millisecond)", time.Now().UnixMilli())
```
zzzckck, line 148: I think we can change the log level to Info

NathanBSC: > I think we can change the log level to Warning

accepted

core/vote/vote_manager.go
```diff
@@ -139,6 +141,15 @@ func (voteManager *VoteManager) loop() {
 			}
 
 			curHead := cHead.Block.Header()
+			if p, ok := voteManager.engine.(*parlia.Parlia); ok {
+				nextBlockMinedTime := time.Unix(int64((curHead.Time + p.Period())), 0)
+				timeForBroadcast := 100 * time.Millisecond
```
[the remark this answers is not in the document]
NathanBSC: accpeted

zzzckck (approved):

unclezoro (approved):

PR #2575, opened 2024-07-12
feat: support cache among blocks

flywukong:
### Description

support a flatten cache which have the some cache function of difflayer and it is maintained among blocks. It stores the account and storage data with the newest version among blocks and help save the cache cost of difflayer.  The new cache used fastcache and the hit rate of  new cache is little better then difflayer , If new cache miss , the bloomfilter in the difflayer will stop iterative query and enters the layer3 cache directly in most cases.

### Rationale
execution performance raise 4% without shared pool, but its optimization effect has some overlap with the storage pool
image
yellow one is the performace with cache among blocks, 

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -268,9 +268,8 @@ type BlockStats struct {
 // included in the canonical one where as GetBlockByNumber always represents the
 // canonical chain.
 type BlockChain struct {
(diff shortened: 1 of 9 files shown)


beage666 (commented):
😃

allformless:
this feature has been implemented.

PR #2579, opened 2024-07-15
fix: only take non-mempool tx to calculate bid price

irrun:
### Description

define bidGasPrice as sum of (gas used * gas price ) for every non-mempool tx / by the sum of gas used for non-mempool tx.

### Rationale

In current implementation, when a builder receive a bundle, it will simulate all txs within the bundle. Bundle price = sum( gas price * gas used ) /sum(gas used ) , ensure bundle price > = 1 gwei. It seems not a good solution. Consider the following example:
There is a public mempool tx with 100 gwei, gas usage 21,000 and its a simple transfer tx. A malicious actor can submit a 0 gas price tx to bundle it up with the public mempool tx, this tx has 0 gwei gas price, it can involve very complex smart contract calls and it has a gas usage of 1,000,000. 
In that case, it will be (100\*21000+0\*1000000)/(21000+1000000)=2.05 gwei. This means this bundle can be included by builders and land on chain.
However, the malicious actor is essentially a "free rider" in this scenario, and this actor is able to "exploit" the bundle scoring algorithm and make high-gas-usage tx land on chain for free.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/bidutil"
 	"github.com/ethereum/go-ethereum/consensus"
 	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/txpool"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/log"
@@ -82,6 +83,7 @@ type bidSimulator struct {
 	delayLeftOver time.Duration
 	minGasPrice   *big.Int
 	chain         *core.BlockChain
+	txpool        *txpool.TxPool
 	chainConfig   *params.ChainConfig
 	engine        consensus.Engine
 	bidWorker     bidWorker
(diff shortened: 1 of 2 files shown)


unclezoro (changes requested):
please hold this PR for a while

unclezoro (dismissed):

buddh0:
this pr just filter non-memory tx, not solve `free rider`, I think

miner/bid_simulator.go
```diff
@@ -625,16 +628,39 @@ func (b *bidSimulator) simBid(interruptCh chan int32, bidRuntime *BidRuntime) {
...
+				if bidRuntime.env.header.BaseFee != nil {
+					effectiveTip.Add(effectiveTip, bidRuntime.env.header.BaseFee)
+				}
 
-		for _, receipt := range bidRuntime.env.receipts {
-			bidGasUsed += receipt.GasUsed
+				gasFee := new(big.Int).Mul(effectiveTip, new(big.Int).SetUint64(receipt.GasUsed))
+				bidGasFee.Add(bidGasFee, gasFee)
+
+				if tx.Type() == types.BlobTxType {
```
buddh0, line 650: don't add blobfees
otherwise 1 blob tx, gas price 1Gwei will be refused, because blobgasprice is low

buddh0 (approved):

miner/bid_simulator.go
irrun: bidGasUsed does not add blobGasUsed

unclezoro (approved):

Issue #2570, opened 2024-07-08
"bsc db compact" got stuck on a node with flag: "--pruneancient"


zzzckck:
#### System information

Geth version: `v1.4.11`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

#### Expected behaviour
db compact can succeed

#### Actual behaviour
db compact failed

#### Steps to reproduce the behaviour
1.run a node with flag: `--pruneancient`
2.run db compact: `bsc --config config.toml db compact --datadir <datadir>`
Then will be stuck and keep printing the following logs:
```
t=2024-07-08T00:21:08+0000 lvl=info msg="Set global gas cap" cap=50000000
t=2024-07-08T00:21:08+0000 lvl=info msg="Initializing the KZG library" backend=gokzg
t=2024-07-08T00:21:08+0000 lvl=info msg="Using leveldb as the backing database"
t=2024-07-08T00:21:08+0000 lvl=info msg="Allocated cache and file handles" database=/server/geth/chaindata cache="408.00 MiB" handles=4096
t=2024-07-08T00:21:23+0000 lvl=info msg="Using LevelDB as the backing database"
t=2024-07-08T00:21:23+0000 lvl=warn msg="NOTICE: You're opening a pruned disk db!"
t=2024-07-08T00:21:23+0000 lvl=info msg="Opened ancient database" database=/server/geth/chaindata/ancient/chain readonly=false frozen=40188881
t=2024-07-08T00:21:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
t=2024-07-08T00:21:23+0000 lvl=info msg="Stats before compaction"
t=2024-07-08T00:21:23+0000 lvl=info msg="Triggering compaction"
t=2024-07-08T00:22:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
t=2024-07-08T00:23:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
t=2024-07-08T00:24:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
t=2024-07-08T00:25:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
t=2024-07-08T00:26:23+0000 lvl=warn msg="Freezer need related env, may wait for a while" err="missing freezer env error"
[...]

PR #2580, opened 2024-07-15
freezer: Opt freezer env checking

galaio:
### Description

This PR adds a silent logic. For scenarios that require Freezer Env, `SetupFreezerEnv` should be called to initialize. For other scenarios, Freezer will silently wait 3 times and return, and will no longer execute freeze.

Using silent logic effectively avoids large-scale API modifications and unnecessary DB initialization operations.

It fixed this issue, https://github.com/bnb-chain/bsc/issues/2570

### Changes

Notable changes: 
* freezer: opt freezer env check logic;
* ...

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -178,19 +178,6 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 			}
 		}
 
-		// check freezer env first, it must wait a while when the env is necessary
-		err := f.checkFreezerEnv()
-		if err == missFreezerEnvErr {
-			log.Warn("Freezer need related env, may wait for a while", "err", err)
-			backoff = true
-			continue
-		}
-		if err != nil {
-			log.Error("Freezer check FreezerEnv err", "err", err)
-			backoff = true
-			continue
-		}
-
 		var (
 			frozen    uint64
 			threshold uint64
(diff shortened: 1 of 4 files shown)


buddh0 (dismissed):

core/rawdb/chain_freezer.go
```diff
@@ -282,6 +276,18 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
 				last = first + freezerBatchLimit
 			}
 		}
+
+		// check env first before chain freeze, it must wait when the env is necessary
+		if err := f.checkFreezerEnv(); err != nil {
+			f.waitEnvTimes++
+			if f.waitEnvTimes >= maxWaitFreezerEnvTimes {
```
zzzckck, line 283: agree to move it after the previous threshold check.
but better not add this try limit, in case it could take more than 3min to make it ready.

core/rawdb/chain_freezer.go
```diff
@@ -282,6 +276,18 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore) {
...
+				log.Warn("Wait freezer env too many times, skip the chain freezing.")
```
zzzckck, line 284: update the log message to avoid mis-understanding, like:
"Wait freezer env too many times, skip the chain freezing, not an issue if not import blocks(db compact, offline prune...)"

buddh0 (approved):

zzzckck (approved):

zzzckck (approved):

buddh0 (approved):

PR #2582, opened 2024-07-16
fix: remove delete and dangling side chains in prunefreezer

jingjunLi:
### Description

When using the freezer deletion strategy, some side chain blocks are not yet finalized and might still be subject to reorganization (reorg). Therefore, they cannot be deleted at this time.

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck (approved):

flywukong (approved):
LGTM

joeylichang (approved):

Issue #2588, opened 2024-07-17
hash schema: prune state failed


zzzckck:
#### System information
Geth version: `geth v1.4.11`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

Although hash schema & levelDB is no longer supported, at current stage, we still hope it can work, including operations like: `prune state`
#### Steps to reproduce the behaviour
> note: in my env, I use --triesInMemory 32, not sure if it matters or not.

1.setup a BSC full node
2.enable flag: `--pruneancient --syncmode full`
3.do prune state: `./bsc snapshot prune-state --datadir <datadir>`
will report this error and failed:
```
INFO [07-17|02:42:20.876] Smartcard socket not found, disabling    err="stat /run/pcscd/pcscd.comm: no such file or directory"
INFO [07-17|02:42:20.879] Set global gas cap                       cap=50,000,000
INFO [07-17|02:42:20.879] Initializing the KZG library             backend=gokzg
INFO [07-17|02:42:21.396] Using leveldb as the backing database
INFO [07-17|02:42:21.396] Allocated cache and file handles         database=/server/geth/chaindata cache=408.00MiB handles=4096
INFO [07-17|02:42:27.029] Using LevelDB as the backing database
WARN [07-17|02:42:27.030] NOTICE: You're opening a pruned disk db!
INFO [07-17|02:42:27.050] Opened ancient database                  database=/server/geth/chaindata/ancient/chain readonly=false frozen=40,449,972
INFO [07-17|02:42:27.074] Snapshot loaded                          diskRoot=388748..a4fa3d root=3c3eac..9565dd
INFO [07-17|02:42:27.754] Initialized state bloom                  size=2.00GiB
ERROR[07-17|02:42:31.100] Failed to prune state                    err="no snapshot paired state"
no snapshot paired state
```

PR #2602, opened 2024-07-22
fix: prune-state when specify `--triesInMemory 32`

jingjunLi:
### Description

To address and fix the issue described in [hash schema: prune state failed #2588](https://github.com/bnb-chain/bsc/issues/2588)

### Rationale

The prune-state command, when specifying --triesInMemory 32, results in a "no snapshot paired state" error, while the command runs successfully with the default value.
Previously, the fixed value of TriesInMemory was set to 128 and then changed to a configurable value. However, there are still some references in the code that have not been updated. For example, during persistence in (bc *BlockChain) Stop(), the constant value 128 is still being used instead of the configurable bc.TriesInMemory().

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1390,7 +1390,7 @@ func (bc *BlockChain) Stop() {
 		if !bc.cacheConfig.TrieDirtyDisabled {
 			triedb := bc.triedb
 			var once sync.Once
-			for _, offset := range []uint64{0, 1, TriesInMemory - 1} {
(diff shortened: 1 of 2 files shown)


zzzckck (approved):

RenRick (approved):
LGTM

Aces42020 (approved):

tawhidnazari57 (commented):
Submit

PR #2603, opened 2024-07-22
faucet: rate limit initial implementation

emailtovamos:
### Description

This PR introduces some rate limiting for faucet.
Also couple of concurrencies being introduced so that the main loop doesn't potentially get stuck in one single request.

Addressing issue: https://github.com/bnb-chain/bsc/issues/2606

### Rationale

### Example

### Changes

Notable changes: 
* ...


cmd/faucet/faucet.go
```diff
@@ -272,6 +282,21 @@ func (f *faucet) webHandler(w http.ResponseWriter, r *http.Request) {
 
 // apiHandler handles requests for Ether grants and transaction statuses.
 func (f *faucet) apiHandler(w http.ResponseWriter, r *http.Request) {
+	ip := r.RemoteAddr
+	if len(r.Header.Get("X-Forwarded-For")) > 0 {
+		ips := strings.Split(r.Header.Get("X-Forwarded-For"), ",")
+		if len(ips) > 0 {
+			ip = strings.TrimSpace(ips[len(ips)-1])
```
zzzckck, line 289: why only limit the last ip?

cmd/faucet/rate_limiter.go
```diff
@@ -0,0 +1,43 @@
...
+	i := &IPRateLimiter{
+		ips: cache,
+		r:   r,
+		b:   b,
+	}
+
+	return i, nil
+}
+
+func (i *IPRateLimiter) AddIP(ip string) *rate.Limiter {
```
zzzckck, line 29: if `AddIP` is only for internal usage, we can name it: `addIP`

cmd/faucet/rate_limiter.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	lru "github.com/hashicorp/golang-lru"
+	"golang.org/x/time/rate"
+)
+
+type IPRateLimiter struct {
+	ips *lru.Cache
+	r   rate.Limit
+	b   int
```
zzzckck, line 11: better to add some comments for `b`, it is the burst value within 5min?

cmd/faucet/faucet.go
zzzckck: there is a potential risk that `X-Forwarded-For` can be faked.

cmd/faucet/faucet.go
```diff
@@ -235,6 +238,12 @@ func newFaucet(genesis *core.Genesis, url string, ks *keystore.KeyStore, index [
 		return nil, err
 	}
 
+	// Allow 1 request per minute with burst of 5, and cache up to 10000 IPs
+	limiter, err := NewIPRateLimiter(rate.Limit(1.0), 5, 1000)
```
zzzckck, line 242: in above comment, it is 10000.

cmd/faucet/faucet.go
```diff
@@ -272,6 +282,22 @@ func (f *faucet) webHandler(w http.ResponseWriter, r *http.Request) {
...
+	ip := r.RemoteAddr
+	if len(r.Header.Get("X-Forwarded-For")) > 0 {
+		ips := strings.Split(r.Header.Get("X-Forwarded-For"), ",")
+		if len(ips) > 0 {
+			ip = strings.TrimSpace(ips[len(ips)-1])
+		}
+	}
+
+	limiter := f.limiter.GetLimiter(ip)
+	if !limiter.Allow() {
```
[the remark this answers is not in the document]
zzzckck: agree

cmd/faucet/faucet.go
```diff
@@ -625,19 +650,22 @@ func (f *faucet) loop() {
...
-				}, time.Second); err != nil {
-					log.Warn("Failed to send stats to client", "err", err)
-					conn.conn.Close()
-					continue
-				}
-				if err := send(conn, head, time.Second); err != nil {
-					log.Warn("Failed to send header to client", "err", err)
-					conn.conn.Close()
-				}
+				go func(conn *wsConn) {
```
zzzckck, line 653: would this async call help?

cmd/faucet/faucet.go
emailtovamos: Usually it corresponds to the original ip address without port. 

Yes `X-Forwarded-For` can be faked but it adds an extra step for an attacker. One extra thing I can think of is to do a validation check on the `ips` unless you have any other relevant suggestion?

cmd/faucet/faucet.go
emailtovamos: It should help. Instead of serially calling `send()` for each `conn`, if we do concurrently there's less change of the overall main loop getting stuck.

zzzckck (approved):

PR #2621, opened 2024-07-31
core: not record `zero hash` beacon block root with Parlia engine

NathanBSC:
### Description

core: not record `zero hash` beacon block root

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck (approved):
For BSC, `beaconRoot` is nil before Bohr hard fork,  and empty after Bohr.

zlacfzy (approved):

zzzckck (approved):

PR #2626, opened 2024-08-02
concensus/parlia.go: reverse iteration order

zlacfzy:
### Description

When import a lots of block, the recentSnaps cache will not be used when distributeFinalityReward.  BTW the ecrecover is take long time. Affect more in erigon stage_sync. So add more snaps to cache when apply snapshots. 

### Rationale

### Example

### Changes

Notable changes: 
* consensus/parlia.go

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -792,11 +792,10 @@ func (p *Parlia) snapshot(chain consensus.ChainHeaderReader, number uint64, hash
 		headers[i], headers[len(headers)-1-i] = headers[len(headers)-1-i], headers[i]
 	}
 
-	snap, err := snap.apply(headers, chain, parents, p.chainConfig)
+	snap, err := snap.apply(headers, chain, parents, p.chainConfig, p.recentSnaps)
 	if err != nil {
 		return nil, err
 	}
-	p.recentSnaps.Add(snap.Hash, snap)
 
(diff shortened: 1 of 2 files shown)


pythonberg1997 (dismissed):

consensus/parlia/parlia.go
```diff
@@ -1152,8 +1152,8 @@ func (p *Parlia) distributeFinalityReward(chain consensus.ChainHeaderReader, sta
 
 	head := header
 	accumulatedWeights := make(map[common.Address]uint64)
-	for height := currentHeight - 1; height+epoch >= currentHeight && height >= 1; height-- {
-		head = chain.GetHeaderByHash(head.ParentHash)
+	for height := uint64(math.Max(float64(currentHeight-epoch), 1)); height < currentHeight; height++ {
```
unclezoro, line 1155: I guess this is  a hardfork logi, we can not change the logic directly.

consensus/parlia/parlia.go
```diff
@@ -1152,8 +1152,8 @@ func (p *Parlia) distributeFinalityReward(chain consensus.ChainHeaderReader, sta
...
+		head = chain.GetHeaderByNumber(height - 1)
```
unclezoro, line 1156: `GetHeaderByNumber` may return different result because of fork.  `GetHeaderByHash` is safe than `GetHeaderByNumber`

PR #2636, opened 2024-08-07
chore: add the wrapper layer of versa

joey0612:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


core/state/caching_versa_db.go
```diff
@@ -0,0 +1,297 @@
...
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
+	"github.com/ethereum/go-ethereum/trie/trienode"
+	"github.com/ethereum/go-ethereum/triedb"
+)
+
+type cachingVersaDB struct {
+	triedb        *triedb.Database
+	versionDB     versa.Database
```
RenRick, line 24: Why to have a "versionDB" here? Is there any difference with "triedb"?

core/state/caching_versa_db.go
```diff
@@ -0,0 +1,297 @@
...
+	codeDB        ethdb.KeyValueStore
+	codeSizeCache *lru.Cache[common.Hash, int]
+	codeCache     *lru.SizeConstrainedCache[common.Hash, []byte]
+
+	accTree  *VersaTree
```
RenRick, line 29: Where is VersaTree impl? And why to have a "accTree"?

core/state/caching_versa_db.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (cv *cachingVersaDB) OpenTrie(root common.Hash) (Trie, error) {
+	if cv.hasState.Load() {
+		//TODO:: will change to log.Error after stabilization
+		panic("account tree has open")
+	}
+
+	// TODO:: if root tree, versa db shouldb ignore check version
+	state, err := cv.versionDB.OpenState(0, root, cv.mode)
```
fynnss, line 55: Suggest to use the block num as the version here, not 0.

core/state/caching_versa_db.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (vt *VersaTree) GetKey(key []byte) []byte {
+	_, val, err := vt.db.Get(vt.handler, key)
+	if err != nil {
+		log.Warn("failed to get key from version db")
+	}
+	return val
+}
+
+func (vt *VersaTree) GetAccount(address common.Address) (*types.StateAccount, error) {
```
fynnss, line 206: I think the version of the account should be returned to the caller, and maintained by the account itself, so that it can be provided to versaDB when opening the storage tree.

But if do this, a lot of changes will be required. Therefore, we need to do some trade off whether the version should be known to the caller.

core/state/caching_versa_db.go
joeylichang: There is no essential difference; here, versiondb is for convenience, while triedb is for potential future needs.

core/state/caching_versa_db.go
joeylichang: Facilitate OpenStorageTree to avoid reopen.

core/state/caching_versa_db.go
joeylichang: It’s fine, but currently, this version is of no use to the upper layer and has brought about a burden. The plan is to make the upper layer aware of the version only if needed in the future. The conclusion is that it can be transparent to the upper layer for the time being.

core/state/caching_versa_db.go
joeylichang: Similarly, currently the upper layer does not have a strong dependency on the version, so it can be transparent to the upper layer for the time being. If needed in the future, it can be exposed then. The focus at this stage should not be on this.

mtrackeros:
> ### Description
> add a description of your changes here...
> 
> ### Rationale
> tell us why we need these changes...
> 
> ### Example
> add an example CLI or API response...
> 
> ### Changes
> Notable changes:
> 
> * add each change in a bullet point here
> * ...

allformless:
developing stopped for two years, close it for now

PR #2660, opened 2024-08-20
core/txpool/legacypool: add overflowpool for txs

emailtovamos:
### Description
This PR incorporates an Overflow Pool which stores transactions in memory for such cases and time to time transfers them to the Main Pool when the Main Pool isn't full. It is not propagated or announced while it is in the Overflow Pool. But when it is finally transferred to the Main Pool, at that point it is considered as a transaction that just arrived at the Main Pool and is subject to the usual checks and propagation logic.
Overflow Pool's size can be chosen by a node depending on its hardware capacity.

### Rationale
During high traffic, transactions are lost as the current legacy pool (Main Pool) has limits on its capacity and propagating them during high traffic also costs network congestions. So this is an attempt to solve 'tx lost' and get better UX.

#### a. Main Pool
There is no new structure for Main Pool, actually it is just the current Transaction Pool (Legacy Pool). When this Pool isn't full, the received new transactions will be broadcasted to all the relevant peers, just the same as the current behaviour.

#### b. Overflow Pool: the local buffer (LRU or Heap)
When the Main Pool is overflowed during high traffic, then it would put lots of stress on the network if we keep broadcasting new transactions. So we put any new transaction into the Overflow Pool and don’t broadcast or announce.
The size of the Overflow Pool could be very large, in order to hold a large traffic volume, like 500K. Suppose the average transaction size is 500B, it will take around 256MB memory, which is acceptable.

#### How to flush transactions from Overflow Pool to Main Pool:

- To satisfy the first-come-first-serve policy, when new transactions arrive & the Main Pool is full, we will push it to the Overflow Pool if it is not full.
- This transaction that is part of Overflow Pool doesn’t get announced or propagated. Its knowledge and existence is completely local to the particular node.
- We regularly try to pop new transactions from the Overflow Pool to add to the Main Pool when Main Pool has space. This can either be time-based (e.g. once every 10 seconds) or block based (once every new block or every 2 blocks)
- When a transaction moves from Overflow Pool to Main Pool, it is treated the same way when a transaction directly arrives in the Main Pool.
- When a new block is imported, we can try to flush the Overflow Pool to Main Pool.


cmd/geth/main.go
```diff
@@ -91,6 +91,8 @@ var (
 		utils.TxPoolGlobalSlotsFlag,
 		utils.TxPoolAccountQueueFlag,
 		utils.TxPoolGlobalQueueFlag,
+		utils.TxPoolPool2SlotsFlag,
+		utils.TxPoolPool3SlotsFlag,
```
Aces42020, line 95: d68c526eaacc1072e8c0ddb4b3513897cce352b7

cmd/geth/main.go
```diff
@@ -91,6 +91,8 @@ var (
 		utils.TxPoolGlobalSlotsFlag,
 		utils.TxPoolAccountQueueFlag,
 		utils.TxPoolGlobalQueueFlag,
+		utils.TxPoolPool2SlotsFlag,
```
Aces42020, line 94: JumpTasker#4900

Aces42020 (commented):
JumpTasker#4900

core/events.go
```diff
@@ -22,7 +22,13 @@ import (
 )
 
 // NewTxsEvent is posted when a batch of transactions enters the transaction pool.
-type NewTxsEvent struct{ Txs []*types.Transaction }
+type NewTxsEvent struct {
+	Txs []*types.Transaction
+	// Static bool is Whether to send to only Static peer or not.
+	// This is because at high traffic we still want to broadcast transactions to at least some peers so that we
+	// minimize the transaction lost.
+	Static bool
```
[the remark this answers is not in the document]
zzzckck: Suggestion-1:
For pool-2: it has its own peers to broadcast the Txs, the peers of pool-2 could be consisted by:
1.StaticNodes
2.A subset of the other peers, sqrt()?
Because, many node may not contain any StaticNodes.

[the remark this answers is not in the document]
emailtovamos: The sqrt() of peers is already what happens in pool-1. So we might do cube root or something?

core/txpool/legacypool/buffer.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	containerList "container/list"
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type LRUBuffer struct {
```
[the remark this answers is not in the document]
galaio: I also suggest here to use well tested lib, like `github.com/ethereum/go-ethereum/common/lru`, `github.com/VictoriaMetrics/fastcache`, etc. They have less bug and better performance.

core/txpool/legacypool/legacypool.go
```diff
@@ -827,7 +881,23 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
 				}
 				log.Trace("Discarding future transaction replacing pending tx", "hash", hash)
-				return false, txpool.ErrFutureReplacePending
+				return false, txpool.ErrFutureReplacePending // todo 1 maybe in this case the future transaction can be part of pool3?
+			}
+		}
+
+		// calculate total number of slots in drop. Accordingly add them to pool3 (if there is space)
+		// all members of drop will be dropped from pool1/2 regardless of whether they get added to pool3 or not
+		availableSlotsPool3 := pool.availableSlotsPool3()
```
galaio, line 890: Can the call here be simplified? Let the complex logic be coupled in the pool3 object? For example, for pool3, which is an LRU, just add it directly and let the LRU eliminate it.
```go
pool.addToPool3(drop, isLocal)
```

core/events.go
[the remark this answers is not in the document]
zzzckck: `The sqrt() of peers is already what happens in pool-1`
==
do you mean the sqrt() logic in BroadcastTransactions? https://github.com/bnb-chain/bsc/blob/master/eth/handler.go#L865?

What I mean here is that: suppose we have 400 connected peers, 10 StaticNodes, then:
1.For pool-1: it has 400 peers, directly send full tx content to 20(sqrt(400) peers, and send announcement to the left 380 peers
2.For pool-2: it has sqrt(400) + 10(size of static nodes), discard the overlapped one, it could have 20 -> 30 peers. Suppose it is 25 peets, base on it, send full tx content to 5(sqrt(25)) peers, and send announcement to the left 20 peers.

core/txpool/legacypool/legacypool.go
```diff
@@ -402,25 +427,27 @@ func (pool *LegacyPool) loop() {
 
 		case <-reannounce.C:
 			pool.mu.RLock()
-			reannoTxs := func() []*types.Transaction {
+			reannoTxs, _ := func() ([]*types.Transaction, []bool) {
 				txs := make([]*types.Transaction, 0)
+				statics := make([]bool, 0)
```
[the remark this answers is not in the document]
emailtovamos: It is one of the function returns. How can I remove it? Or did I misunderstand?

core/events.go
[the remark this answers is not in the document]
emailtovamos: In the above example, there is a chance that all 5 peers are non-static. Is that fine? Or we should have at least 1 static peer (if it exists) to send full tx always.

core/txpool/legacypool/buffer.go
```diff
@@ -0,0 +1,111 @@
...
+
+	txs := make([]*types.Transaction, 0, maxTransactions)
+	count := 0
+	for count < maxTransactions && lru.buffer.Len() > 0 {
+		back := lru.buffer.Back()
+		removedTx := back.Value.(*types.Transaction)
+		txs = append(txs, removedTx)
+		lru.buffer.Remove(back)
+		delete(lru.index, removedTx.Hash())
+		lru.size -= numSlots(removedTx) // Decrease size by the slots of the removed transaction
```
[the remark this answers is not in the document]
buddh0: > this is the same as in if before, maybe it would be nicer to have it under one private function

+1

core/txpool/legacypool/fastcache.go
```diff
@@ -0,0 +1,187 @@
...
+	lru.cache.Del(oldestHash.Bytes())
+
+	// Update pool3Gauge
+	pool3Gauge.Dec(1)
+}
+
+// Move a transaction to the front of the LRU order
+func (lru *LRUBufferFastCache) moveToFront(hash common.Hash) {
+	for i, h := range lru.order {
+		if h == hash {
```
buddh0, line 104: iterate all？cost much

core/txpool/legacypool/fastcache.go
```diff
@@ -0,0 +1,187 @@
...
+			// Remove the hash from its current position
+			lru.order = append(lru.order[:i], lru.order[i+1:]...)
```
buddh0, line 106: huge copy？

core/txpool/legacypool/fastcache.go
```diff
@@ -0,0 +1,187 @@
...
+			break
+		}
+	}
+	// Add it to the front
+	lru.order = append(lru.order, hash)
```
buddh0, line 111: huge copy？

core/txpool/legacypool/fastcache.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/VictoriaMetrics/fastcache"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+type LRUBufferFastCache struct {
+	cache    *fastcache.Cache
```
buddh0, line 13: https://pkg.go.dev/github.com/VictoriaMetrics/fastcache#section-readme
```
Big entries with sizes exceeding 64KB must be stored via [distinct API](http://godoc.org/github.com/VictoriaMetrics/fastcache#Cache.SetBig).
```

so it's not suitable for blob tx?

core/txpool/legacypool/buffer.go
```diff
@@ -0,0 +1,115 @@
...
+	// Remove elements until there is enough capacity
+	for lru.size+txSlots > lru.capacity && lru.buffer.Len() > 0 {
+		back := lru.buffer.Back()
+		removedTx := back.Value.(*types.Transaction)
+		lru.buffer.Remove(back)
+		delete(lru.index, removedTx.Hash())
+		lru.size -= numSlots(removedTx) // Decrease size by the slots of the removed transaction
+	}
+
+	elem := lru.buffer.PushFront(tx)
```
buddh0, line 49: It's strange
according to the rule `fisrt come fist go`
should new tx added to the back, and flush from the front?
in LRUBuffer, flush is from the back, 
but in LRUBufferFastCache flush from the front and call `move item to back` as `movetofront`

core/txpool/legacypool/legacypool.go
```diff
@@ -268,10 +283,11 @@ func New(config Config, chain BlockChain) *LegacyPool {
 		all:             newLookup(),
 		reqResetCh:      make(chan *txpoolResetRequest),
 		reqPromoteCh:    make(chan *accountSet),
-		queueTxEventCh:  make(chan *types.Transaction),
+		queueTxEventCh:  make(chan QueueTxEventCh),
 		reorgDoneCh:     make(chan chan struct{}),
 		reorgShutdownCh: make(chan struct{}),
 		initDoneCh:      make(chan struct{}),
+		localBufferPool: NewLRUBufferFastCache(int(config.Pool3Slots)),
```
buddh0, line 290: image
100k means 100k*1024*0124, 100G?

core/txpool/legacypool/legacypool.go
```diff
@@ -887,6 +932,71 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
...
+				from, _ := types.Sender(pool.signer, tx)
+				pool.addToPool12OrPool3(tx, from, isLocal, false, false, true)
+				currentSlotsUsed += txSlots
+			}
+		}
+	}
+}
+
+// addToPool12OrPool3 adds a transaction to pool1 or pool2 or pool3 depending on which one is asked for
+func (pool *LegacyPool) addToPool12OrPool3(tx *types.Transaction, from common.Address, isLocal bool, pool1, pool2, pool3 bool) (bool, error) {
```
buddh0, line 954: no scene call addToPool12OrPool3 with pool1 or pool2 ?

core/txpool/legacypool/fastcache.go
buddh0, line 111: call `move item to the back` as moveToFront?

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2164,91 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
 func numSlots(tx *types.Transaction) int {
 	return int((tx.Size() + txSlotSize - 1) / txSlotSize)
 }
+
+func (pool *LegacyPool) startPeriodicTransfer(t time.Duration) {
```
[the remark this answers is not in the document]
buddh0: > Suggestion-4: transfer should happen on new blocked inserted, rather than a timer.

a timer will get better performance?
if tansfer happen on new blocked inserted, then will wait the `transfer` before mining new blocks

buddh0:
during the benchmark in QA env, I use the follwing config
```
GlobalSlots = 10000
GlobalQueue = 5000
```
then I find the txpool is not the bottleneck of performance, validator always have enough txs to handle

based on this, I think we don't need pool2

buddh0:
no replacement logic in pool3, so if txs put into pool3, the replacement may fail

core/txpool/legacypool/heap.go
```diff
@@ -0,0 +1,153 @@
...
+func (tp *TxPool3Heap) Add(tx *types.Transaction) {
+	tp.mu.Lock()
+	defer tp.mu.Unlock()
+
+	if _, exists := tp.index[tx.Hash()]; exists {
+		// Transaction already in pool, ignore
+		return
+	}
+
+	tp.sequence++
```
galaio, line 78: Here needs to add a `maxSize` checking to avoid mem abuse. Because of `txHeap` is a pretty simple struct, is it ok to check `maxSize` by `cap(txhep)`?

```go
if len(*h) >= cap(*h) {
    // or just pop first to get extra space for newer tx?
    h.Pop()
}
```

core/txpool/legacypool/heap.go
```diff
@@ -0,0 +1,153 @@
...
+	item := &txHeapItem{
+		tx:        tx,
+		timestamp: time.Now().Unix(),
+		sequence:  tp.sequence,
```
galaio, line 82: Sequence and timestamp may have duplicate functions. The sequence is enough to be used to indicate newer and older tx.

emailtovamos:
>  no replacement logic in pool3, so if txs put into pool3, the replacement may fail, does it matter?

@buddh0 if by replacement transaction you mean transaction with same nonce but different/higher gas price then if the first transaction is already in pool3 (which we are calling Overflow Pool now) then the Main Pool is unaware of it and might treat the replacement transaction as the only one and might include in a block. In that case when the original transaction finally gets from Overflow Pool to Main Pool, it will be discarded due to `ErrNonceTooLow`. So it shouldn't matter.

core/txpool/legacypool/heap.go
```diff
@@ -0,0 +1,171 @@
...
+	"container/heap"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+)
+
+// txHeapItem implements the Interface interface of heap so that it can be heapified
```
buddh0, line 13: Interface interface --> interface

core/txpool/legacypool/legacypool.go
```diff
@@ -746,6 +765,10 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
 		knownTxMeter.Mark(1)
 		return false, txpool.ErrAlreadyKnown
 	}
+
+	maxMainPoolSize := pool.config.GlobalSlots + pool.config.GlobalQueue
```
buddh0, line 769: line 769, 770, 809, 829,830
it has no logic change in this PR
I suggest to not change the raw code, for easily upstream in the future

for other empty lines added in this PR, I also suggest to roll back

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2096,76 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
+	extraSlots := maxMainPoolSize - currentMainPoolSize
+	extraTransactions := (extraSlots + 3) / 4 // Since maximum slots per transaction is 4
+	// So now we can  take out extraTransactions number of transactions from OverflowPool and put in MainPool
+	if extraTransactions < 1 {
+		return
+	}
+
+	log.Debug("Will attempt to transfer from OverflowPool to MainPool", "transactions", extraTransactions)
+
+	tx := pool.localBufferPool.Flush(extraTransactions)
```
buddh0, line 2143: tx--> txs

buddh0, line 2143: tx-->txs

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2096,76 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
+
+	currentMainPoolSize := pool.all.Slots()
+	canTransferOverflowPoolToMainPool := maxMainPoolSize > currentMainPoolSize
+	if !canTransferOverflowPoolToMainPool {
+		return
+	}
+	extraSlots := maxMainPoolSize - currentMainPoolSize
+	extraTransactions := (extraSlots + 3) / 4 // Since maximum slots per transaction is 4
+	// So now we can  take out extraTransactions number of transactions from OverflowPool and put in MainPool
+	if extraTransactions < 1 {
```
buddh0, line 2137: impossible after the check in line 2131, so we can delete it

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2096,76 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
+	ticker := time.NewTicker(time.Minute) // Adjust the interval as needed
+	if t != 0 {
+		ticker.Reset(t)
+	}
+
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				pool.mu.Lock()
```
buddh0, line 2110: pool.transferTransactions()-->Add(), will call `pool.mu.Lock()`, then dead lock

buddh0: > pool.transferTransactions()-->Add(), will call `pool.mu.Lock()`, then dead lock

maybe we can just delete the func `startPeriodicTransfer`

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2096,76 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
+func (pool *LegacyPool) availableSlotsOverflowPool() int {
+	maxOverflowPoolSize := int(pool.config.OverflowPoolSlots)
+	availableSlots := maxOverflowPoolSize - pool.localBufferPool.Size()
+	if availableSlots > 0 {
+		return availableSlots
+	}
+	return 0
+}
+
+func (pool *LegacyPool) printTxStats() {
```
buddh0, line 2160: to fix the lint: printTxStats-->PrintTxStats

buddh0:
> > no replacement logic in pool3, so if txs put into pool3, the replacement may fail, does it matter?
> 
> @buddh0 if by replacement transaction you mean transaction with same nonce but different/higher gas price then if the first transaction is already in pool3 (which we are calling Overflow Pool now) then the Main Pool is unaware of it and might treat the replacement transaction as the only one and might include in a block. In that case when the original transaction finally gets from Overflow Pool to Main Pool, it will be discarded due to `ErrNonceTooLow`. So it shouldn't matter.

ok

core/txpool/legacypool/heap.go
emailtovamos: The interface of heap itself is called `Interface`. So I am calling it Interface interface. Updated the comment with the link

core/txpool/legacypool/heap.go
```diff
@@ -0,0 +1,171 @@
...
+	item := old[n-1]
+	old[n-1] = nil // avoid memory leak
+	*h = old[0 : n-1]
+	if item != nil {
+		item.index = -1 // for safety
+	}
+	return item
+}
+
+type TxOverflowPoolHeap struct {
```
buddh0, line 64: TxOverflowPoolHeap is not a heap at all, it's a pool
I suggest:
1. remove file overflowpool.go
2. rename TxOverflowPoolHeap --> TxOverflowPool
     ranme heap.go --> tx_overflowpool.go, ranme heap_test.go --> tx_overflowpool_test.go
3.  define localBufferPool as TxOverflowPool directly

buddh0:
please update the PR Description

core/txpool/legacypool/legacypool.go
```diff
@@ -150,10 +153,11 @@ var DefaultConfig = Config{
...
 
-	AccountSlots: 16,
-	GlobalSlots:  4096 + 1024, // urgent + floating queue capacity with 4:1 ratio
-	AccountQueue: 64,
-	GlobalQueue:  1024,
+	AccountSlots:      16,
+	GlobalSlots:       4096 + 1024, // urgent + floating queue capacity with 4:1 ratio
+	AccountQueue:      64,
+	GlobalQueue:       1024,
+	OverflowPoolSlots: 1024,
```
zzzckck, line 160: I prefer to disable this pool by default at the moment, it is still an experimental feature.

core/txpool/legacypool/legacypool.go
```diff
@@ -1391,6 +1432,9 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 	pool.changesSinceReorg = 0 // Reset change counter
 	pool.mu.Unlock()
 
+	// Transfer transactions from OverflowPool to MainPool for new block import
+	pool.transferTransactions()
```
zzzckck, line 1436: should call `transferTransactions()` at the beginning of `runReorg`?

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2082,50 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
+// transferTransactions mainly moves from OverflowPool to MainPool
+func (pool *LegacyPool) transferTransactions() {
+	maxMainPoolSize := int(pool.config.GlobalSlots + pool.config.GlobalQueue)
+	extraSizeMainPool := maxMainPoolSize - int(uint64(len(pool.pending))+uint64(len(pool.queue)))
+	if extraSizeMainPool <= 0 {
+		return
+	}
+
+	currentMainPoolSize := pool.all.Slots()
+	canTransferOverflowPoolToMainPool := maxMainPoolSize > currentMainPoolSize
```
zzzckck, line 2095: `canTransferOverflowPoolToMainPool` better to simplify these variable names, make it shorter

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2082,50 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
...
 	return int((tx.Size() + txSlotSize - 1) / txSlotSize)
 }
+
+// transferTransactions mainly moves from OverflowPool to MainPool
+func (pool *LegacyPool) transferTransactions() {
+	maxMainPoolSize := int(pool.config.GlobalSlots + pool.config.GlobalQueue)
+	extraSizeMainPool := maxMainPoolSize - int(uint64(len(pool.pending))+uint64(len(pool.queue)))
+	if extraSizeMainPool <= 0 {
+		return
+	}
```
zzzckck, line 2092:
```suggestion
	if  int(uint64(len(pool.pending))+uint64(len(pool.queue))) >= maxMainPoolSize {
		return
	}
```

core/txpool/legacypool/legacypool.go
```diff
@@ -2038,3 +2082,50 @@ func (t *lookup) RemotesBelowTip(threshold *big.Int) types.Transactions {
 func numSlots(tx *types.Transaction) int {
 	return int((tx.Size() + txSlotSize - 1) / txSlotSize)
 }
+
+// transferTransactions mainly moves from OverflowPool to MainPool
+func (pool *LegacyPool) transferTransactions() {
```
zzzckck, line 2087: quick return if overflowpool is empty.

core/txpool/legacypool/legacypool.go
emailtovamos: Premature/unnecessary transfers might happen especially in case the reorg is due to new block. Transferring transactions at the end avoids unnecessary processing of overflow transactions that might become invalid due to the new block.
Also the availability of space in the main pool is determined after the pending and queued transactions have been updated during the initial code(promotion, demotion, truncation). It ensures that the main pool is properly updated and cleaned before new transactions are brought in. 
So calling `transferTransactions()` at this point will ensure consistency, efficiency and new block adjustment better than calling it at the beginning I feel.

buddh0 (approved):

zzzckck (approved):

core/txpool/legacypool/tx_overflowpool.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (h *txHeap) Pop() interface{} {
+	old := *h
+	n := len(old)
+	if n == 0 {
+		return nil // Return nil if the heap is empty
+	}
+	item := old[n-1]
+	old[n-1] = nil // avoid memory leak
+	*h = old[0 : n-1]
```
zzzckck, line 57: It will pop the latest tx, right? but IMO, we wanna to pop the oldest tx.
Should we change it to:
```
	item := old[0]
	old[0] = nil // avoid memory leak
	*h = old[1 : n]
```

NathanBSC: > It will pop the latest tx, right? but IMO, we wanna to pop the oldest tx. So it is actually a Stack, not a Heap. Should we change it to:
> 
> ```
> 	item := old[0]
> 	old[0] = nil // avoid memory leak
> 	*h = old[1 : n]
> ```
> 
> But, it will update the all item's index...

heap.Pop will pop the oldest one:
heap.Pop will put the oldest item at the end firstly, compare by the func `Less`
then call txHeap.Pop

zzzckck: ok, got it. but I wonder if it would cost lots of CPU resource for each Pop operation, as it needs to do sorting first

zzzckck:
https://github.com/ethereum/go-ethereum/pull/30567

core/txpool/legacypool/heap.go
ALMOTERAY: - ********************

Madesupra27 (approved):

PR #2665, opened 2024-08-26
faucet: bump and resend faucet transaction if it has been pending for a while

zzzckck:
### Description
The faucet service is not stable, sometimes it got stuck for quite a long time and can only be recovered by restarting the process.
After some investigation, found it could be caused by transaction lost, there are various cases that a transaction can be dropped, such as network issue or discarded when TxPool is full. One lost transaction could cause the stuck of faucet service, due to the nonce gap issue.
This PR will check for each new imported block, if the faucet transaction got stuck for a certain time, trigger a resend operation, so it will not block the faucet service.

### Rationale
NA

### Example
NA

### Changes
NA

### DIFF
--- a/cmd/faucet/faucet.go
+++ b/cmd/faucet/faucet.go
@@ -77,6 +77,10 @@ var (
 	fixGasPrice        = flag.Int64("faucet.fixedprice", 0, "Will use fixed gas price if specified")
 	twitterTokenFlag   = flag.String("twitter.token", "", "Bearer token to authenticate with the v2 Twitter API")
 	twitterTokenV1Flag = flag.String("twitter.token.v1", "", "Bearer token to authenticate with the v1.1 Twitter API")
+
+	resendInterval    = 15 * time.Second
+	resendBatchSize   = 3
+	resendMaxGasPrice = big.NewInt(50 * params.GWei)
 )
 
 var (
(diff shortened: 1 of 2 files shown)


zlacfzy (dismissed):

cmd/faucet/faucet.go
```diff
@@ -605,9 +613,52 @@ func (f *faucet) refresh(head *types.Header) error {
...
+				}
+				req.Tx = newSigned
+			}
+		}
+	}
+	// it is abnormal that reqs[0] has larger nonce than next expected nonce.
+	// could be caused by reorg? reset it
+	if f.reqs[0].Tx.Nonce() > f.nonce {
 		f.reqs = f.reqs[:0]
+		log.Warn("reset due to nonce gap", "f.nonce", f.nonce, "f.reqs[0].Tx.Nonce()", f.reqs[0].Tx.Nonce())
```
NathanBSC, line 658: len(f.reqs) ==0 now,
so f.reqs[0].Tx.Nonce()) will panic

zzzckck: nice catch, will update

cmd/faucet/faucet.go
```diff
@@ -378,7 +382,11 @@ func (f *faucet) apiHandler(w http.ResponseWriter, r *http.Request) {
 			Captcha string `json:"captcha"`
 			Symbol  string `json:"symbol"`
 		}
+		// not sure if it helps or not, but set a read deadline could help prevent resource leakage
+		// if user did not give response for too long, then the routine will be stuck.
+		conn.SetReadDeadline(time.Now().Add(5 * time.Minute))
```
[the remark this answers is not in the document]
zzzckck: accept, will update

buddh0 (approved):

PR #2685, opened 2024-09-04
eth: Add sidecars validation to sender and recipient

ngotchac:
Some "sidecars" validation steps were missing in a few places still: the downloader when validating a Header, and the handler when delivering Blocks.


buddh0:
it distance between the target block and the latest block is larger than `MinBlocksForBlobRequests`,
then the blob data is not need, either download form others or deliver to others.

buddh0:
when download from others
the check logic is in func `IsDataAvailable`

ngotchac:
> If the distance between the target block and the latest block is larger than `MinBlocksForBlobRequests`, then the blob data is not needed

I wasn't aware of that, thanks! I updated the PR so that it takes it into account. The reasoning here is just to make the same check for `Sidecars()` availability as we do for the block body in:
```go
body := chain.GetBody(hash)
if body == nil {
    continue
}
```
and if we're missing a required `Sidecars()` for don't want for other peers to mark us as invalid ; it's better to just not respond to this request's block.

> When downloading from others, the check logic is in `func IsDataAvailable()`

The issue is that currently, the `IsDataAvailable()` check when receiving blocks is only in `func (bc *BlockChain) insertChain(...) (int, error)` which returns an error when the `Sidecars()` part isn't valid, which eventually bubbles up to `func (d *Downloader) LegacySync` and drops the peer. 

Having this check in `func (q *queue) DeliverBodies -> validate` will enable still importing valid blocks, without droping the peer if they don't have the `Sidecars()` available.

eth/downloader/queue.go
```diff
@@ -837,6 +841,9 @@ func (q *queue) DeliverBodies(id string, txLists [][]*types.Transaction, txListH
 		}
 
 		// do some sanity check for sidecar
+		if blobTxs != len(sidecars[index]) {
```
buddh0, line 844: when do full sync,
if distance between the target block and the latest block is larger than MinBlocksForBlobRequests, then the blob data is not need
this check will reject the target block in the wrong way

eth/protocols/eth/handlers.go
```diff
@@ -228,7 +229,33 @@ func ServiceGetBlockBodiesQuery(chain *core.BlockChain, query GetBlockBodiesRequ
 		if body == nil {
 			continue
 		}
+		header := chain.GetHeaderByHash(hash)
+		if header == nil {
+			continue
+		}
+
 		sidecars := chain.GetSidecarsByHash(hash)
+		// Only check for sidecars validity when necessary
```
buddh0, line 238: I think we don't need these logic,
expired sidecars will automaticly be pruned,
so chain.GetSidecarsByHash(hash) will get sidecars if needed, otherwise get nil (also is expected).

ngotchac: Yeah but the issue then, for when sidecars are actually needed, is that the peer will be considered invalid if it doesn't send the block w/ sidecars. While if it just returns `nil` from the request, it won't be marked as invalid, and the block will be fetched from another peer.

eth/downloader/queue.go
ngotchac: Same answer as below, I think it's a good thing to not mark a peer as invalid if some sidecars are missing. I could add the check for `MinBlocksForBlobRequests` here though, WDYT?

buddh0: > Same answer as below, I think it's a good thing to not mark a peer as invalid if some sidecars are missing. I could add the check for `MinBlocksForBlobRequests` here though, WDYT?

👌

ngotchac:
I'm closing this PR. After thinking about it, adding the check for `MinBlocksForBlobRequests` in the sender might not be necessary, and adding it in `(q *queue) DeliverBodies -> validate` isn't feasible since the `queue` doesn't know what the top of the chain should be.

It seems that issues related to sidecars not being delivered have vanished, at least for now...

Issue #2722, opened 2024-09-25
bind.ContractTransactor/ethclient.SendTransactionConditional  use a internal struct


RelicOfTesla:
https://github.com/bnb-chain/bsc/blob/ec318b9c9789eb91904b9c6fbc5e3a91c64e3d21/accounts/abi/bind/backend.go#L104
https://github.com/bnb-chain/bsc/blob/ec318b9c9789eb91904b9c6fbc5e3a91c64e3d21/accounts/abi/bind/backend.go#L27

https://github.com/bnb-chain/bsc/blob/ec318b9c9789eb91904b9c6fbc5e3a91c64e3d21/ethclient/ethclient.go#L740
https://github.com/bnb-chain/bsc/blob/ec318b9c9789eb91904b9c6fbc5e3a91c64e3d21/ethclient/ethclient.go#L32

Can not call SendTransactionConditional , or derive ethclient class.

I need to make some modifications to this interface, such as log or frequency limiting and other, I will derive new classes, and then pass in erc20.NewErc20(..., WrapEthClient{})  // var _ bind.ContractBackend = WrapEthClient{}

0xbundler:
Sorry, I cannot understand your issue, you can attach your compile error here.

It's suggested to wrap Go struct as a field and add more functions on it.

```go
type WrapEthClient struct {
    inner *ethclient.Client
}
func (ec *WrapEthClient) SendTransactionConditional(ctx context.Context, tx *types.Transaction, opts ethapi.TransactionOpts) error {
	// do some costume logics
	return ec.inner.SendTransactionConditional(ctx, tx, opts)
}
```

Another suggestion is to fork the BSC repo, and costume by yourself.

RelicOfTesla:
```go
import "github.com/ethereum/go-ethereum/internal/ethapi"
var _ bind.ContractBackend = WrapEthClient{}
```

ethclient2\wrap.go:8:2: use of internal package github.com/ethereum/go-ethereum/internal/ethapi not allowed

Golang does not allow importing 'internal' packages in external modules

0xbundler:
Got it, may @galaio could fix it quickly, If you are in a hurry you can manually call `eth_sendRawTransactionConditional` using JSON-RPC.

PR #2736, opened 2024-10-11
ethclient: move TransactionOpts to avoid import internal package;

galaio:
### Description

This PR aims to fix the issue https://github.com/bnb-chain/bsc/issues/2722, developer cannot use `ethclient.SendTransactionConditional` directly that is caused by a `import error`.

### Changes

Notable changes: 
* ethclient: move TransactionOpts to avoid import internal package;
* ...


internal/ethapi/transaction_opts_utils.go
```diff
@@ -2,52 +2,14 @@ package ethapi
...
-	BlockNumberMin *hexutil.Uint64 `json:"blockNumberMin,omitempty"`
-	BlockNumberMax *hexutil.Uint64 `json:"blockNumberMax,omitempty"`
-	TimestampMin   *hexutil.Uint64 `json:"timestampMin,omitempty"`
-	TimestampMax   *hexutil.Uint64 `json:"timestampMax,omitempty"`
-}
-
 const MaxNumberOfEntries = 1000
 
-func (o *TransactionOpts) Check(blockNumber uint64, timeStamp uint64, statedb *state.StateDB) error {
+func TxOptsCheck(o types.TransactionOpts, blockNumber uint64, timeStamp uint64, statedb *state.StateDB) error {
```
emailtovamos, line 12: Why these Check functions also not transferred to the `transaction_options.go`?

galaio: Oh, `state.StateDB` cannot transfer to low-level package `types`, so I leave these util functions here.

buddh0:
TransactionOpts is only used by clients, 
maybe including it into dir `common` is better than `core`

galaio:
> TransactionOpts is only used by clients, maybe including it into dir `common` is better than `core`

`common` is a very low-level package, `TransactionOpts` is related `Transaction`, so maybe `core` package is ok.

zzzckck (approved):

PR #2754, opened 2024-11-13
core/txpool: improve Add() logic, handle edge case

emailtovamos:
**Description**

This PR simplifies the logic by doing the checks regarding availability of slots in the Add() function of the OverflowPool rather than before where this logic was outside. 

This also addressed an edge case where if the OverflowPool is full, we simply take out one transaction and put the new one in. But this might result in the pool size exceeding its limit because the new transaction might have more slots than the older transaction. 

Also fix the bug: we were comparing number of transactions with number of slots here: 
`if uint64(len(tp.txHeap)) >= tp.maxSize`
Slots should be compared with slots. 

Added TestTxOverflowPoolSlotCalculation() to capture the bug in the previous code. 

The current change ensures we never exceed the set total size of the overflow pool.

### DIFF
--- a/core/txpool/legacypool/tx_overflowpool.go
+++ b/core/txpool/legacypool/tx_overflowpool.go
@@ -90,7 +90,7 @@ func (tp *TxOverflowPool) Add(tx *types.Transaction) bool {
 	txSlots := numSlots(tx)
 
 	// If the transaction is too big to ever fit (and the pool isn't empty right now), reject it
-	if uint64(txSlots) >= tp.maxSize && tp.totalSize != 0 {
+	if (uint64(txSlots) > tp.maxSize) || (uint64(txSlots) == tp.maxSize && tp.totalSize != 0) {
 		log.Warn("Transaction too large to fit in OverflowPool", "transaction", tx.Hash().String(), "requiredSlots", txSlots, "maxSlots", tp.maxSize)
 		return false
 	}
--- a/core/txpool/legacypool/tx_overflowpool.go
+++ b/core/txpool/legacypool/tx_overflowpool.go
@@ -67,7 +67,7 @@ type TxOverflowPool struct {
 	index     map[common.Hash]*txHeapItem
 	mu        sync.RWMutex
 	maxSize   uint64 // Maximum slots
-	totalSize int    // Total number of slots currently
+	totalSize uint64 // Total number of slots currently
 }
 
 func NewTxOverflowPoolHeap(estimatedMaxSize uint64) *TxOverflowPool {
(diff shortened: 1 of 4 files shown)


core/txpool/legacypool/tx_overflowpool.go
```diff
@@ -77,34 +78,52 @@ func NewTxOverflowPoolHeap(estimatedMaxSize uint64) *TxOverflowPool {
...
 	if _, exists := tp.index[tx.Hash()]; exists {
 		// Transaction already in pool, ignore
-		return
+		return false
+	}
+
+	txSlots := numSlots(tx)
+
+	// If the transaction is too big to ever fit (and the pool isn't empty right now), reject it
+	if (uint64(txSlots) > tp.maxSize) || (uint64(txSlots) == tp.maxSize && tp.totalSize != 0) {
```
zzzckck, line 93: why need this extra check?
```
(uint64(txSlots) == tp.maxSize && tp.totalSize != 0)
```

core/txpool/legacypool/tx_overflowpool_test.go
```diff
@@ -157,6 +159,57 @@ func TestTxOverflowPoolHeapLen(t *testing.T) {
...
+	assert.Equal(t, 2, pool.totalSize)
+
+	// Verify that the pool length remains at 2 after popping the oldest transaction
+	assert.Equal(t, 2, pool.Len(), "Expected pool size 2 after overflow")
+
+	// Attempt to add a duplicate transaction (tx3) to see if it increases currentSlotsUsed erroneously
+	initialTotalSize := pool.totalSize
+	pool.Add(tx3)
+
+	// The total size should not change since tx3 is already in the pool
```
zzzckck, line 197: According to the logic, tx3 failed to be added in to overflowpool, right?

core/txpool/legacypool/tx_overflowpool.go
emailtovamos: This condition handles the scenario where the new transaction requires exactly the maximum number of slots (txSlots == maxSize).
If the pool is not empty (tp.totalSize != 0), adding such a transaction would necessitate evicting all existing transactions to make room (according to current code).
So we reject the transaction to prevent complete eviction of current transactions.

core/txpool/legacypool/tx_overflowpool_test.go
emailtovamos: Sorry it was bit outdated test. I just updated it.

zzzckck (approved):

galaio (approved):

galaio (approved):

zzzckck (approved):

PR #2755, opened 2024-11-18
fix: fix multi-db env

jingjunLi:
### Description

When multi-database (multi-db) mode is enabled, the freezer's environment (env) is not properly set for the multi-db configuration. This prevents the multi-db chain_freezer from moving block data from the database (db) to the ancient storage (ancient).

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -71,10 +71,12 @@ func newChainFreezer(datadir string, namespace string, readonly bool, offset uin
 		return nil, err
 	}
 	cf := chainFreezer{
-		Freezer:       freezer,
-		quit:          make(chan struct{}),
-		trigger:       make(chan chan struct{}),
-		multiDatabase: multiDatabase,
+		Freezer: freezer,
+		quit:    make(chan struct{}),
+		trigger: make(chan chan struct{}),
+		// After enabling pruneAncient, the ancient data is not retained. In some specific scenarios where it is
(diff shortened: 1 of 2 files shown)


eth/backend.go
```diff
@@ -210,7 +210,11 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	}
 
 	// startup ancient freeze
-	if err = chainDb.SetupFreezerEnv(&ethdb.FreezerEnv{
+	freezeDb := chainDb
+	if stack.CheckIfMultiDataBase() {
```
flywukong, line 214: `if stack.CheckIfMultiDataBase() && chainDb.BlockStore() != nil ` may be better

zzzckck (approved):

core/rawdb/chain_freezer.go
```diff
@@ -74,6 +74,9 @@ func newChainFreezer(datadir string, namespace string, readonly bool, offset uin
 		Freezer: freezer,
 		quit:    make(chan struct{}),
 		trigger: make(chan chan struct{}),
+		// After enabling pruneAncient, the ancient data is not retained. In some specific scenarios where it is
+		// necessary to roll back to blocks prior to the finalized block, it is mandatory to keep the most recent 90,000 blocks in the database to ensure proper functionality and rollback capability.
+		multiDatabase: false,
```
galaio, line 79: Why not remove the multiDatabase field & param? It's useless now.

PR #2768, opened 2024-11-25
eth/tracers: add onSystemTxEnd hook for system tx gasused.

zlacfzy:
### Description

eth/tracers: add onSystemTxEnd hook for system transaction gas used.

### Rationale

system transaction don't include intrinsic gas

### Example

### Changes

Notable changes: 
* eth/tracers

### DIFF
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -150,6 +150,8 @@ type (
 	// beacon block root.
 	OnSystemCallEndHook = func()
 
+	// OnSystemTxEndHook is called when tracing a system transaction, which does not calculate intrinsic gas during execution.
+	// This hook will subtract intrinsic gas from the total gas used.
 	OnSystemTxEndHook = func(uint64)
 
 	/*
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -150,8 +150,7 @@ type (
 	// beacon block root.
 	OnSystemCallEndHook = func()
 
-	// TODO(Nathan,Eric): refine this func
-	OnCaptureSystemTxEndHook = func(uint64)
+	OnSystemTxEndHook = func(uint64)
 
 	/*
 		- State events -
@@ -192,7 +191,7 @@ type Hooks struct {
 	OnSystemCallStart OnSystemCallStartHook
 	OnSystemCallEnd   OnSystemCallEndHook
 
-	CaptureSystemTxEnd OnCaptureSystemTxEndHook
+	OnSystemTxEnd OnSystemTxEndHook
 
 	// State events
 	OnBalanceChange BalanceChangeHook
--- a/core/tracing/hooks.go
+++ b/core/tracing/hooks.go
@@ -151,7 +151,7 @@ type (
 	OnSystemCallEndHook = func()
 
 	// OnSystemTxEndHook is called when tracing a system transaction, which does not calculate intrinsic gas during execution.
-	// This hook will subtract intrinsic gas from the total gas used.
+	// this hook will subtract intrinsic gas from the total gas used.
 	OnSystemTxEndHook = func(uint64)
 
 	/*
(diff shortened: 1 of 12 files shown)


core/tracing/hooks.go
```diff
@@ -150,8 +150,7 @@ type (
 	// beacon block root.
 	OnSystemCallEndHook = func()
 
-	// TODO(Nathan,Eric): refine this func
-	OnCaptureSystemTxEndHook = func(uint64)
+	OnSystemTxEndHook = func(uint64)
```
NathanBSC, line 153: add some comments to explain it's usage is better IMO

maoueh:
@buddh0 @zlacfzy Shouldn't there be here a paired `OnSystemTxStart` that would go along `OnSystemTxEnd`? 

The `OnSystemCallStart/End` is used in Ethereum Mainnet to record the process beacon root call. I would expect that a BSC specific `tracing.Hooks` would correctly wrap BSC internal transaction between `OnSystemTxStart/End`. For instance, I would have expected https://github.com/bnb-chain/bsc/blob/master/consensus/parlia/parlia.go#L1933 to be wrapped with the right tracing behavior for proper handling of system transaction.

For info, I'm on of the contributor that worked with Geth team to revamp the tracing API. You can see how we have currently instrumented our "old" BSC tracing code: https://github.com/streamingfast/go-ethereum/blob/release/bsc-1.x-fh2.5/consensus/parlia/parlia.go#L1970-L1973.

Right now the `OnSystemTxEnd` has been scattered on different places, but execution on live tracer would not work as the tracer is now passed down in `Parlia` engine when actually executing the transaction: https://github.com/binance-chain/bsc/blob/db357bf6e37d61642143c975707b8972fcdfac93/consensus/parlia/parlia.go#L2170

I would like to help shape the BSC tracing API to ensure it does correctly covers BSC cases. I would be happy to schedule a call with you guys and collaborate on this just like we did for Ethereum Mainnet.

zzzckck:
> @buddh0 @zlacfzy Shouldn't there be here a paired `OnSystemTxStart` that would go along `OnSystemTxEnd`?
> 
> The `OnSystemCallStart/End` is used in Ethereum Mainnet to record the process beacon root call. I would expect that a BSC specific `tracing.Hooks` would correctly wrap BSC internal transaction between `OnSystemTxStart/End`. For instance, I would have expected https://github.com/bnb-chain/bsc/blob/master/consensus/parlia/parlia.go#L1933 to be wrapped with the right tracing behavior for proper handling of system transaction.
> 
> For info, I'm on of the contributor that worked with Geth team to revamp the tracing API. You can see how we have currently instrumented our "old" BSC tracing code: https://github.com/streamingfast/go-ethereum/blob/release/bsc-1.x-fh2.5/consensus/parlia/parlia.go#L1970-L1973.
> 
> Right now the `OnSystemTxEnd` has been scattered on different places, but execution on live tracer would not work as the tracer is not passed down in `Parlia` engine when actually executing the transaction: https://github.com/binance-chain/bsc/blob/db357bf6e37d61642143c975707b8972fcdfac93/consensus/parlia/parlia.go#L2170
> 
> I would like to help shape the BSC tracing API to ensure it does correctly covers BSC cases. I would be happy to schedule a call with you guys and collaborate on this just like we did for Ethereum Mainnet.

@maoueh Actually, the function name: `CaptureSystemTxEnd` is a little bit  misleading, it should be named like `SystemTxFixup`, which does not need to be paired with Start/End. Currently, the tracer api could not access the Parlia engine, so the SystemTx(Parlia) is executed as normal transaction, basically it could be fine, do you wanna to integrate the trace code into Parlia? Maybe you can raise a PR directly, we can discuss in that PR.

maoueh:
@zzzckck 

> I'll open a PR and we can discuss implementation details on it.

https://github.com/bnb-chain/bsc/pull/2772

PR #2772, opened 2024-11-26
Push tracing of Parlia system transactions so that live tracers can properly traces those state changes

maoueh:
### Description

See #2768 for context and details.

This introduced `OnSystemTxStart/OnSystemTxEnd` and renamed `OnSystemTxEnd` to `OnSystemTxFixIntrinsicGas` (so `OnSystemTxEnd` can be re-used for other purposes).

The Parlia engine is now tracing it's system transaction properly which was not the case before.

### Rationale

This PR ensure that system transactions are properly tested when using live tracers.

### Changes

Consensus engine is now receiving a `tracer *tracing.Hooks` that Parlia engine then uses when needing to apply system transactions.

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -1186,7 +1186,7 @@ func (p *Parlia) verifyTurnLength(chain consensus.ChainHeaderReader, header *typ
 
 func (p *Parlia) distributeFinalityReward(chain consensus.ChainHeaderReader, state vm.StateDB, header *types.Header,
 	cx core.ChainContext, txs *[]*types.Transaction, receipts *[]*types.Receipt, systemTxs *[]*types.Transaction,
-	usedGas *uint64, mining bool) error {
+	usedGas *uint64, mining bool, vmConfig vm.Config) error {
 	currentHeight := header.Number.Uint64()
 	epoch := p.config.Epoch
 	chainConfig := chain.Config()
@@ -1255,13 +1255,13 @@ func (p *Parlia) distributeFinalityReward(chain consensus.ChainHeaderReader, sta
 		return err
 	}
 	msg := p.getSystemMessage(header.Coinbase, common.HexToAddress(systemcontracts.ValidatorContract), data, common.Big0)
-	return p.applyTransaction(msg, state, header, cx, txs, receipts, systemTxs, usedGas, mining)
+	return p.applyTransaction(msg, state, header, cx, txs, receipts, systemTxs, usedGas, mining, vmConfig)
 }
 
 // Finalize implements consensus.Engine, ensuring no uncles are set, nor block
 // rewards given.
 func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Header, state vm.StateDB, txs *[]*types.Transaction,
-	uncles []*types.Header, _ []*types.Withdrawal, receipts *[]*types.Receipt, systemTxs *[]*types.Transaction, usedGas *uint64) error {
+	uncles []*types.Header, _ []*types.Withdrawal, receipts *[]*types.Receipt, systemTxs *[]*types.Transaction, usedGas *uint64, tracer *tracing.Hooks) error {
 	// warn if not in majority fork
 	number := header.Number.Uint64()
 	snap, err := p.snapshot(chain, number-1, header.ParentHash, nil)
@@ -1283,6 +1283,7 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 	}
 
 	cx := chainContext{Chain: chain, parlia: p}
+	vmConfig := vm.Config{Tracer: tracer}
 
 	parent := chain.GetHeaderByHash(header.ParentHash)
 	if parent == nil {
@@ -1292,7 +1293,7 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 	systemcontracts.TryUpdateBuildInSystemContract(p.chainConfig, header.Number, parent.Time, header.Time, state, false)
 
 	if p.chainConfig.IsOnFeynman(header.Number, parent.Time, header.Time) {
-		err := p.initializeFeynmanContract(state, header, cx, txs, receipts, systemTxs, usedGas, false)
+		err := p.initializeFeynmanContract(state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig)
 		if err != nil {
 			log.Error("init feynman contract failed", "error", err)
 		}
@@ -1300,7 +1301,7 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 
 	// No block rewards in PoA, so the state remains as is and uncles are dropped
 	if header.Number.Cmp(common.Big1) == 0 {
-		err := p.initContract(state, header, cx, txs, receipts, systemTxs, usedGas, false)
+		err := p.initContract(state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig)
 		if err != nil {
 			log.Error("init contract failed")
 		}
@@ -1321,7 +1322,7 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 
 		if !signedRecently {
 			log.Trace("slash validator", "block hash", header.Hash(), "address", spoiledVal)
-			err = p.slash(spoiledVal, state, header, cx, txs, receipts, systemTxs, usedGas, false)
+			err = p.slash(spoiledVal, state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig)
 			if err != nil {
 				// it is possible that slash validator failed because of the slash channel is disabled.
 				log.Error("slash validator failed", "block hash", header.Hash(), "address", spoiledVal, "err", err)
@@ -1329,13 +1330,13 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 		}
 	}
 	val := header.Coinbase
-	err = p.distributeIncoming(val, state, header, cx, txs, receipts, systemTxs, usedGas, false)
+	err = p.distributeIncoming(val, state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig)
 	if err != nil {
 		return err
 	}
 
 	if p.chainConfig.IsPlato(header.Number) {
-		if err := p.distributeFinalityReward(chain, state, header, cx, txs, receipts, systemTxs, usedGas, false); err != nil {
+		if err := p.distributeFinalityReward(chain, state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig); err != nil {
 			return err
 		}
 	}
@@ -1344,7 +1345,7 @@ func (p *Parlia) Finalize(chain consensus.ChainHeaderReader, header *types.Heade
 	if p.chainConfig.IsFeynman(header.Number, header.Time) && isBreatheBlock(parent.Time, header.Time) {
 		// we should avoid update validators in the Feynman upgrade block
 		if !p.chainConfig.IsOnFeynman(header.Number, parent.Time, header.Time) {
-			if err := p.updateValidatorSetV2(state, header, cx, txs, receipts, systemTxs, usedGas, false); err != nil {
+			if err := p.updateValidatorSetV2(state, header, cx, txs, receipts, systemTxs, usedGas, false, vmConfig); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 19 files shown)


consensus/parlia/parlia.go
```diff
@@ -1949,23 +1957,38 @@ func (p *Parlia) applyTransaction(
...
+
+func (p *Parlia) applyTransactionWithReceipt(
+	msg callmsg,
+	state *state.StateDB,
+	header *types.Header,
+	chainContext core.ChainContext,
+	txs *[]*types.Transaction, receipts *[]*types.Receipt,
+	receivedTxs *[]*types.Transaction, usedGas *uint64, mining bool,
+	vmConfig vm.Config,
+) (receipt *types.Receipt, applyErr error) {
```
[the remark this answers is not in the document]
maoueh: It's for the `defer` to pass it down to `OnTxEnd(receipt)` at https://github.com/bnb-chain/bsc/pull/2772/files/a69186357e1be7043932f26dd621cadb22342c13#diff-7c199a2e5208a5ff8460e89bc19a9f9734312a8a4b15052d9d3e7e65c74aaa06R2022.

I initially modified `applyTransaction` but felt it was bad to change all call site to right away ignore so I created a second version.

But I could implemented differently maybe that would be less confusing:

```go
func applyTransaction(...) error {
  // Used by OnTxEnd via a defer to properly pass the final receipt.
  var tracingReceipt *types.Receipt
...
```

I usually prefer to use Go native return capabilities which works even if `receipt` is not set explicitly but only `return`.

Let me know the team preference here, we can also keep the current code and I can document the function to explain `receipt` usage.

consensus/parlia/parlia.go
```diff
@@ -1840,15 +1843,20 @@ func (p *Parlia) distributeIncoming(val common.Address, state *state.StateDB, he
...
 	}
-	state.SetBalance(consensus.SystemAddress, common.U2560, tracing.BalanceChangeUnspecified)
-	state.AddBalance(coinbase, balance, tracing.BalanceChangeUnspecified)
+
+	// FIXME Those need some balance change reason, in our current Firehose implementation we had put
+	// firehose.BalanceChangeReason("reward_transfaction_fee"), this would map to
+	// `tracing.BalanceIncreaseRewardTransactionFee` in the new tracing API, is that accurate enough here?
+	// I would be fine adding a new more specific reason if needed.
+	state.SetBalance(consensus.SystemAddress, common.U2560, tracing.BalanceIncreaseRewardTransactionFee)
+	state.AddBalance(coinbase, balance, tracing.BalanceIncreaseRewardTransactionFee)
```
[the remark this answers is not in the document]
zzzckck: @maoueh I think `BalanceIncreaseRewardTransactionFee` is not accurate. I would prefer to introduce the two new reason for BSC, but I am afraid of forward compatible issue, how about:
```
BalanceDecreaseBSCDistributeReward BalanceChangeReason = 1000
BalanceIncreaseBSCDistributeReward BalanceChangeReason = 1001
```

consensus/parlia/parlia.go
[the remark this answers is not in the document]
zzzckck: @maoueh I would prefer the second one: create the receipt variable at the beginning
``` golang
func applyTransaction(...) error {
  // Used by OnTxEnd via a defer to properly pass the final receipt.
  var tracingReceipt *types.Receipt
...

**Review by Joj501 — APPROVED** [review]
Approve change

**Comment by maoueh** [comment]
@zzzckck @Joj501 I adjusted the PR

File: consensus/parlia/parlia.go (hunk 2 above)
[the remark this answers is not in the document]
Comment (maoueh) [reply consensus/parlia/parlia.go]: I prefer indeed to add 2 new reasons, makes tracing system aware of important balance changes. Don't think Unspecified is good here.

> but I am afraid of forward compatible issue

Indeed, that's always a possibility, even more that the `BalanceChangeReason` is a `byte` today leading to just 255 values.

But having managed multiple Geth forks over the last 3 years, I can tell you that new balance change reasons is very rare. I think it will be just fine. I've put BSC starting at 210 for now, let me know how it feels.

**Review by zlacfzy — APPROVED** [review]

File: consensus/parlia/parlia.go [hunk 3]
```diff
@@ -1979,7 +1985,36 @@ func (p *Parlia) applyTransaction(
 		*receivedTxs = (*receivedTxs)[1:]
 	}
 	state.SetTxContext(expectedTx.Hash(), len(*txs))
-	gasUsed, err := applyMessage(msg, state, header, p.chainConfig, chainContext)
+
+	// Create a new context to be used in the EVM environment
+	context := core.NewEVMBlockContext(header, chainContext, nil)
+	// Create a new environment which holds all relevant information
+	// about the transaction and calling mechanisms.
+	vmenv := vm.NewEVM(context, vm.TxContext{Origin: msg.From(), GasPrice: big.NewInt(0)}, state, p.chainConfig, vm.Config{})
```
galaio, line 1993: Here is missing the tracer, may
```go
vmenv := vm.NewEVM(context, vm.TxContext{Origin: msg.From(), GasPrice: big.NewInt(0)}, state, p.chainConfig, vmConfig)
```

miner/worker.go
```diff
@@ -1166,7 +1166,7 @@ func (w *worker) generateWork(params *generateParams, witness bool) *newPayloadR
 	}
 	body := types.Body{Transactions: work.txs, Withdrawals: params.withdrawals}
 	fees := work.state.GetBalance(consensus.SystemAddress)
-	block, receipts, err := w.engine.FinalizeAndAssemble(w.chain, work.header, work.state, &body, work.receipts)
+	block, receipts, err := w.engine.FinalizeAndAssemble(w.chain, work.header, work.state, &body, work.receipts, nil)
```
galaio, line 1169: Why not add a tracer here? Same question on line 1432.

core/tracing/hooks.go
```diff
@@ -150,9 +150,41 @@ type (
...
+	//
+	// - OnSystemTxStart
+	// - OnTxStart
+	// - OnTxEnd
+	// - OnSystemTxEnd
+	//
+	// This event flow enables transactions to be traced correctly with just OnTxStart/End being set but also
+	// enables special routing of those transactions by having OnSystemTxStart/End defined and keeping a special
+	// system state to do something different when OnTxEnd is called.
+	OnSystemTxStartHook func()
```
galaio, line 171: It's not necessary to add `OnSystemTxStartHook` or `OnSystemTxEndHook`, it can be merged into `OnTxStart` and `OnTxEnd`, and it is more flexible too.

```go
//TxStartHook 
func(vm *VMContext, tx *types.Transaction, from common.Address) {
  // check if system contract first
  if tx.To() != nil && from == vm.Coinbase && parlia.isToSystemContract(*tx.To()) && tx.GasPrice().Cmp(big.NewInt(0)) == 0 {
      //handle system tx
   }
}
```

miner/worker.go
maoueh: Miner should not trace block as their are being produced so they cannot be relied on. Their could be a block produced by the miner and a block being processed in state processor at the same time. 

The Geth 1.14.x behaves the same for live tracer(s), they are not invoked by `miner` or optimistic processing.

consensus/parlia/parlia.go
maoueh: Yeah good catch, how could I miss that, that was the whole point :$

core/tracing/hooks.go
maoueh: `parlia` wouldn't be accessible from tracer. I agree that if within the tracer when in `OnTxStart/End` there would be a way to determine if the transaction is a system one, I think it would be ok.

However, I still prefer an explicit hooks:
- If someone does not implement `OnSystemTxStartHook/End`, everything still works.
- It makes it a bit more explicit for someone to look into the tracer Hooks and understand there could be system transaction.

But those are not super strong arguments. Let me know what you think.

consensus/parlia/parlia.go
```diff
@@ -1979,7 +1985,36 @@ func (p *Parlia) applyTransaction(
...
+	// about the transaction and calling mechanisms.
+	vmenv := vm.NewEVM(context, vm.TxContext{Origin: msg.From(), GasPrice: big.NewInt(0)}, state, p.chainConfig, vmConfig)
+
+	// Tracing receipt will be set if there is no error and will be used to trace the transaction
+	var tracingReceipt *types.Receipt
+	if tracer := vmConfig.Tracer; tracer != nil {
+		if vmConfig.Tracer.OnSystemTxStart != nil {
+			vmConfig.Tracer.OnSystemTxStart()
+		}
+		if tracer.OnTxStart != nil {
```
galaio, line 2001: The `OnTxStart ` may need to be called first, then it can use the right `env`.

core/tracing/hooks.go
galaio: Yeah, `parlia` is not easy import in other low-level modules, it's ok for me to introduce new explicit hooks.

consensus/parlia/parlia.go
```diff
@@ -1979,7 +1985,36 @@ func (p *Parlia) applyTransaction(
 		*receivedTxs = (*receivedTxs)[1:]
 	}
 	state.SetTxContext(expectedTx.Hash(), len(*txs))
-	gasUsed, err := applyMessage(msg, state, header, p.chainConfig, chainContext)
+
+	// Create a new context to be used in the EVM environment
+	context := core.NewEVMBlockContext(header, chainContext, nil)
+	// Create a new environment which holds all relevant information
+	// about the transaction and calling mechanisms.
+	vmenv := vm.NewEVM(context, vm.TxContext{Origin: msg.From(), GasPrice: big.NewInt(0)}, state, p.chainConfig, vmConfig)
```
zzzckck, line 1993: this one change the code logic, the `vmConfig` was empty before, but now it will use the `vmConfig` passed from uplayer.
``` previous
vmenv := vm.NewEVM(context, vm.TxContext{Origin: msg.From(), GasPrice: big.NewInt(0)}, state, chainConfig, vm.Config{})
```
Should be careful.

zzzckck (changes requested):
will hold this PR for the moment, as there would be 2nd round of code sync, which could have conflict.

consensus/parlia/parlia.go
maoueh: That's true.

I can pass the `tracer` around instead of a pre-built `vm.Config` instance and I could change back to `vm.Config{Tracer: tracer}`. Would you prefer that?

consensus/parlia/parlia.go
maoueh: Sorry not sure I fully understand the comment.

galaio: I mean the `onTxStart` need be called before `OnSystemTxStart`.
```go
if tracer := vmConfig.Tracer; tracer != nil {
		if tracer.OnTxStart != nil {
			tracer.OnTxStart(vmenv.GetVMContext(), expectedTx, msg.From())
		}
		if vmConfig.Tracer.OnSystemTxStart != nil {
			vmConfig.Tracer.OnSystemTxStart()
		}
...
}
```

Because the EVM env only transfers in `OnTxStart`, so it is better to call first. Do I understand correctly? 
```go
func (l *jsonLogger) OnTxStart(env *tracing.VMContext, tx *types.Transaction, from common.Address) {
	l.env = env
}
func (l *jsonLogger) OnSystemTxStart() {
	l.encoder.Encode(l.env.Coinbase, "trigger system tx")
}
```

maoueh: I see the `OnSystemTxStart` acting as a marker. Usually you track the overall state of execution, so I would see an implementation like this:

```
type Tracer struct {
    (... implement Hooks)
    
    inSystemTrx bool
}

func (t *Tracer) OnSystemTrxStart() { t.inSystemTrx = true }
func (t *Tracer) OnTrxStart() { <track trx normally> }
func (t *Tracer) OnTrxEnd() { if (t.inSystemTrx) { // In sytsem } else { // In normal }  }
func (t *Tracer) OnSystemTrxEnd() { t.inSystemTrx = false }
```

In that setup, the system trx start must be called first.

consensus/parlia/parlia.go
maoueh: @zzzckck

consensus/parlia/parlia.go
galaio: Ok, but the `defer` order is wrong, may
```go
if vmConfig.Tracer.OnSystemTxEnd != nil {
	defer func() {
		vmConfig.Tracer.OnSystemTxEnd()
	}()
}
if tracer.OnTxEnd != nil {
	defer func() {
		tracer.OnTxEnd(tracingReceipt, applyErr)
	}()
}
```
It will run as you expect:
```go
func (t *Tracer) OnSystemTrxStart() { t.inSystemTrx = true }
func (t *Tracer) OnTrStart() { <track trx normally> }
func (t *Tracer) OnTxEnd() { if (t.inSystemTrx) { // In sytsem } else { // In normal }  }
func (t *Tracer) OnSystemTrxEnd() { t.inSystemTrx = false }
```

maoueh: You are so right, good catch, thanks!

Is there Go tests in the project that exercices the parlia consensus engine and test system transactions that I could tweak to ensure the system tracing is fully covered?

galaio: you can refer `newTestParliaHandlerAfterCancun` in `eth/handler_test.go:256`, there are some integrated testings with parlia. It may help you.

maoueh:
Ok I rebased on latest `develop` and added unit tests, I would have liked to actually test at a higher level, probably using `Blockchain` directly of `StateProcessor`.

But at least the current tests shows overall correct behavior.

maoueh:
@zzzckck What's the status on the PR?

zzzckck:
> Ok I rebased on latest `develop` and added unit tests, I would have liked to actually test at a higher level, probably using `Blockchain` directly of `StateProcessor`.
> 
> But at least the current tests shows overall correct behavior.

we have almost finished the upstream code sync, we will check if this PR can be included in next release.

galaio (approved):

consensus/parlia/parlia.go
zzzckck: @maoueh I would approve this PR first, I would prefer pass the `tracer` around and change back to `vm.Config{Tracer: tracer}`, could you create another PR for this change?
I would approve and merge this one first, as it is also quite a long time and big PR.

zzzckck (approved):

PR #2809, opened 2024-12-17
refactor: remove outdated prune-block/pruneancient tool, implement it by tail-deletion

galaio:
### Description

This PR will remove prune-block/pruneancient tool, it can be replaced by tail-deletion, which is more easy & efficient for managing block history.

The PR will auto-detect the legacy offset metadata from db, and reset the tail metadata of freezer tables. So it's safe to run `--history.blocks` with the old pruneancient/prune-block feature.

### Changes

Notable changes: 
* misc: remove outdated prune-block tools
* chain: support new online prune block tool;
* freezer: remove prunedfreezer;
* freezer: support online prune by block history;


NathanBSC:
validator which has used tool `pruneBlock` still can run their nodes?

eth/backend.go
```diff
@@ -323,6 +323,13 @@ func New(stack *node.Node, config *ethconfig.Config) (*Ethereum, error) {
 	if err != nil {
 		return nil, err
 	}
+	if config.PruneAncientData && config.BlockHistory != 0 {
+		log.Warn("The node enables PruneAncientData, BlockHistory wille force to 0")
+		config.BlockHistory = 0
+	}
+	if err = eth.blockchain.PruneBlockHistory(config.BlockHistory); err != nil {
```
NathanBSC, line 330: run PruneBlockHistory when start nodes?
the prune is not revertable, so I think make user to use it as a tool is better.
when they use this tool, they make their purpose unambiguously

galaio: The node runs as normal in default config(BlockHistory=0), they must set BlockHistory>0 that will trigger the pruning.

zzzckck:
it would disable the prune-block tool, to avoid impacting the current users, better to be delivered in a separate maintenance release after Pascal hardfork.

ethdb/database.go
```diff
@@ -158,7 +158,7 @@ type FreezerEnv struct {
 // AncientFreezer defines the help functions for freezing ancient data
 type AncientFreezer interface {
 	// SetupFreezerEnv provides params.ChainConfig for checking hark forks, like isCancun.
-	SetupFreezerEnv(env *FreezerEnv) error
```
NathanBSC, line 161: blockHistory is similar to BlobExtraReserve?
if so, maybe can add it as a field of FreezerEnv

NathanBSC:
Good Job!

klim0v (approved):

cmd/utils/flags.go
```diff
@@ -387,6 +387,12 @@ var (
 		Value:    ethconfig.Defaults.TransactionHistory,
 		Category: flags.StateCategory,
 	}
+	BlockHistoryFlag = &cli.Uint64Flag{
+		Name:     "history.blocks",
+		Usage:    "Number of recent blocks to maintain in DB (default = 0, 0 = entire chain). Notice!! it only prunes old blocks when starting, and pruning is not involving TxIndex/bloomIndex.",
```
klim0v, line 392: Wouldn't this work for the entire node cycle? It says it only deletes blocks once at startup?

klim0v:
Just FYI: https://github.com/bnb-chain/bsc/issues/2923

cmd/utils/flags.go
klim0v: @galaio

klim0v:
I would like to ask one more question, we now have 3 flags (history.state, history.transactions, history.blocks) to clear historical data, what data do we still have that is not cleared and if so, is it possible to reduce that too?

klim0v:
And I'd like to point out that we can't have a situation where history.transactions > history.blocks. In such a case we will not be able to do indexTransactions/unindexTransactions, because there will be an error when getting the transaction hashes of the block. So TxIndexTail must be less than or equal to an existing block

https://github.com/bnb-chain/bsc/blob/87e622e51fab01204f6d57712f9d86192e7a9ac5/core/rawdb/chain_iterator.go#L150-L162

klim0v:
I also suggest checking `TxIndexTail` > `expectTail` in the tryPruneHistoryBlock function

cmd/utils/flags.go
galaio: accepted

galaio:
> I would like to ask one more question, we now have 3 flags (history.state, history.transactions, history.blocks) to clear historical data, what data do we still have that is not cleared and if so, is it possible to reduce that too?

Hi @klim0v , You can refer to the inspected data from 2024 bnbchain storage report. https://www.bnbchain.org/en/blog/bnb-smart-chain-annual-storage-report-2024
image

The most burden is state history, chain history, and Index data. These data has a way to prune now. There also has some proposal for live state expiry, but it's more difficult. 

And if you have good suggestions, it's greate to create an issue/proposal or discuss it in forum/discord.

galaio:
> And I'd like to point out that we must avoid a situation where history.transactions > history.blocks. In such a case we will not be able to do indexTransactions/unindexTransactions, because there will be an error when getting the transaction hashes of the block. So TxIndexTail must be less than or equal to an existing block
> 
> https://github.com/bnb-chain/bsc/blob/87e622e51fab01204f6d57712f9d86192e7a9ac5/core/rawdb/chain_iterator.go#L150-L162

Yeah, accepted, there is a corner case, when un-indexing & indexing again in node. Because the index and ancient data are saved in different ways, so they cannot generate or prune together. But it can be handled well to avoid bad behavior.

zzzckck:
The following tools will be removed after this PR:
```
./bsc snapshot prune-state --datadir <datadir> --triesInMemory <size>
./bsc snapshot prune-block --datadir <datadir> --datadir.ancient <ancientdir> --block-amount-reserved <reserveNum>
```

buddh0 (approved):

PR #2840, opened 2025-01-03
fix: enlarge pay bid gas to adapt 7702

irrun:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


unclezoro (approved):

NathanBSC (approved):

buddh0 (approved):

unclezoro (changes requested):
not make sense

zzzckck:
`PayBidTxGasLimit = 25000` is a reasonable value, as we don't expect the bidTx consume too much gas, as it could be a zero-gasprice tx, the target address better be a EOA address, if it is a contract, it should avoid too complicated operation.

PR #2845, opened 2025-01-06
feat: wait miner finish the later multi-proposals when restarting the node;

galaio:
### Description

The PR aims to avoid validator being slashed when restarting the node, the node will propose multi blocks after enabling the https://github.com/bnb-chain/bsc/pull/2482.

So the PR will wait for the multi-block proposals to be done to avoid the slashing when restarting, and the node will wait `MaxWaitProposalInSecs`.

The default value of `MaxWaitProposalInSecs` is set to 30 seconds.   Because the average restart time in mainnet is around 30s, so the node try to wait for the next multi-proposals to be done.

> Attention: if you maintain the node with `systemd`, pls set `TimeoutStopSpec` >= 60s, it will allow the node shutdown safely when enabling the `MaxWaitProposalInSecs` > 0.

### Changes

Notable changes: 
* feat: wait miner finish the later multi-proposals when restarting the node;
* ...

### DIFF
--- a/consensus/parlia/snapshot.go
+++ b/consensus/parlia/snapshot.go
@@ -423,12 +423,8 @@ func (s *Snapshot) nexValidatorsChangeBlock() uint64 {
 // nextProposalBlock returns the validator next proposal block.
 func (s *Snapshot) nextProposalBlock(proposer common.Address) (uint64, uint64, error) {
 	validators := s.validators()
-	currentProposer := validators[s.Number/uint64(s.TurnLength)%uint64(len(validators))]
-	currentIndex := s.indexOfVal(currentProposer)
+	currentIndex := int(s.Number / uint64(s.TurnLength) % uint64(len(validators)))
 	expectIndex := s.indexOfVal(proposer)
-	if currentIndex < 0 {
-		return 0, 0, errors.New("currentProposer not in validator set")
-	}
 	if expectIndex < 0 {
 		return 0, 0, errors.New("proposer not in validator set")
 	}
--- a/consensus/parlia/snapshot.go
+++ b/consensus/parlia/snapshot.go
@@ -411,6 +411,38 @@ func (s *Snapshot) inturnValidator() common.Address {
 	return validators[offset]
 }
 
+func (s *Snapshot) nexValidatorsChangeBlock() uint64 {
+	currentEpoch := s.Number - s.Number%s.config.Epoch
+	checkLen := s.minerHistoryCheckLen()
+	if s.Number%s.config.Epoch < checkLen {
+		return currentEpoch + checkLen
+	}
+	return currentEpoch + s.config.Epoch + checkLen
+}
+
+// nextProposalBlock returns the validator next proposal block.
+func (s *Snapshot) nextProposalBlock(proposer common.Address) (uint64, uint64, error) {
+	validators := s.validators()
+	currentProposer := validators[s.Number/uint64(s.TurnLength)%uint64(len(validators))]
+	currentIndex := s.indexOfVal(currentProposer)
+	expectIndex := s.indexOfVal(proposer)
+	if expectIndex < 0 {
+		return 0, 0, errors.New("proposer not in validator set")
+	}
+	startBlock := uint64((expectIndex+len(validators)-currentIndex)%len(validators)*int(s.TurnLength)) + s.Number
+	startBlock = startBlock - startBlock%uint64(s.TurnLength)
+	endBlock := startBlock + uint64(s.TurnLength) - 1
+
+	changeValidatorsBlock := s.nexValidatorsChangeBlock()
+	if startBlock >= changeValidatorsBlock {
+		return 0, 0, errors.New("next proposal block is out of current epoch")
+	}
+	if endBlock >= changeValidatorsBlock {
+		endBlock = changeValidatorsBlock - 1
+	}
+	return startBlock, endBlock, nil
+}
+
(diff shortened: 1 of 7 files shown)


miner/worker.go
```diff
@@ -1508,6 +1508,34 @@ func (w *worker) getSealingBlock(params *generateParams) *newPayloadResult {
...
+	currentHeader := w.chain.CurrentBlock()
+	startBlock, endBlock, err := posa.NextProposalBlock(w.chain, w.chain.CurrentBlock(), w.coinbase)
+	if err != nil {
+		log.Warn("Failed to get next proposal block", "err", err)
+		return
+	}
+
+	currentBlock := currentHeader.Number.Uint64()
+	waitSecs := (endBlock - currentBlock) * posa.BlockInterval()
+	if waitSecs > 0 && waitSecs <= w.config.MaxWaitProposalInSecs {
```
emailtovamos, line 1527: Maybe we can exit early if it is negative? 
```
if waitSecs <= 0 || waitSecs > w.config.MaxWaitProposalInSecs {
	return
}
```

consensus/parlia/snapshot.go
```diff
@@ -411,6 +411,38 @@ func (s *Snapshot) inturnValidator() common.Address {
...
+	startBlock := uint64((expectIndex+len(validators)-currentIndex)%len(validators)*int(s.TurnLength)) + s.Number
+	startBlock = startBlock - startBlock%uint64(s.TurnLength)
+	endBlock := startBlock + uint64(s.TurnLength) - 1
+
+	changeValidatorsBlock := s.nexValidatorsChangeBlock()
+	if startBlock >= changeValidatorsBlock {
+		return 0, 0, errors.New("next proposal block is out of current epoch")
+	}
+	if endBlock >= changeValidatorsBlock {
+		endBlock = changeValidatorsBlock - 1
```
buddh0, line 441: -->
endBlock = changeValidatorsBlock

miner/worker.go
```diff
@@ -1508,6 +1508,34 @@ func (w *worker) getSealingBlock(params *generateParams) *newPayloadResult {
...
+
+	currentHeader := w.chain.CurrentBlock()
+	startBlock, endBlock, err := posa.NextProposalBlock(w.chain, w.chain.CurrentBlock(), w.coinbase)
+	if err != nil {
+		log.Warn("Failed to get next proposal block", "err", err)
+		return
+	}
+
+	currentBlock := currentHeader.Number.Uint64()
+	waitSecs := (endBlock - currentBlock) * posa.BlockInterval()
```
buddh0, line 1526: so, if 19 seconds later, the miner will in turn, will we shutdown the validator imediately?

buddh0, line 1526: issue: endBlock - currentBlock

consensus/parlia/snapshot.go
```diff
@@ -411,6 +411,38 @@ func (s *Snapshot) inturnValidator() common.Address {
...
+	if expectIndex < 0 {
+		return 0, 0, errors.New("proposer not in validator set")
+	}
+	startBlock := uint64((expectIndex+len(validators)-currentIndex)%len(validators)*int(s.TurnLength)) + s.Number
+	startBlock = startBlock - startBlock%uint64(s.TurnLength)
+	endBlock := startBlock + uint64(s.TurnLength) - 1
+
+	changeValidatorsBlock := s.nexValidatorsChangeBlock()
+	if startBlock >= changeValidatorsBlock {
+		return 0, 0, errors.New("next proposal block is out of current epoch")
```
zzzckck, line 438: with `turnLength == 4`, the estimated chance to hit it on mainnet would be around: `(21-1)*4/200/2 => 20%`.
Divide by two to roughly get the average value.
For example:
```
for block range of 44 -> 163(243-80): it would not hit this error
for block 164: one validator would hit it, so the rate is 1/21
for block 240: 20 validators would hit it, so the rate is 20/21
```
Basically, large `turnLength` would increase the chance to hit this error, wonder if it is possible to eliminate this error.

miner/worker.go
zzzckck: I think we can wait more time to avoid shutdown immediately after miner producing the end block for safety, maybe another 10 second?

consensus/parlia/snapshot.go
galaio: Yes, you are right, but the across-epoch scenario is too complex, I ignore it currently. Maybe we can discuss about it more~

zzzckck (dismissed):

consensus/parlia/snapshot.go
```diff
@@ -426,10 +426,13 @@ func (s *Snapshot) nextProposalBlock(proposer common.Address) (uint64, uint64, e
 	currentProposer := validators[s.Number/uint64(s.TurnLength)%uint64(len(validators))]
 	currentIndex := s.indexOfVal(currentProposer)
 	expectIndex := s.indexOfVal(proposer)
+	if currentIndex < 0 {
```
buddh0, line 429: currentIndex := s.Number/uint64(s.TurnLength)%uint64(len(validators))
so `if currentIndex < 0` is impossible

miner/worker.go
```diff
@@ -1567,3 +1571,5 @@ func signalToErr(signal int32) error {
 		panic(fmt.Errorf("undefined signal %d", signal))
 	}
 }
+
```
buddh0, line 1574: why these two lines exsits?

buddh0 (approved):

zzzckck (approved):

PR #2850, opened 2025-01-10
metric: revert default expensive metrics in blockchain;

galaio:
### Description

This PR will revert default expensive metrics in the blockchain. Previously the upstream code https://github.com/ethereum/go-ethereum/pull/29191 has removed `--metrics.expensive`, and used `NewRegisteredResettingTimer` for all time metrics. It's not suitable for BSC now, a PR https://github.com/bnb-chain/bsc/pull/2814 reverts some metrics before, but it will cause misunderstandings about execution/validation/commit metrics(it will minus db read/commit time when enabling expensive metrics).

So this PR reverts default expensive metrics and makes the blockchain time metrics work as before.

### Changes

Notable changes: 
* metric: revert default expensive metrics in blockchain;
* ...

### DIFF
--- a/core/state/state_object.go
+++ b/core/state/state_object.go
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/metrics"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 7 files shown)


zzzckck (approved):

core/state/state_object.go
```diff
@@ -224,13 +226,18 @@ func (s *stateObject) GetCommittedState(key common.Hash) common.Hash {
 	}
 	s.db.StorageLoaded++
 
-	start := time.Now()
+	var start time.Time
+	if metrics.EnabledExpensive() {
```
buddh0, line 230:
```
	var start time.Time
	if metrics.EnabledExpensive() {
		start = time.Now()
	}
```
can't save more time than 
```
start := time.Now()
```	

so, for simplicity, I think this change can be reverted

buddh0 (approved):

PR #2885, opened 2025-02-09
fix: fix pbss async node buffer force flush buffer

sysvm:
### Description

Fix `TestDisable` UT failure.

### Rationale

Trie uses `bytesPool` to improve updateTrie performance to avoid extra memory allocation. However, PBSS uses asyncnodebuffer to store data which is a async operation. So `TestDisable` UT cannot pass due to the data in PBSS is async mode and in UT use force flush.

### Example

N/A

### Changes

Notable changes: 
* N/A

### DIFF
--- a/triedb/pathdb/database_test.go
+++ b/triedb/pathdb/database_test.go
@@ -450,6 +450,10 @@ func TestDatabaseRollback(t *testing.T) {
 
 	// Verify state histories
 	tester := newTester(t, 0, false, 32)
+	bottom := tester.db.tree.bottom()
+	if err := bottom.buffer.flush(tester.db.diskdb, tester.db.freezer, bottom.nodes, bottom.id, true); err != nil {
+		t.Fatalf("Failed to force flush: %v", err)
+	}
 	defer tester.release()
 
 	if err := tester.verifyHistory(); err != nil {
@@ -520,10 +524,7 @@ func TestDatabaseRecoverable(t *testing.T) {
 	}
 }
 
-// TODO(joey): fail when using asyncbuffer
-//
-//nolint:unused
-func testDisable(t *testing.T) {
+func TestDisable(t *testing.T) {
 	// Redefine the diff layer depth allowance for faster testing.
 	maxDiffLayers = 4
 	defer func() {
@@ -531,6 +532,10 @@ func testDisable(t *testing.T) {
 	}()
 
 	tester := newTester(t, 0, false, 32)
+	bottom := tester.db.tree.bottom()
+	if err := bottom.buffer.flush(tester.db.diskdb, tester.db.freezer, nil, bottom.id, true); err != nil {
+		t.Fatalf("Failed to force flush: %v", err)
+	}
 	defer tester.release()
 
 	stored := crypto.Keccak256Hash(rawdb.ReadAccountTrieNode(tester.db.diskdb, nil))


triedb/pathdb/asyncnodebuffer.go
```diff
@@ -194,6 +194,18 @@ func (a *asyncnodebuffer) getSize() (uint64, uint64) {
...
 
+func (a *asyncnodebuffer) forceFlush(db ethdb.KeyValueStore, freezer ethdb.AncientWriter, clean *fastcache.Cache, id uint64) error {
+	for {
+		if atomic.LoadUint64(&a.background.immutable) == 1 {
+			time.Sleep(time.Duration(DefaultBackgroundFlushInterval) * time.Second)
+			log.Info("Waiting background memory table flushed into disk for forcing flush node buffer")
+			continue
+		}
+		atomic.StoreUint64(&a.current.immutable, 1)
+		return a.current.flush(db, freezer, clean, id, true)
```
joeylichang, line 205: maybe the current is switching to the background?

triedb/pathdb/asyncnodebuffer.go
```diff
@@ -194,6 +194,18 @@ func (a *asyncnodebuffer) getSize() (uint64, uint64) {
 	return a.current.size(), a.background.size()
 }
 
+func (a *asyncnodebuffer) forceFlush(db ethdb.KeyValueStore, freezer ethdb.AncientWriter, clean *fastcache.Cache, id uint64) error {
```
will-2012, line 197: What is the difference between this function and asyncnodebuffer.flush with force flag?

triedb/pathdb/disklayer.go
```diff
@@ -75,6 +75,9 @@ type trienodebuffer interface {
 
 	// getSize return the trienodebuffer used size.
 	getSize() (uint64, uint64)
+
+	// forceFlush used for asyncnodebuffer unit test
+	forceFlush(db ethdb.KeyValueStore, freezer ethdb.AncientWriter, clean *fastcache.Cache, id uint64) error
```
will-2012, line 80: `clean` seems useless.

triedb/pathdb/asyncnodebuffer.go
will-2012: If there is no semantic or behavioral difference, why not just reuse the existed method?

triedb/pathdb/database_test.go
```diff
@@ -520,17 +524,18 @@ func TestDatabaseRecoverable(t *testing.T) {
...
+func TestDisable(t *testing.T) {
 	// Redefine the diff layer depth allowance for faster testing.
 	maxDiffLayers = 4
 	defer func() {
 		maxDiffLayers = 128
 	}()
 
 	tester := newTester(t, 0, false, 32)
+	bottom := tester.db.tree.bottom()
+	if err := bottom.buffer.flush(tester.db.diskdb, tester.db.freezer, nil, bottom.id, true); err != nil {
```
will-2012, line 536: The above case includes `bottom.nodes`. Does this parameter have any effect? ​​If not, it is best to keep it consistent.

will-2012 (approved):

triedb/pathdb/database_test.go
galaio, line 536: Hi, I have a question. In actual use, are there any problems similar to UT, such as shutdown or other scenarios? Currently, the modification in UT is manually flushed once, but are there any similar problems in actual use?

galaio (dismissed):

zzzckck (approved):

NathanBSC (approved):

buddh0 (approved):

galaio (approved):

PR #2892, opened 2025-02-14
fix: wait builder's final bid during the peak time

Jolly23:
### Description

When the block gas limit is reached in the miner's local building loop, the packing process stops prematurely, even though there might be still much time till the block.timestamp. At that time, some builders may not have provided their best bids, leading to suboptimal block selection.

On February 14th, during a peak usage period on the BNB Chain mainnet, builders struggled to have their bids accepted by validators, making it difficult for bundles and private transactions to finalize.

This PR adds a simple mechanism: if there's still ample time before sealing, introduce a short wait. This ensures the validator makes full use of the 3-second window to pack transactions and wait for potentially better bids.

### Rationale

- Extends the packing window to its full duration.
- Minimizes the risk of missing higher bids submitted later.
- Improves block building efficiency, especially during peak times.

### Example

- **Before**: The loop stops very early after hitting the gas limit during the peak time, and the validator loads what may be a suboptimal best bid or even no bid coming yet.
- **After**: If the block isn't due to be sealed yet, the miner waits, giving more time for better bids to arrive.

### Changes

- Introduced a wait timer before sealing if sufficient time remains.
- Ensured validators utilize their 3-second window for potential better bids.
- Tweaked packing logic to avoid early exit.

Note: This is one of many possible approaches; for example, a separate Go routine could be used to wait for builder bids. This PR offers a simple solution.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1391,6 +1391,24 @@ LOOP:
 	// when in-turn, compare with remote work.
 	from := bestWork.coinbase
 	if w.bidFetcher != nil && bestWork.header.Difficulty.Cmp(diffInTurn) == 0 {
+		// We want to start sealing the block as late as possible here if mev is enabled, so we could give builder the chance to send their final bid.
+		// Time left till sealing the block.
+		tillSealingTime := time.Until(time.Unix(int64(bestWork.header.Time), 0)) - w.config.DelayLeftOver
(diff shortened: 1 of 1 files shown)


buddh0 (approved):

zzzckck (approved):
LGTM, as a short term solution, we can check later if the mine logic can be improved further in the future.

PR #2915, opened 2025-02-25
vdn: support init network, add event loop;

galaio:
### Description

vdn: support init network, add event loop;

### Changes

Notable changes: 
* vdn: support init network, add event loop;
* ...


emailtovamos:
Just for the understanding, a static peer doesn't have to be a VDN peer and a VDN peer doesn't have to be a static peer right?

PR #2918, opened 2025-02-25
core/txpool/legacypool/legacypool.go: add gasTip check when reset

zlacfzy:
### Description
If we're reorging an old state, reinject all dropped transactions, there are MEV transaction have zero gas price. They should not add to the txpool.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1423,8 +1423,9 @@ func (pool *LegacyPool) reset(oldHead, newHead *types.Header) {
 					}
 				}
 				lost := make([]*types.Transaction, 0, len(discarded))
+				gasTip := pool.gasTip.Load().ToBig()
 				for _, tx := range types.TxDifference(discarded, included) {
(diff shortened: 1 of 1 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1423,8 +1423,9 @@ func (pool *LegacyPool) reset(oldHead, newHead *types.Header) {
 					}
 				}
 				lost := make([]*types.Transaction, 0, len(discarded))
+				gasTip := pool.gasTip.Load().ToBig()
 				for _, tx := range types.TxDifference(discarded, included) {
-					if pool.Filter(tx) {
+					if pool.Filter(tx) && tx.GasTipCapIntCmp(gasTip) >= 0 {
```
buddh0, line 1428: may be better to compare with PriceLimit

buddh0 (approved):

emailtovamos (approved):

zzzckck (approved):

PR #2941, opened 2025-03-07
feat(vdn): add bootnode implementation

MatusKysel:
### Description

This PR is adding new Rendezvous VDN bootnode standalone binary. With this we will be able to bootstrap new network easily.

### Example

```
make all
./build/bin/vdn-bootnode -debug -external-ip 127.0.0.1
```


vdn/server.go
```diff
@@ -136,9 +124,65 @@ func (s *Server) Start() {
...
-		log.Error("Kad-DHT bootstrap failed", "err", err)
+		log.Error("DHT bootstrap failed", "err", err)
+		return
 	}
+
+	// Set up the discovery interface
+	routingDiscovery := drouting.NewRoutingDiscovery(s.dht)
+	// Advertise ourselves under the Rendezvous string
+	log.Info("Advertising Rendezvous", "topic", Rendezvous)
+	dutil.Advertise(s.ctx, routingDiscovery, Rendezvous)
```
galaio, line 150: I think your implementation of Rendezvous is not centralized. Right? It does not relay on a central point, because of every node will Advertise?

galaio: Does Rendezvous guarantee that all nodes will be discovered eventually?

PR #2943, opened 2025-03-09
refactor: replace golang.org/x/exp with stdlib

StevenMia:
### Description

Since Go 1.21, the functions in x/exp used here can already be replaced by functions from the standard library.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/rawdb/freezer.go
+++ b/core/rawdb/freezer.go
@@ -22,11 +22,11 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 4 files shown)


zzzckck (approved):
approved,  check:
https://pkg.go.dev/slices@go1.21.0
https://github.com/golang/go/issues/57433

core/rawdb/freezer.go
```diff
@@ -22,11 +22,11 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/exp/slices"
```
NathanBSC, line 30: a redundant line here

buddh0 (approved):

PR #3063, opened 2025-04-30
rpc: add method name length limit and configurable message size limit

MatusKysel:
This PR introduces two important improvements to the RPC system:

1. Method Name Length Limit
- Adds a limit of 256 characters for RPC method names to prevent potential abuse
- Prevents large method names from leading to excessive response sizes
- Limit is enforced in:
  * handleCall for regular RPC method calls
  * handleSubscribe for subscription method calls
- Added tests in websocket_test.go to verify the length limit functionality

2. Configurable WebSocket Message Size Limit
- Adds support for configuring message size limits on both client and server sides
- Previously, server side used a fixed default limit while client side supported configurable limits
- Changes include:
  * Modified WebsocketHandler to accept a messageSizeLimit parameter
  * Added messageSizeLimit field to wsConfig struct
  * Updated server code to pass the message size limit through to WebSocket handler
  * Updated test cases to pass appropriate message size limits
  * Fixed linter errors in test files by providing proper int64 values

Both changes maintain backward compatibility:
- Method name length limit only affects new calls with names exceeding 256 characters
- Message size limit defaults to wsDefaultReadLimit when not specified

### DIFF
--- a/node/api.go
+++ b/node/api.go
@@ -257,6 +257,7 @@ func (api *adminAPI) StartWS(host *string, port *int, allowedOrigins *string, ap
 		Modules: api.node.config.WSModules,
 		Origins: api.node.config.WSOrigins,
 		// ExposeAll: api.node.config.WSExposeAll,
+		messageSizeLimit: api.node.config.WSMessageSizeLimit,
 		rpcEndpointConfig: rpcEndpointConfig{
 			batchItemLimit:         api.node.config.BatchRequestLimit,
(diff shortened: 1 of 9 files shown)


alex-10072 (approved):

alex-10072:
similar PR in Go-Ethereum: https://github.com/ethereum/go-ethereum/pull/31711/files

buddh0 (approved):

PR #3097, opened 2025-05-15
refactor: use the built-in max/min to simplify the code

cuinix:
### Description

use the built-in max/min to simplify the code

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck (approved):

zzzckck:
thank you

Issue #2986, opened 2025-03-25
improvement: use default value if the option is not configure in config.toml


zzzckck:
#### System information

Geth version: `v1.5.7`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

take `DelayLeftOver` for example:
it has default value: 50ms
https://github.com/bnb-chain/bsc/blob/v1.5.7/miner/minerconfig/config.go#L56

if a validator have the following setup in its config.toml
```
[Eth.Miner]
GasCeil = 140000000
GasFloor = 140000000
GasPrice = 1000000000
Recommit = 10000000000
VoteEnable = true
````
obviously, `DelayLeftOver` is not set.
#### Expected behaviour
- we expect it use the default 50ms

#### Actual behaviour
- but it is 0 actually

#### Steps to reproduce the behaviour
NA, mark: there are lots of such kind of cases, need to be improved.

#### Backtrace
NA

**Comment by zzzckck** [comment]
better to make config.toml as simple as possible.

**Comment by zzzckck** [comment]
more options better to enable the default value, 
may refer this PR https://github.com/bnb-chain/bsc/pull/3022
```
[Eth]
TrieTimeout = 600000000000
TriesInMemory = 128

[Eth.Miner]
Recommit = 3000000000

[Eth.Miner.Mev]
Enabled = true
BuilderFeeCeil= "0"

[Eth.TxPool]
Journal = "transactions.rlp"
Rejournal = 3600000000000
PriceLimit = 1
PriceBump = 10
AccountSlots = 16
GlobalSlots = 5120
AccountQueue = 64
GlobalQueue = 1024
Lifetime = 10800000000000

[Eth.GPO]
Blocks = 20
Percentile = 60
OracleThreshold = 1000
```
With default value, it could simplify the config.toml

PR #3111, opened 2025-05-20
config: default config values if not set in config.toml

emailtovamos:
### Description

Addresses https://github.com/bnb-chain/bsc/issues/2986 with the default values mentioned here: https://github.com/bnb-chain/bsc/issues/2986#issuecomment-2853213196

### Rationale

### Example

### Changes

Notable changes: 
* ...


cmd/utils/flags.go
```diff
@@ -826,6 +826,12 @@ var (
 		Value:    strings.Join(node.DefaultConfig.GraphQLVirtualHosts, ","),
 		Category: flags.APICategory,
 	}
+	LogTimeFormatFlag = &cli.StringFlag{
+		Name:     "log.timeformat",
```
zzzckck, line 830: why need flag?

zzzckck:
Here is some of the options that better to have expected default value:
```
[Eth.Miner]
Recommit = 10000000000  //  10 seconds

[Eth.Miner.Mev]
BuilderFeeCeil= "0"

[Eth.TxPool]
Journal = "transactions.rlp"
Rejournal = 3600000000000   //     3600 seconds
PriceLimit = 1                              
PriceBump = 10                           
AccountSlots = 200
GlobalSlots = 8000
AccountQueue = 200
GlobalQueue = 4000
Lifetime = 600000000000

[Eth.GPO]
Blocks = 20
Percentile = 60
OracleThreshold = 1000

[Node.LogConfig]
TimeFormat = "01-02|15:04:05.000"
```

cmd/utils/flags.go
```diff
@@ -2491,7 +2492,7 @@ func parseMiningFeatures(ctx *cli.Context, cfg *ethconfig.Config) string {
 		return ""
 	}
 	var features []string
-	if cfg.Miner.Mev.Enabled {
```
zzzckck, line 2494: I think we don't need change the default value for MEV, as it is already false by default.

miner/bid_simulator.go
```diff
@@ -156,7 +156,7 @@ func newBidSimulator(
 
 	b.chainHeadSub = b.chain.SubscribeChainHeadEvent(b.chainHeadCh)
 
-	if config.Enabled {
```
zzzckck, line 159: I think it is not necessary as commented above, MEV is already disabled by default.

node/defaults.go
```diff
@@ -77,8 +77,13 @@ var DefaultConfig = Config{
 	},
 	DBEngine: "", // Use whatever exists, will default to Pebble if non-existent and supported
 	Instance: 1,
+	LogConfig: &LogConfig{
+		TimeFormat: stringPtr("01-02|15:04:05.000"),
+	},
 }
 
+func stringPtr(s string) *string { return &s }
```
zzzckck, line 85: I think we don't need a dedicated function to convert string to pointer.
We can have a global string variable on top and refer to it.

zzzckck (approved):

PR #3112, opened 2025-05-20
feat: add storagechange object pool for better performance

MatusKysel:
## Description
This PR introduces an object pool for `storageChange` entries in the state journal to improve memory allocation performance. The changes include:

1. Added a `sync.Pool` for `storageChange` objects to reduce GC pressure
2. Modified `storageChange` methods to work with pointers instead of values
3. Optimized memory allocations in journal operations

## Changes
- Introduced `storageChangePool` using `sync.Pool` for efficient object reuse
- Changed `storageChange` methods to use pointer receivers
- Optimized `journal.copy()` to pre-allocate slices with exact capacity
- Improved memory management in `storageChange` operations

## Performance Impact
The changes reduce memory allocations by:
- Reusing `storageChange` objects instead of creating new ones
- Reducing GC pressure through object pooling

## Benchmark
Old profile (note % not total GB as I was running it for different blocks)
Screenshot 2025-05-19 at 14 12 51
New Profile after the change 
Screenshot 2025-05-19 at 13 39 48

### DIFF
--- a/core/state/journal.go
+++ b/core/state/journal.go
@@ -167,7 +167,7 @@ func (j *journal) copy() *journal {
 		dirties:        maps.Clone(j.dirties),
 		validRevisions: slices.Clone(j.validRevisions),
 		nextRevisionId: j.nextRevisionId,
-		storageEntries: slices.Clone(j.storageEntries),
+		storageEntries: make([]*storageChange, 0, defaultNumOfSlots),
 	}
 }
 


core/state/journal.go
```diff
@@ -141,11 +161,13 @@ func (j *journal) copy() *journal {
 	for i := 0; i < j.length(); i++ {
 		entries = append(entries, j.entries[i].copy())
 	}
+
 	return &journal{
 		entries:        entries,
 		dirties:        maps.Clone(j.dirties),
 		validRevisions: slices.Clone(j.validRevisions),
 		nextRevisionId: j.nextRevisionId,
+		storageEntries: slices.Clone(j.storageEntries),
```
buddh0, line 170: the beheavior of storageChangePool is not right IMO.
after copying, a journal may still holding pointers in storageEntries which have been put back.

core/state/journal.go
```diff
@@ -396,20 +418,21 @@ func (ch codeChange) copy() journalEntry {
...
-		account:   ch.account,
-		key:       ch.key,
-		prevvalue: ch.prevvalue,
-	}
+func (ch *storageChange) copy() journalEntry {
+	cpy := storageChangePool.Get().(*storageChange)
+	cpy.account = ch.account
+	cpy.key = ch.key
+	cpy.prevvalue = ch.prevvalue
+	cpy.origvalue = ch.origvalue
```
buddh0, line 434: we may keep the beheavior as before:
```
cpy.origvalue = common.Hash{}
```

buddh0: Agree, It's a bug.
I have created a PR to feed back to go-ethereum, so they will double check this
https://github.com/ethereum/go-ethereum/pull/31874

zzzckck (approved):

PR #3158, opened 2025-06-10
feat: blind bid serves the validator's best interest

Jolly23:
### Description
Remove `mev_bestBidGasFee` RPC method and stop returning current best bid price in rejection responses to ensure true blind bidding.

### Rationale
Currently builders can discover the best bid price through:
1. Calling `BestBidGasFee` RPC method
2. Getting price feedback when their bid is rejected

This allows builders to incrementally increase bids instead of submitting their best price upfront, which reduces validator revenue.

### Changes
* Removed `BestBidGasFee` RPC method
* Removed best bid price from bid rejection responses
* Updated error messages to not reveal pricing information


miner/bid_simulator.go
```diff
@@ -434,13 +430,11 @@ func (b *bidSimulator) newBidLoop() {
 				} else if !bestBidToRun.IsCommitted() {
 					// bestBidToRun is not committed yet, this newBid will trigger bestBidToRun to commit
 					bidRuntime = bestBidRuntime
-					replyErr = genDiscardedReply(bidRuntime)
```
buddh0, line 448: so if a bid have less reward,
will result `accepted = true` in line 484

suggested keep genDiscardedReply, but change it into 
```
	genDiscardedReply := func(betterBid *BidRuntime) error {
		return fmt.Errorf("bid was discarded for offering a lower reward than competing bids")
	}
```

buddh0 (approved):

kuordstankm13545:
بهترین کارموجود

PR #3205, opened 2025-06-30
ethdb: tuning pebble compaction parameter

flywukong:
### Description
This value determine the amount of L0 read-amplification necessary to trigger an L0 compaction. Reduces the threshold for triggering compaction help make compaction more frequently and reduce the accumulated debt by 90%.  

The debt defined in pebble as bleow:
```

// An estimate of the number of bytes that need to be compacted for the LSM
// to reach a stable state.
EstimatedDebt uint64

```

### Rationale

image

the  bsc fullnode instance represented by the green line（run with pr new code） has an average compression debt of only 82.9 MiB, which is significantly lower than the average compression debt of about 8.62 GiB for the instance represented by the orange line

image
For execution performance, there also seems to be some improvement (P99).
This is because pebble's compaction negatively impacts read/write I/O,
and more frequent triggering helps reduce the negative effect of compaction on read/write performance.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/ethdb/pebble/pebble.go
+++ b/ethdb/pebble/pebble.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"fmt"
 	"runtime"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


ethdb/pebble/pebble.go
```diff
@@ -122,11 +135,30 @@ func (d *Database) onWriteStallBegin(b pebble.WriteStallBeginInfo) {
...
 
 func (d *Database) onWriteStallEnd() {
 	d.writeDelayTime.Add(int64(time.Since(d.writeDelayStartTime)))
 	d.writeStalled.Store(false)
+
+	if d.writeDelayReason != "" {
+		metrics.GetOrRegisterResettingTimer(d.namespace+"stall/time/"+d.writeDelayReason, nil).UpdateSince(d.writeDelayStartTime)
+		d.writeDelayReason = ""
+	}
+	d.writeDelayStartTime = time.Time{}
```
buddh0, line 161: useless? because line 135 ine func `onWriteStallBegin`

ethdb/pebble/pebble.go
```diff
@@ -253,6 +285,10 @@ func New(file string, cache int, handles int, namespace string, readonly bool) (
 		// By setting the WALBytesPerSync, the cached WAL writes will be periodically
 		// flushed at the background if the accumulated size exceeds this threshold.
 		WALBytesPerSync: 5 * ethdb.IdealBatchSize,
+
+		// Reduce L0CompactionThreshold from the default 4 to 2 to trigger compactions more frequently,
+		// which helps reduce compaction debt from ~10GB to below 500MB.
```
buddh0: we can keep the more detail comment from https://github.com/ethereum/go-ethereum/pull/31988/files

```
		// L0CompactionThreshold specifies the number of L0 read-amplification
		// necessary to trigger an L0 compaction. It essentially refers to the
		// number of sub-levels at the L0. For each sub-level, it contains several
		// L0 files which are non-overlapping with each other, typically produced
		// by a single memory-table flush.
		//
		// The default value in Pebble is 4, which is a bit too large to have
		// the compaction debt as around 10GB. By reducing it to 2, the compaction
		// debt will be less than 1GB, but with more frequent compactions scheduled.
```

if possible, I recommend just pick PRs https://github.com/ethereum/go-ethereum/pull/31782 and  https://github.com/ethereum/go-ethereum/pull/31988 directly

buddh0 (approved):

fynnss (approved):

zzzckck (approved):

PR #3208, opened 2025-07-02
ethdb : disable blockstore of multidatabase 

flywukong:
### Description

The online data currently does not show any benefit from separating the blockstore.
On the contrary, it leads to a decline in commit performance and increases the complexity of the DB-related code.
With this change, when using the --multidatabase mode, only trie-related data will be separated, while the statestore functionality is preserved.

### Rationale

remove blockstore which increases the complexity of upstream merges.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/triedb/database.go
+++ b/triedb/database.go
@@ -126,7 +126,7 @@ func NewDatabase(diskdb ethdb.Database, config *Config) *Database {
 		preimages = newPreimageStore(triediskdb)
 	}
 	db := &Database{
-		disk:      triediskdb,
+		disk:      diskdb,
 		config:    config,
 		preimages: preimages,
 	}
--- a/trie/hbss2pbss.go
+++ b/trie/hbss2pbss.go
@@ -68,10 +68,10 @@ func (t *Trie) resloveWithoutTrack(n node, prefix []byte) (node, error) {
 
 func (h2p *Hbss2Pbss) writeNode(pathKey []byte, n *trienode.Node, owner common.Hash) {
 	if owner == (common.Hash{}) {
-		rawdb.WriteAccountTrieNode(h2p.db.Disk(), pathKey, n.Blob)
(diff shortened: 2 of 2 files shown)


ethdb/database.go
```diff
@@ -230,12 +218,6 @@ type StateStore interface {
 	GetStateStore() Database
```
NathanBSC, line 217: what's the difference between 'StateStore' and `GetStateStore`, do we need them both?

node/node.go
```diff
@@ -775,11 +772,10 @@ func (n *Node) OpenDatabase(name string, cache, handles int, namespace string, r
...
 	var (
-		err                          error
-		stateDiskDb                  ethdb.Database
-		blockDb                      ethdb.Database
-		disableChainDbFreeze         = false
-		blockDbHandlesSize           int
+		err                  error
+		stateDiskDb          ethdb.Database
+		disableChainDbFreeze = false
+		//	blockDbHandlesSize           int
```
NathanBSC: can we delete it?

flywukong: removed

ethdb/database.go
flywukong: some place need to judge if StateStore is nil or not ,  replace the method with HasSeparateStateStore

triedb/database.go
```diff
@@ -127,7 +126,7 @@ func NewDatabase(diskdb ethdb.Database, config *Config) *Database {
 		preimages = newPreimageStore(triediskdb)
 	}
 	db := &Database{
-		disk:      diskdb,
```
allformless: ？

allformless (approved):

triedb/database.go
flywukong: this disk  is used to write code data, so change it to diskdb

fynnss (approved):

buddh0 (approved):

PR #3212, opened 2025-07-03
p2p: treat all EVN peer as trust node;

galaio:
### Description

After Maxwell, BSC introduced EVN. Currently, there are some connection problems. When EVN peers connect to each other, they are rejected due to problems such as too many peers.

In order to solve this problem, the current PR regards EVN peers as trust nodes to ensure that EVN peers can establish connections.

Due to the restrictions on EVN Peer chain registration, each validator can only register up to 5 validators at present. The local whitelist configuration is still provided, so the overall scale is controllable and will not cause functional abuse.

### Example

Registering your nodeID on the chain, or configuring it in the EVN whitelist.

```toml
[Eth]
EVNNodeIDsToAdd = ["validator NodeID", "other nodeids"]

[Node]
EnableEVNFeatures = true
EnableQuickBlockFetching = true

[Node.P2P]
EVNNodeIdsWhitelist = ["<whitelist nodeids>"]
```

### Changes

Notable changes: 
* p2p: treat all EVN peer as trust node;
* ...

### DIFF
--- a/p2p/metrics.go
+++ b/p2p/metrics.go
@@ -69,8 +69,8 @@ var (
 	serveEncHandshakeError   = metrics.NewRegisteredMeter("p2p/serves/error/rlpx/enc", nil)
 	serveProtoHandshakeError = metrics.NewRegisteredMeter("p2p/serves/error/rlpx/proto", nil)
 
-	peerLatencyStat    = metrics.NewRegisteredTimer("p2p/peers/latency", nil)
-	evnPeerLatencyStat = metrics.NewRegisteredTimer("p2p/peers/evn/latency", nil)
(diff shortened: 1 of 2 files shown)


p2p/metrics.go
```diff
@@ -69,7 +69,8 @@ var (
 	serveEncHandshakeError   = metrics.NewRegisteredMeter("p2p/serves/error/rlpx/enc", nil)
 	serveProtoHandshakeError = metrics.NewRegisteredMeter("p2p/serves/error/rlpx/proto", nil)
 
-	peerLatencyStat = metrics.NewRegisteredTimer("p2p/peers/latency", nil)
+	peerLatencyStat    = metrics.NewRegisteredTimer("p2p/peers/latency", nil)
```
zzzckck: non-evn?

buddh0 (approved):

PR #3225, opened 2025-07-11
state: extra check for nano address

zzzckck:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -170,6 +170,10 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		ProcessConsolidationQueue(&requests, evm)
 	}
 
+	err = checkNanoBlackList(statedb, evm)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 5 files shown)


core/state_processor.go
```diff
@@ -170,6 +170,10 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.StateDB, cfg
 		ProcessConsolidationQueue(&requests, evm)
 	}
 
+	if err = checkNanoBlackList(statedb, evm); err != nil {
```
buddh0: also check before FinalizeAndAssemble?

galaio: Could we add this check in tx execution? It will fail when fill transaction in mining.

zzzckck: ok, then the check can be moved to Parlia, which would be more clean.

zzzckck: there is already BlackList check in EVM level.

buddh0 (approved):

PR #3235, opened 2025-07-15
refactor: use maps.Copy for cleaner map handling

socialsister:
### Description

use maps.Copy for cleaner map handling

https://pkg.go.dev/maps#Copy

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/snapshot/difflayer.go
+++ b/core/state/snapshot/difflayer.go
@@ -19,6 +19,7 @@ package snapshot
 import (
 	"encoding/binary"
 	"fmt"
+	maps0 "maps"
 	"math"
 	"math/rand"
 	"slices"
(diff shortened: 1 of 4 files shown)


core/state/snapshot/difflayer.go
```diff
@@ -19,6 +19,7 @@ package snapshot
 import (
 	"encoding/binary"
 	"fmt"
+	maps0 "maps"
```
zzzckck: maybe we can replace "golang.org/x/exp/maps" with standard "maps"

zzzckck (dismissed):

zlacfzy (dismissed):

core/state/snapshot/difflayer.go
MatusKysel: I think it should be replaced here

socialsister: Thank you for your suggestion. However, it's unfortunate that `maps.Keys` in "golang.org/x/exp/maps" is completely different from `maps.Keys` in the standard library (due to the iterator). 

So it cannot serve as a replacement here. But we can use `Copy` from "golang.org/x/exp/maps" to simplify your code.

For more details:
- [[GitHub Issue #61538](https://github.com/golang/go/issues/61538)](https://github.com/golang/go/issues/61538)
- [[GoLang Nuts Discussion](https://groups.google.com/g/golang-nuts/c/l3jamHayWD4)](https://groups.google.com/g/golang-nuts/c/l3jamHayWD4)

Issue #3219, opened 2025-07-07
Peer Discovery Regression in v1.5.16+ (Issue #3168 Unresolved)


wdsse:
#### System information

Geth version: `geth version 1.5.17-d689f3bb-20250624`  (geth 1.5.16 has the same problems)
OS & Version: Linux

After upgrading to v1.5.16 and later v1.5.17, the node is frequently unable to discover peers or maintain stable connections. Syncing becomes extremely unreliable — blocks are downloaded sporadically every few minutes, and then the node resumes searching for peers with almost no success.

This appears to be the same issue as described in [#3168](https://github.com/bnb-chain/bsc/issues/3168), which was unfortunately closed without resolution, though the underlying problem still persists.

Typical log output:

```
Jul 07 09:35:08 bsc1 sh[3540979]: INFO [07-07|09:35:08.731] Looking for peers                        peercount=1 tried=5  static=57
Jul 07 09:35:18 bsc1 sh[3540979]: INFO [07-07|09:35:18.884] Looking for peers                        peercount=1 tried=7  static=57
Jul 07 09:35:29 bsc1 sh[3540979]: INFO [07-07|09:35:29.333] Looking for peers                        peercount=1 tried=44 static=57
........
Jul 07 09:36:28 bsc sh[3540979]: INFO [07-07|09:36:28.885] Looking for peers                        peercount=1 tried=7  static=57
Jul 07 09:36:39 bsc sh[3540979]: INFO [07-07|09:36:39.334] Looking for peers                        peercount=1 tried=44 static=57
Jul 07 09:36:53 bsc sh[3540979]: INFO [07-07|09:36:53.732] Looking for peers                        peercount=1 tried=5  static=57
Jul 07 09:37:03 bsc sh[3540979]: INFO [07-07|09:37:03.886] Looking for peers                        peercount=1 tried=7  static=57
Jul 07 09:37:14 bsc sh[3540979]: INFO [07-07|09:37:14.334] Looking for peers                        peercount=1 tried=44 static=57
Jul 07 09:37:28 bsc sh[3540979]: INFO [07-07|09:37:28.731] Looking for peers                        peercount=1 tried=5  static=57
Jul 07 09:37:38 bsc sh[3540979]: INFO [07-07|09:37:38.887] Looking for peers                        peercount=1 tried=7  static=57
[...]

PR #3237, opened 2025-07-17
p2p: opt enr filter in peer discovery;

galaio:
### Description

Recently, the community has reported that it is unable to find a peer to sync with only when using bootnodes. By checking the logs, it can be seen that the node has been discovering nodes, but the connection fails or cannot be synchronized normally.

Ref: https://github.com/bnb-chain/bsc/issues/3219

```bash
Jul 07 09:40:23 bsc sh[3540979]: INFO [07-07|09:40:23.736] Looking for peers peercount=1 tried=5 static=57
Jul 07 09:40:33 bsc sh[3540979]: INFO [07-07|09:40:33.889] Looking for peers peercount=1 tried=7 static=57
Jul 07 09:40:44 bsc sh[3540979]: INFO [07-07|09:40:44.346] Looking for peers peercount=1 tried=44 static=57
```

Through code analysis and actual testing, it is found that the bootnode node and code can be optimized to help quickly find BSC network nodes.

### Rationale

The purpose is to keep the P2P route table clean and filter non-BSC nodes, which can make the network more compact and efficient;

image

In the current bootnode implementation, if an exception such as a UDP timeout occurs, non-BSC nodes are placed in the router table and broadcast throughout the BSC P2P network, see https://github.com/bnb-chain/bsc/blob/e7b198c10cc8c3e32a5b6d98cd34938115f08ade/p2p/discover/table.go#L515-L517 This issue needs to be fixed.

In addition, many P2P nodes do not register their own ENR records for node filtering; this issue will also be fixed.

### Changes

Notable changes: 
* p2p: opt enr filter in peer discovery;
* ...

### DIFF
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -26,8 +26,6 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/cmd/utils"
-	"github.com/ethereum/go-ethereum/core"
-	"github.com/ethereum/go-ethereum/core/forkid"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p/discover"
@@ -35,7 +33,6 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/ethereum/go-ethereum/p2p/nat"
 	"github.com/ethereum/go-ethereum/p2p/netutil"
-	"github.com/ethereum/go-ethereum/params"
 )
 
 func main() {
@@ -133,8 +130,10 @@ func main() {
 			listenerAddr = natAddr
 		}
 	}
-	var eth enr.EthRecord
-	eth.ForkID = forkid.NewID(params.BSCChainConfig, core.DefaultBSCGenesisBlock().ToBlock(), uint64(0), uint64(0))
+	eth, err := discover.GetEthRecord(*networkFilter)
+	if err != nil {
+		utils.Fatalf("-network: %v", err)
+	}
 	ln.Set(enr.WithEntry("eth", &eth))
 	printNotice(&nodeKey.PublicKey, *listenerAddr)
 	cfg := discover.Config{
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -26,12 +26,17 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/cmd/utils"
+	"github.com/ethereum/go-ethereum/core"
+	"github.com/ethereum/go-ethereum/core/forkid"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/ethereum/go-ethereum/p2p/discover"
 	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/ethereum/go-ethereum/p2p/nat"
 	"github.com/ethereum/go-ethereum/p2p/netutil"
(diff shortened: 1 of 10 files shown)


eth/protocols/eth/discovery.go
```diff
@@ -59,6 +60,11 @@ func StartENRUpdater(chain *core.BlockChain, ln *enode.LocalNode) {
 }
 
 func StartENRFilter(chain *core.BlockChain, p2p *p2p.Server) {
+	// fill eth entry to local node
+	var eth enr.EthRecord
+	head := chain.CurrentHeader()
+	eth.ForkID = forkid.NewID(chain.Config(), chain.Genesis(), head.Number.Uint64(), head.Time)
+	p2p.LocalNode().Set(enr.WithEntry("eth", &eth))
```
MatusKysel: I don't think this is needed as it's already done here https://github.com/bnb-chain/bsc/blob/develop/eth/protocols/eth/handler.go#L114 -> https://github.com/bnb-chain/bsc/blob/develop/p2p/server.go#L480

cmd/bootnode/main.go
```diff
@@ -129,7 +130,11 @@ func main() {
 			listenerAddr = natAddr
 		}
 	}
-
+	eth, err := discover.GetEthRecord(*networkFilter)
+	if err != nil {
+		utils.Fatalf("-network: %v", err)
+	}
+	ln.Set(enr.WithEntry("eth", &eth))
```
MatusKysel: does this really help ? I mean that's why accepted if there was no response on ENR request - to not disconnect the bootnodes.

eth/protocols/eth/discovery.go
galaio: Good! This is a bit hidden, so this modification is useless, I will remove it.

cmd/bootnode/main.go
galaio: Em...I think all nodes with correct eth record are necessary, including bootnode nodes. In the future, the entire bsc network will enable filtering nodes, otherwise, it will not be discovered by other nodes.

p2p/discover/common.go
```diff
@@ -72,6 +72,22 @@ func ParseEthFilter(chain string) (NodeFilterFunc, error) {
...
 }
 
+func GetEthEntry(chain string) (enr.Entry, error) {
+	var eth struct {
+		ForkID forkid.ID
+		Tail   []rlp.RawValue `rlp:"tail"`
+	}
+	switch chain {
+	case "bsc":
+		eth.ForkID = forkid.NewID(params.BSCChainConfig, core.DefaultBSCGenesisBlock().ToBlock(), uint64(0), uint64(0))
```
allformless, line 82: use `NewStaticFilter` instead like `ParseEthFilter`?

p2p/config.go
```diff
@@ -76,6 +76,9 @@ type Config struct {
 	// protocol.
 	BootstrapNodesV5 []*enode.Node `toml:",omitempty"`
 
+	// EnableENRFilter enables the ENR filter for the discovery protocol.
+	EnableENRFilter bool `toml:",omitempty"`
```
allformless, line 80: default to true?

p2p/config.go
```diff
@@ -76,6 +76,9 @@ type Config struct {
 	// protocol.
 	BootstrapNodesV5 []*enode.Node `toml:",omitempty"`
 
+	// EnableENRFilter enables the ENR filter for the discovery protocol.
```
allformless, line 79: leave a todo for setting it to true by default

galaio: The node filtering logic has been optimized, so a switch with a default value of false has been added to avoid compatibility issues. After the node version is upgraded for a while, it can be set to true by default.

here
image

buddh0 (approved):

zzzckck (approved):

PR #3241, opened 2025-07-18
fix: limit builder submissions by name instead of pubkey

jetbldr:
## Summary
Change bid submission limit mechanism from builder address (pubkey) to builder name for more flexible configuration.

## Problem
The original bid limiting mechanism was based on builder address, but in some scenarios, the same builder might use different addresses, or we want to group limits by builder name.

## Solution
1. **Extend BuilderConfig struct**: Add `Name` field to support explicit builder name configuration
2. **Smart name extraction**: Auto-extract builder name from URL domain when Name is not configured
3. **Update limit logic**: Use builderName instead of address in bidSimulator
4. **Backward compatibility**: Maintain compatibility with existing config format

### 1. Config structure update
```go
type BuilderConfig struct {
    Address common.Address
    URL     string
    Name    string `toml:",omitempty"` // Builder name, if empty will use domain suffix from URL
}
```

### 2. New methods
- `GetBuilderName()`: Get builder name, prioritize configured Name, otherwise extract from URL
- `extractDomainSuffix()`: Extract builder name from URL domain

### 3. Limit logic adjustment
- bidSimulator limits by builderName instead of address
- Keep existing `maxBidsPerBuilder` config parameter

## Config example

```toml
[[builders]]
address = "0x1234567890123456789012345678901234567890"
url = "https://tokyo.builder.blockrazor.io"
name = "blockrazor"  # Optional, auto-extract from URL if not set

[[builders]]
address = "0xabcdefabcdefabcdefabcdefabcdefabcdefabcd"
url = "https://builder.flashbots.net"
# No name configured, will auto-extract as "flashbots"
```

## Backward compatibility
- ✅ Existing config files work without modification
- ✅ Auto-extract from URL when Name field is not configured
- ✅ Keep existing `maxBidsPerBuilder` config parameter

## Testing
- ✅ Test with configured Name field
- ✅ Test auto-extraction from URL
- ✅ Test various URL formats
- ✅ Test edge cases

## Impact
- **Positive**: More flexible builder grouping limit mechanism
- **No negative impact**: Fully backward compatible

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -117,8 +117,9 @@ type bidSimulator struct {
 	simBidCh chan *simBidReq
 	newBidCh chan newBidPackage
 
-	pendingMu sync.RWMutex
-	pending   map[uint64]map[common.Address]map[common.Hash]struct{} // blockNumber -> builder -> bidHash -> struct{}
+	pendingMu     sync.RWMutex
+	pending       map[uint64]map[common.Address]map[common.Hash]struct{} // blockNumber -> builder -> bidHash -> struct{}
+	pendingByName map[uint64]map[string]map[common.Hash]struct{}         // blockNumber -> builderName -> bidHash -> struct{}
 
 	bestBidMu    sync.RWMutex
 	bestBid      map[common.Hash]*BidRuntime // prevBlockHash -> bidRuntime
@@ -128,7 +129,7 @@ type bidSimulator struct {
 	simulatingBid map[common.Hash]*BidRuntime // prevBlockHash -> bidRuntime, in the process of simulation
 	bidsToSim     map[uint64][]*BidRuntime    // blockNumber -->  bidRuntime list, used to discard envs
 
-	maxBidsPerBuilder uint32 // Maximum number of bids allowed per builder per block
+	maxBidsPerBuilder uint32 // Maximum number of bids allowed per builder per block (now used for builder name)
 }
 
 func newBidSimulator(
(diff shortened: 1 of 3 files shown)


zzzckck:
Currently, each builder can have at most 6 addresses, each address can bid twice for each block at most.
With this proposal, the bid limitation will be builder granularity, rather than builder-address granularity.

2 feedback2:
- 1.lower the builder's bid for each block from 12(6*2) to 6? Builder should be self-constraint from sending to many bids.
- 2.this proposal will impact builder's bid strategy, needs to be confirmed with builders and set up a timeline for it, so builders can have enough time to migrate.

jetbldr:
@zzzckck thank you for response

🔧 Assumption Behind the Proposal
With blind bidding,  each builder may submit more blocks, increasing the validator's simulation load

**This isn’t malicious** — just rational behavior under the new rules: since builders are only competing against their own previous bids, they’ll send a new block as long as it shows any improvement.


✅ Feedback 1: Limit from 12 to 6
We’re okay with this change.
It reduces the max simulations per block from 60 to 30 — a meaningful improvement.

✅ Feedback 2: Need more time
Agreed. Since this assumption isn't backed by hard data yet, it’s reasonable to monitor post-launch behavior before enforcing stricter rules.

jetbldr:
If this limitation is going to be enforced, we suggest first updating the configuration format used by the [Good Will Alliance](https://github.com/bnb-chain/good-will-alliance).

This would simplify the implementation and make the enforcement logic much cleaner in code.

hilljessica8787-dot (approved):

PR #3294, opened 2025-08-20
fix: set all chain tables to be prunable

sysvm:
### Description

Set all chain tables to be prunable.

### Rationale

geth needs to truncate all chain tables when merging incr snapshot.

### Example

N/A

### Changes

Notable changes: 
* N/A

### DIFF
--- a/core/blockchain_test.go
+++ b/core/blockchain_test.go
@@ -3020,7 +3020,7 @@ func testDeleteRecreateSlotsAcrossManyBlocks(t *testing.T, scheme string) {
 			e.exist = false
 			e.values = nil
 		}
-		// t.Logf("block %d; adding destruct\n", e.blocknum)
+		//t.Logf("block %d; adding destruct\n", e.blocknum)
 		return tx
 	}
 	var newResurrect = func(e *expectation, b *BlockGen) *types.Transaction {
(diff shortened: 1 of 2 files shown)


core/blockchain_test.go
```diff
@@ -3020,7 +3020,7 @@ func testDeleteRecreateSlotsAcrossManyBlocks(t *testing.T, scheme string) {
 			e.exist = false
 			e.values = nil
 		}
-		//t.Logf("block %d; adding destruct\n", e.blocknum)
```
buddh0: maybe your style is better
but I think this type of change is trival, just keep as origin, to easier to do upstream.

core/blockchain_test.go
```diff
@@ -4510,14 +4510,18 @@ func testChainReorgSnapSync(t *testing.T, ancientLimit uint64) {
...
 // chain cutoff point. In this case the chain segment before the cutoff should
 // be persisted without the receipts and bodies; chain after should be persisted
 // normally.
-func TestInsertChainWithCutoff(t *testing.T) {
+//
+//nolint:unused
+func testInsertChainWithCutoff1(t *testing.T) {
 	const chainLength = 64
 
-	// Configure and generate a sample block chain
```
buddh0, line 4520: follow above

core/blockchain_test.go
sysvm: fixed

PR #3300, opened 2025-08-21
core/fitermaps: fix final block logic

allformless:
### Description

core/fitermaps: fix final block logic

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/backend.go
+++ b/eth/backend.go
@@ -862,11 +862,17 @@ func (s *Ethereum) updateFilterMapsHeads() {
 			head = newHead
 			chainView := s.newChainView(head)
 			historyCutoff, _ := s.blockchain.HistoryPruningCutoff()
-			var finalBlock uint64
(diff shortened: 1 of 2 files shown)


eth/backend.go
```diff
@@ -862,11 +862,17 @@ func (s *Ethereum) updateFilterMapsHeads() {
 			head = newHead
 			chainView := s.newChainView(head)
 			historyCutoff, _ := s.blockchain.HistoryPruningCutoff()
-			var finalBlock uint64
+			var finalBlock, currentBlock int64
 			if fb := s.blockchain.CurrentFinalBlock(); fb != nil {
-				finalBlock = fb.Number.Uint64()
```
allformless, line 867: when sync with chapel or bsc mainnet
before luban hardfork, finalBlock will always be 0, will prevent filtermaps indexing

Issue #3307, opened 2025-08-25
snapshot size issue: v1.5.16 vs v1.5.19


zzzckck:
#### System information

Geth version: `geth v1.5.19`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)

#### Expected behaviour
snapshot generated with v1.5.16 and v1.5.19 should have similar size

#### Actual behaviour
two node both synced from: 58104301 to 58588397, which is around 500k blocks, check the storage size
- with v1.5.16
prune-block block-amount-reserved=475200 & execute DB compact ==> 782GB

- with v1.5.19
 history.blocks=360000 & execute DB compact ==> 787GB
![Image](https://github.com/user-attachments/assets/e5245657-95c1-4589-8768-2f4bf58ed797)

#### Steps to reproduce the behaviour
NA

galaio:
This may be related to the lowering of the freeze limit in v1.5.19. Previously, a large amount of data was left in the pebble for the "--pruneanceint" freeze issue (see https://github.com/bnb-chain/bsc/pull/3251).

The new version will frequently trigger the migration of 30,000 blocks to ancient, causing performance issues. Currently, the number of blocks is relatively low, so you can monitor it, and the node will gradually migrate and prune it.

The size of ancient is lower than in v1.5.19, so it's likely that most of the block history is still in pebble and is being gradually migrated.

PR #3310, opened 2025-08-25
freezer: slow down freeze when live sync;

galaio:
### Description

In version 1.5.19, the freeze limit parameter was lowered. Previously, a large amount of data was left in pebble due to the "pruneanceint" feature. See https://github.com/bnb-chain/bsc/pull/3251.

This PR uses a freeze limit of 30,000 records during historical synchronization. When synchronization is about to end, it is changed to a freeze limit of 100 records to avoid performance issues.

related issue: https://github.com/bnb-chain/bsc/issues/3307

### Changes

Notable changes: 
* freezer: slow down freeze when live sync; 
* ...

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -744,7 +744,7 @@ func (f *chainFreezer) SyncAncient() error {
 }
 
 func trySlowdownFreeze(head *types.Header) {
-	if time.Since(time.Unix(int64(head.Time), 0)) > SlowdownFreezeWindow {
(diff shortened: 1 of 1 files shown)


zzzckck (approved):

zzzckck (changes requested):
1.better to optimize the big lock
2.otherwise, prefer use specific flag or config option to determine the limit, it could be much more straightforward.

NathanBSC (approved):

zzzckck (approved):

PR #3316, opened 2025-08-26
core/filtermaps: stop indexing if target block is pruned

allformless:
### Description

core/filtermaps: stop indexing if target block is pruned

### Rationale

fix https://github.com/bnb-chain/bsc/issues/3317

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -816,6 +816,7 @@ func (bc *BlockChain) loadLastState() error {
 	}
 	if posa, ok := bc.engine.(consensus.PoSA); ok {
 		if currentFinalizedHeader := posa.GetFinalizedHeader(bc, headHeader); currentFinalizedHeader != nil {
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -816,6 +816,7 @@ func (bc *BlockChain) loadLastState() error {
 	}
 	if posa, ok := bc.engine.(consensus.PoSA); ok {
 		if currentFinalizedHeader := posa.GetFinalizedHeader(bc, headHeader); currentFinalizedHeader != nil {
+			bc.currentFinalBlock.Store(currentFinalizedHeader)
```
allformless, line 819: so line https://github.com/bnb-chain/bsc/blob/develop/eth/backend.go#L378
can get the right finalized block to new filtermaps

galaio (approved):

PR #3340, opened 2025-09-05
core: rework fast node

allformless:
### Description

core: rework fast node

### Rationale

In version **v1.6.0-alpha**, the **snapshot has been integrated with triedb** under PBSS mode, which broke the original fast node functionality.
This PR aims to fix the issue and ensure compatibility in the following scenarios:

1. Legacy standalone snapshot data** created by older versions can still work with the new version.
2. Integrated snapshot data** generated by the new version can run properly.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/filtermaps/filtermaps.go
+++ b/core/filtermaps/filtermaps.go
@@ -287,6 +287,10 @@ func (f *FilterMaps) Start() {
 			log.Error("Could not load head filter map snapshot", "error", err)
 		}
 	}
+	if f.indexedView == nil {
+		log.Error("FilterMaps fail to Start, restart maybe ok!")
+		return
+	}
 	f.closeWg.Add(2)
 	go f.removeBloomBits()
 	go f.indexerLoop()


eth/handler.go
```diff
@@ -233,7 +233,8 @@ func newHandler(config *handlerConfig) (*handler, error) {
 	for _, address := range config.ProxyedValidatorAddresses {
 		h.proxyedValidatorAddressMap[address] = struct{}{}
 	}
-	if config.Sync == ethconfig.FullSync {
+	if h.chain.NoTries() {
+	} else if config.Sync == ethconfig.FullSync {
```
zzzckck, line 237: as NoTries does not support snapsync, can we move the check to line#265?

triedb/pathdb/database.go
```diff
@@ -303,9 +302,6 @@ func New(diskdb ethdb.Database, config *Config, isVerkle bool) *Database {
 // repairHistory truncates leftover state history objects, which may occur due
 // to an unclean shutdown or other unexpected reasons.
 func (db *Database) repairHistory() error {
-	if db.config.NoTries {
```
zzzckck, line 306: it is ok, as NoTires will not call PBSS after this commit.

PR #3348, opened 2025-09-10
core/rawdb: align more code with upstream

allformless:
### Description

core/rawdb: align more code with upstream

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/rawdb/chain_freezer.go
+++ b/core/rawdb/chain_freezer.go
@@ -243,7 +243,48 @@ func (f *chainFreezer) freeze(db ethdb.KeyValueStore, continueFreeze bool) {
 		//      blocks being retained. If the node is forcefully killed, it may fail to repair
 		//      itself during restart.
 		useFinalizedForFreeze := false
-		if !useFinalizedForFreeze {
+		if useFinalizedForFreeze {
+			threshold, err = f.freezeThreshold(nfdb)
+			if err != nil {
+				backoff = true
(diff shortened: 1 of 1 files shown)


core/rawdb/accessors_metadata.go
```diff
@@ -194,19 +192,3 @@ func WriteTransitionStatus(db ethdb.KeyValueWriter, data []byte) {
 		log.Crit("Failed to store the eth2 transition status", "err", err)
 	}
 }
-
-// ReadSafePointBlockNumber return the number of block that roothash save to disk
-func ReadSafePointBlockNumber(db ethdb.KeyValueReader) uint64 {
```
allformless, line 199: there is no place to call ReadSafePointBlockNumber even with version in 2022

so delete them.

core/rawdb/ancient_utils.go
```diff
@@ -143,7 +140,7 @@ func InspectFreezerTable(ancient string, freezerName string, tableName string, s
 	)
 	switch freezerName {
 	case ChainFreezerName:
-		path, tables = resolveChainFreezerDir(ancient), stateFreezerTableConfigs
+		path, tables = resolveChainFreezerDir(ancient), chainFreezerTableConfigs
```
allformless, line 143: a bug here, fixed
only impact on tool for inspecting

PR #3351, opened 2025-09-11
ethdb:  support multidb mode with dedicated snapshot DB and index DB

flywukong:
### Description
This PR primarily implements the following features:

1）Support storing snapshot data in a dedicated independent DB instance.
2）Store contract code in the trie database 
3）Support storing  transaction index data in a dedicated  independent DB database 
4）Support  separate  snapshot db after the snapshot being integrated into Pbss
This feature increases the number of databases supported by BSC in multi-DB mode from 2 to 4, providing greater scalability and deployment flexibility.

**Usage:** When --multidatabase is enabled, it means that /snapshot/txnIndex/Trie data will both run in three separate database. Run from gensis:
` ./geth --config config.toml  --datadir data-seed init --multidatabase  genesis.json 

### Rationale

make scalabilty stronger and this serves as the foundation for further sharding of the database.
image

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -153,7 +153,7 @@ func TestNodeCloseClosesDB(t *testing.T) {
 	stack, _ := New(testNodeConfig())
 	defer stack.Close()
 
-	db, err := stack.OpenDatabase("mydb", 0, 0, "", false)
+	db, err := stack.OpenDatabase("mydb", 0, 0, "", false, false)
 	if err != nil {
 		t.Fatal("can't open DB:", err)
 	}
@@ -176,7 +176,7 @@ func TestNodeOpenDatabaseFromLifecycleStart(t *testing.T) {
 	var err error
 	stack.RegisterLifecycle(&InstrumentedService{
 		startHook: func() {
-			db, err = stack.OpenDatabase("mydb", 0, 0, "", false)
+			db, err = stack.OpenDatabase("mydb", 0, 0, "", false, false)
 			if err != nil {
 				t.Fatal("can't open DB:", err)
 			}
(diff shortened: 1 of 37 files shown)


node/node_test.go
```diff
@@ -153,7 +153,7 @@ func TestNodeCloseClosesDB(t *testing.T) {
 	stack, _ := New(testNodeConfig())
 	defer stack.Close()
 
-	db, err := stack.OpenDatabase("mydb", 0, 0, "", false)
+	db, err := stack.OpenDatabase("mydb", 0, 0, "", false, false)
```
buddh0: OpenDatabase only have 5 params

triedb/pathdb/buffer.go
```diff
@@ -171,18 +183,61 @@ func (b *buffer) flush(root common.Hash, db ethdb.KeyValueStore, freezer ethdb.A
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				accounts, slots = b.states.write(snapBatch, progress, statesCache)
+				rawdb.WriteSnapshotRoot(snapBatch, root)
+				snapSize = snapBatch.ValueSize()
+				if err := snapBatch.Write(); err != nil {
+					snapErr = err
+				}
+			}()
```
buddh0, line 224: WritePersistentStateID and WriteSnapshotRoot in two separate writings
more likely to regenerate snapshots

flywukong: Yes. Multib cannot guarantee that they are written in the same batch unless  transaction is enabled  .Only the final correctness is guaranteed.   The benefits of concurrency seem to be acceptable for this low probability possibility

node/node_test.go
flywukong: fixed

triedb/pathdb/buffer.go
NathanBSC: if this can‘t easily solved right now, you can leave a todo here to mention this issue and risk

flywukong: done

NathanBSC (approved):

buddh0 (approved):

PR #3358, opened 2025-09-16
core/state: simplify statedb copy

allformless:
### Description

core/state: simplify statedb copy

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -787,17 +787,6 @@ func (s *StateDB) StateForPrefetch() *StateDB {
 	return state
 }
 
-// Copy creates a deep, independent copy of the state.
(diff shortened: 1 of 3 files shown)


buddh0:
useless difference with go-ethereum

galaio (dismissed):

buddh0:
now the `statedb` has cache now, so copy items in prefetcher is useless
so just remove `CopyDoPrefetch`, also align more code with go-ethereum

zzzckck (changes requested):
need more investigation about the TriePrefetch copy code, marked it pending first.

allformless:
will be included into next upstream

PR #3361, opened 2025-09-18
cmd/utils: set journalfile default to true

allformless:
### Description

cmd/utils: set journalfile default to true

### Rationale

### Example

### Changes

Notable changes: 
* ...


buddh0:
ignore snapshot data?
maybe fix after syncing with geth( geth also introduce file-based journal recently)

zzzckck (approved):

PR #3373, opened 2025-09-29
ethapi: reject oversize storage keys before hex decode (#32750)

MatusKysel:
## Summary
  - bail out of decodeHash when the raw hex string is longer than 32 bytes
  - keep error semantics identical while avoiding unnecessary hex decoding work


buddh0 (approved):

WTDuck2255 (approved):

WTDuck2255 (commented):
Thx

zzzckck (approved):

PR #3374, opened 2025-09-29
Implement BEP-592: Non-Consensus Based Block-Level Access List

zzzckck:
## Description
This PR implements [BEP-592: Non-Consensus Based Block-Level Access List](https://github.com/bnb-chain/BEPs/pull/592) to improve block execution performance. The implementation introduces a block access list (BAL) feature that tracks account and storage access during transaction execution, enabling efficient prefetching of state data.

Key changes include:
- Implementation of BAL data structures for generation, propagation, and usage phases
- P2P protocol enhancements (BSC/3) to support BAL transmission between peers
- Integration with mining, state management, and prefetching systems

image

### 1.About Code
**a.BAL lifecycle**
Each BAL will mainly have three stages: Generation, Propagation and Prefetch. Each stage would have a data structure to represent.

- Stage#1: BAL generation: validator will generate the BAL data during mining phase
```golang
// BlockAccessListRecord & AccountAccessListRecord are used to record access list during Tx execution.
type AccountAccessListRecord struct {
	TxIndex      uint32 // index of the first transaction in the block that accessed the account
	StorageItems map[common.Hash]StorageAccessItem
}

type BlockAccessListRecord struct {
	Version  uint32 // Version of the access list format
	Accounts map[common.Address]AccountAccessListRecord
}
```
- Stage#2: BAL propagation
BAL will be encoded before propagation, it is represented by BlockAccessListEncode
```golang
// StorageAccessItem is a single storage key that is accessed in a block.
type StorageAccessItem struct {
	TxIndex uint32 // index of the first transaction in the block that accessed the storage
	Dirty   bool   // true if the storage was modified in the block, false if it was read only
	Key     common.Hash
}

// AccountAccessListEncode & BlockAccessListEncode are for BAL serialization.
type AccountAccessListEncode struct {
	TxIndex      uint32 // index of the first transaction in the block that accessed the account
	Address      common.Address
	StorageItems []StorageAccessItem
}

type BlockAccessListEncode struct {
	Version  uint32 // Version of the access list format
	Number   uint64      // number of the block that the BAL is for
	Hash     common.Hash // hash of the block that the BAL is for
	SignData []byte // sign data for BAL
	Accounts []AccountAccessListEncode
}
```
- 3.BAL Prefetch
Current BAL is mainly used during prefetch, so the data structure would be:
```golang
// TxAccessListPrefetch & BlockAccessListPrefetch are for BAL prefetch
type StorageAccessItemPrefetch struct {
	Dirty bool
	Key   common.Hash
}

type TxAccessListPrefetch struct {
	Accounts map[common.Address][]StorageAccessItemPrefetch
}

type BlockAccessListPrefetch struct {
	AccessListItems map[uint32]TxAccessListPrefetch
}
```

**b.BAL Prune**
In order to avoid adding extra complexity to the current freezer module, BAL will only be kept in KVDB, but only for the recent `FullImmutabilityThreshold(360_000)` blocks, elder BAL will be deleted.

**c.BSC/3 P2P protocol**
BSC/3 would be used to indicate the BAL capability of the remote peer, only peer with bsc/3 protocol will be able to handle BAL data.

### 2.Try It:
Currently, validator is the only producer of BAL, so validator needs to be upgraded first, then the connected full node can get the BAL data and use it to accelerate it syncing performance.

Enable BAL both on validator and connected full nodes, it can be done by running with flag `--enablebal`, or update the config.toml:
```
...
[Node]
EnableBAL = true
...
```

### DIFF
--- a/cmd/geth/main.go
+++ b/cmd/geth/main.go
@@ -146,6 +146,7 @@ var (
 		utils.MinerRecommitIntervalFlag,
 		utils.MinerNewPayloadTimeoutFlag, // deprecated
 		utils.MinerDelayLeftoverFlag,
+		utils.EnableBALFlag,
 		// utils.MinerNewPayloadTimeout,
 		utils.NATFlag,
 		utils.NoDiscoverFlag,
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -182,6 +182,11 @@ var (
 		Usage:    "Chapel network: pre-configured Proof-of-Stake-Authority BSC test network",
 		Category: flags.EthCategory,
 	}
+	EnableBALFlag = &cli.BoolFlag{
+		Name:     "enablebal",
+		Usage:    "Enable block access list feature, validator will generate BAL for each block",
+		Category: flags.EthCategory,
+	}
 	// Dev mode
 	DeveloperFlag = &cli.BoolFlag{
 		Name:     "dev",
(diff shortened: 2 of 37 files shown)


buddh0:
will approve after ut can pass

fynnss (approved):
LGTM

PR #3375, opened 2025-09-29
Create python-package-conda.yml

Yardapestr:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


buddh0:
there is no scripts written in python in this repo

PR #3379, opened 2025-09-30
fix: add secrets config to nancy ci

flywukong:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


zzzckck:
The CI Nancy uses the main branch of https://github.com/sonatype-nexus-community/nancy-github-action/tree/main
The latest nancy requires user token to login, which can be registered under: https://ossindex.sonatype.org/

PR #3404, opened 2025-10-15
miner: validator not inturn backoff before mining

allformless:
### Description

miner: validator not inturn backoff before mining

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1323,6 +1323,25 @@ LOOP:
 			log.Debug("Not enough time for commitWork")
 			break
 		} else {
+			if !w.inTurn() && len(workList) == 1 {
+				if parliaEngine, ok := w.engine.(*parlia.Parlia); ok {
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -1323,6 +1323,25 @@ LOOP:
...
+							select {
+							case <-time.After(wait):
+							case <-interruptCh:
+								log.Debug("CommitWork interrupted: new block imported or resubmission triggered", "block", work.header.Number)
+								return
+							}
+						}
+					}
+				}
+			}
```
zzzckck, line 1344: need update delay one more time.

zzzckck: `delay := w.engine.Delay(w.chain, work.header, w.config.DelayLeftOver)` will return 425ms in this case, no it is ok.

zzzckck (approved):

PR #3417, opened 2025-10-22
p2p: define ProxyedNodeIds in Config

allformless:
### Description

p2p: define ProxyedNodeIds in Config

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -997,8 +997,7 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
 				hasher.Write(from.Bytes())
 
 				hasher.Read(hash)
-				if peer.ProxyedPeerFlag.Load() ||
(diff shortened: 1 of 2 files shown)


eth/handler.go
```diff
@@ -958,7 +997,8 @@ func (h *handler) BroadcastTransactions(txs types.Transactions) {
 				hasher.Write(from.Bytes())
 
 				hasher.Read(hash)
-				if new(big.Int).Mod(new(big.Int).SetBytes(hash), total).Cmp(direct) < 0 {
+				if peer.ProxyedPeerFlag.Load() ||
```
zzzckck: broadcast tx to proxyed peer may block the connection, block broadcast could be delayed?

NathanBSC: > broadcast tx to proxyed peer may block the connection, block broadcast could be delayed?

adjusted

zzzckck (approved):

galaio (approved):

PR #3423, opened 2025-10-28
miner: improve prefetch mining by using cached reader

allformless:
### Description

miner: improve prefetch mining by using cached reader

### Rationale

### Example

### Changes

Notable changes: 
* ...


core/state_prefetcher.go
```diff
@@ -277,7 +277,10 @@ func (p *statePrefetcher) PrefetchBAL(block *types.Block, statedb *state.StateDB
 // the transaction messages using the statedb, but any changes are discarded. The
 // only goal is to warm the state caches. Only used for mining stage.
 func (p *statePrefetcher) PrefetchMining(txs TransactionsByPriceAndNonce, header *types.Header, gasLimit uint64, statedb *state.StateDB, cfg vm.Config, interruptCh <-chan struct{}, txCurr **types.Transaction) {
-	var signer = types.MakeSigner(p.config, header.Number, header.Time)
+	var (
+		signer = types.MakeSigner(p.config, header.Number, header.Time)
+		reader = statedb.Reader()
```
flywukong, line 284: statedb should check if it is not nil beacause StateForPrefetch() may return nil

PR #3441, opened 2025-11-10
fix: opt black list checking logic;

galaio:
### Description

This PR will fix black list checking logic, it can avoid black list in 7702 tx.

### Changes

Notable changes: 
* fix: opt black list checking logic;
* ...


NathanBSC:
issue has been addressed in https://github.com/bnb-chain/bsc/pull/3436

PR #3445, opened 2025-11-12
eth: improve prefetch by using cached reader

flywukong:
### Description

improve prefetch by using cached reader

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -66,7 +66,7 @@ func (p *statePrefetcher) Prefetch(transactions types.Transactions, header *type
 
 	// Iterate over and process the individual transactions
 	for i, tx := range transactions {
-		stateCpy := statedb.StateForPrefetch() // lightweight state for prefetch with shared cache
+		stateCpy := statedb.CopyDoPrefetch()
 		workers.Go(func() error {
 			// If block precaching was interrupted, abort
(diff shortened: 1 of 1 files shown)


core/state_prefetcher.go
```diff
@@ -66,7 +66,7 @@ func (p *statePrefetcher) Prefetch(transactions types.Transactions, header *type
 
 	// Iterate over and process the individual transactions
 	for i, tx := range transactions {
-		stateCpy := statedb.CopyDoPrefetch() // closure
```
buddh0, line 69: I think it's better to keep using `CopyDoPrefetch` here:

1. All `stateCpy` instances share the same `cacheStats`, so the functionality remains correct.
2. Although `CopyDoPrefetch` is not very efficient, we can address this in the next upstream update:
   [https://github.com/ethereum/go-ethereum/pull/32134](https://github.com/ethereum/go-ethereum/pull/32134)

Also, note that `CopyDoPrefetch` will be removed in the upcoming upstream changes — see:
[https://github.com/bnb-chain/bsc/pull/3358/files](https://github.com/bnb-chain/bsc/pull/3358/files)

flywukong: reverted

zzzckck (approved):

PR #3459, opened 2025-11-24
feat: transactionReceipts auto-unsubscribe implementation

### LINKED ISSUE #3452: the number of receiptSub keep growing
#### System information
Geth version: `v1.6.3`
OS & Version: Windows/Linux/OSX
Commit hash : (if `develop`)
Arguments:

#### Description
https://github.com/bnb-chain/bsc/issues/3363

when all receipts for given txHash have been notified, the sub is still not released
...

- Expected behaviour
...

- Actual behaviour
...

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

### PR DESCRIPTION
**Pull request by zlacfzy**
### Description

fix #3452

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/filters/api.go
+++ b/eth/filters/api.go
@@ -494,7 +494,7 @@ func (api *FilterAPI) TransactionReceipts(ctx context.Context, filter *Transacti
 							}
 						}
 						if len(pending) == 0 && gracePeriod == nil {
-							gracePeriod = time.After(3 * time.Second) // Grace period for reorg handling
+							gracePeriod = time.After(12 * time.Second) // Grace period for reorg handling
 						}
 					}
 				}

### CODE REVIEW

File: eth/filters/api.go [hunk 1]
```diff
@@ -453,6 +453,7 @@ func (api *FilterAPI) TransactionReceipts(ctx context.Context, filter *Transacti
 		defer receiptsSub.Unsubscribe()
 
 		signer := types.LatestSigner(api.sys.backend.ChainConfig())
+		remaining := len(txHashes) // 0 means never auto-unsubscribe
```
NathanBSC: use a map is better
one tx will return two receipts if reorg happens

eth/filters/api.go
```diff
@@ -473,6 +474,14 @@ func (api *FilterAPI) TransactionReceipts(ctx context.Context, filter *Transacti
...
 					// Send a batch of tx receipts in one notification
 					notifier.Notify(rpcSub.ID, marshaledReceipts)
+
+					// Auto-unsubscribe when all receipts received
+					if remaining > 0 {
+						remaining -= len(receiptsWithTxs)
+						if remaining <= 0 {
+							return
+						}
+					}
```
NathanBSC, line 499: after all the txHashes matched, a timer can be set as a delay
for simplicty, can use 132s?
image

eth/filters/api.go
```diff
@@ -473,7 +477,17 @@ func (api *FilterAPI) TransactionReceipts(ctx context.Context, filter *Transacti
...
+
+					// Auto-unsubscribe when all receipts received (with grace period for reorgs)
+					if remaining > 0 {
+						remaining -= len(receiptsWithTxs)
+						if remaining <= 0 && gracePeriod == nil {
+							gracePeriod = time.After(3 * time.Second) // BSC: ~0.75s/block, wait 4 blocks
+						}
+					}
 				}
+			case <-gracePeriod:
```
NathanBSC, line 501: will panic here？

eth/filters/api.go
NathanBSC: 12 seconds is acceptable

eth/filters/api.go
zlacfzy: No, refer to https://go.dev/ref/spec#Select_statements

PR #3471, opened 2025-12-01
refactor: use strings.Builder to improve performance

zjumathcode:
### Description

strings.Builder has fewer memory allocations and better performance.
More info: [golang/go#75190](https://github.com/golang/go/issues/75190)

### Rationale

### Example

### Changes

Notable changes: 
* ...


allformless:
thx
bsc starts as a fork of go-ethereum
so if the change only improve a little, we prefer to keep align with upstream

may be you can try to merge this improvement to go-ethereum

PR #2942, opened 2025-03-08
refactor: use t.TempDir() instead of os.MkdirTemp

tgolang:
### Description

`TempDir()` is a method introduced in Go 1.15 for `testing.T`. It automatically creates a temporary directory and cleans it up after the test is complete, eliminating the hassle of manually handling errors and cleanup.  More detail about TempDir  https://pkg.go.dev/testing#B.TempDir

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes: 
* add each change in a bullet point here
* ...

PR #3479, opened 2025-12-06
refactor: use t.TempDir() instead of os.MkdirTemp

tgolang:
### Description

TempDir() is a method introduced in Go 1.15 for testing.T. It automatically creates a temporary directory and cleans it up after the test is complete, eliminating the hassle of manually handling errors and cleanup. More detail about TempDir https://pkg.go.dev/testing#B.TempDir

The new pr for https://github.com/bnb-chain/bsc/pull/2942

### Rationale

### Example

### Changes

Notable changes: 
* ...


allformless:
lint errors exist
and the modified testcase can't pass compilation

PR #3496, opened 2025-12-17
miner: enable more prefetch threads in local mining mode

flywukong:
### Description
1) Increase PrefetchMining parallelism for local mining when MEV is disabled. Keep MEV mode behavior unchanged .
2)  Replaced return with continue  in PrefetchMining to ensure the prefetch worker goroutine stays alive by skipping invalid transactions instead of terminating on error.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state_prefetcher.go
+++ b/core/state_prefetcher.go
@@ -40,8 +40,9 @@ const checkInterval = 10
 // from disk. Transactions are executed in parallel to fully leverage the
 // SSD's read performance.
 type statePrefetcher struct {
-	config *params.ChainConfig // Chain configuration options
-	chain  *HeaderChain        // Canonical block chain
+	config     *params.ChainConfig // Chain configuration options
+	chain      *HeaderChain        // Canonical block chain
+	mevEnabled bool                // Indicate whether MEV is enabled
 }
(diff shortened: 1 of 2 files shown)


core/state_prefetcher.go
```diff
@@ -285,8 +291,14 @@ func (p *statePrefetcher) PrefetchMining(txs TransactionsByPriceAndNonce, header
...
 
-	txCh := make(chan *types.Transaction, 2*prefetchMiningThread)
-	for i := 0; i < prefetchMiningThread; i++ {
+	// When MEV is not enabled, use more threads for local mining
+	threadCount := prefetchMiningThread
+	if !p.mevEnabled {
+		threadCount = max(prefetchMiningThread, 3*runtime.NumCPU()/5)
+	}
+
+	txCh := make(chan *types.Transaction, 2*threadCount)
```
NathanBSC, line 300: 3/5*2=6/5 now, is right?

buddh0 (dismissed):

buddh0 (approved):

zzzckck (approved):

Issue #3487, opened 2025-12-10
v1.6.4 BAD BLOCK


pantaovay:
#### System information
Geth version: 1.6.4
OS & Version: Ubuntu 24.04.3

#### Description
- latest Geth: 1.6.4
- latest Snapshot: Pruned Snapshot | mainnet-geth-pbss-20251205-pruneancient

- Expected behaviour

- Actual behaviour
`
[...]

PR #1840, opened 2023-08-24
Big merge v1.10.16 v1.12.2

NathanBSC:
### Description

upstream from go-ethereum [v1.10.16, v1.12.2]

### Rationale

import new features, bugfix, code clear up from go-ethereum

### Example

1. `make test` pass
2. `make truffle-test` pass
3.  can work as miner using node-deploy(branch [adapt_for_big_merge](https://github.com/bnb-chain/node-deploy/tree/adapt_for_big_merge))
4. Testnet full sync from genesis to the latest block: passed
5. full sync mainnet works ( sync 5M blocks for test)
6. transfer using metamask pass

#### RPC
[internal/ethapi: add debug_getRawReceipts RPC method (#24773)](https://github.com/bnb-chain/bsc/pull/1840/commits/ae7d834bc752a2d94fef9d354ee78fcb9425f3d1) 
[node, rpc: add ReadHeaderTimeout config option (#25338)](https://github.com/bnb-chain/bsc/pull/1840/commits/9244f87dc1c8869a2632176f719e515217720a43) 
[rpc: check that "version" is "2.0" in request objects (#25570)](https://github.com/bnb-chain/bsc/pull/1840/commits/38e002f4641c2779c897ccaca575ec5ddeee9254) 
[rpc: support injecting HTTP headers through context (#26023)](https://github.com/bnb-chain/bsc/pull/1840/commits/add337e0f7bad02f3cf535c66cd31f252b0b5c99) 
[rpc: websocket should respect the "HTTP_PROXY" by default (#27264)](https://github.com/bnb-chain/bsc/pull/1840/commits/73697529994e14996b7740730481e926d5ec3e40)
[rpc: change BlockNumber constant values to match ethclient (#27219)](https://github.com/bnb-chain/bsc/pull/1840/commits/9231770811cda0473a7fa4e2bccc95bf62aae634) 
[eth: make debug_StorageRangeAt take a block hash or number (#27328)](https://github.com/bnb-chain/bsc/pull/1840/commits/d789c68b667e13eb5cefd19d09ae84f7d016df6a)
[eth,core: add api debug_getTrieFlushInterval (#27303)](https://github.com/bnb-chain/bsc/pull/1840/commits/0783cb7d91ad7b3cdf72ac6c6edaec8318673eb6) 
[rpc: add limit for batch request items and response size (#26681)](https://github.com/bnb-chain/bsc/pull/1840/commits/f3314bb6df4c86e650f0e47cbb5a21ca0616ac11) 
[...]

PR #3501, opened 2025-12-24
statedb.go: fix NoTries mode snapshot storage delete

zlacfzy:
### Description

In NoTries mode (`--tries-verify-mode none`), storage data is not properly deleted from the snapshot when `handleDestruction` is called during SELFDESTRUCT.

### Rationale

In NoTries mode, `stateObject.updateRoot()` always sets `Root` to `EmptyRootHash` regardless of actual storage content:
```
if s.db.db.NoTries() {
    s.data.Root = types.EmptyRootHash
}
```
This causes two issues in the SELFDESTRUCT flow:

1. `handleDestruction` skips storage deletion: The check `prev.Root == types.EmptyRootHash `incorrectly assumes no storage exists, bypassing `deleteStorage()`.
2. `fastDeleteStorage` hash verification fails: Even if `deleteStorage` is called, the hash check `stack.Hash() != root` fails because `root` is `EmptyRootHash` but the actual storage hash is not.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1289,7 +1289,7 @@ func (s *StateDB) handleDestruction(noStorageWiping bool) (map[common.Hash]*acco
 		// In NoTries mode, prev.Root is always EmptyRootHash (see updateRoot()),
 		// so we cannot rely on it to determine if storage exists.
 		// We must try to delete storage via snapshot regardless.
-		if !s.db.NoTries() && (prev.Root == types.EmptyRootHash || s.db.TrieDB().IsVerkle()) {
+		if (!s.db.NoTries() && prev.Root == types.EmptyRootHash) || s.db.TrieDB().IsVerkle() {
 			continue
 		}
 		if noStorageWiping {
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -1169,7 +1169,9 @@ func (s *StateDB) fastDeleteStorage(snaps *snapshot.Tree, addrHash common.Hash, 
 	if err := iter.Error(); err != nil { // error might occur during iteration
 		return nil, nil, nil, err
 	}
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -1284,7 +1286,10 @@ func (s *StateDB) handleDestruction(noStorageWiping bool) (map[common.Hash]*acco
 		deletes[addrHash] = op
 
 		// Short circuit if the origin storage was empty.
-		if prev.Root == types.EmptyRootHash || s.db.TrieDB().IsVerkle() {
+		// In NoTries mode, prev.Root is always EmptyRootHash (see updateRoot()),
+		// so we cannot rely on it to determine if storage exists.
+		// We must try to delete storage via snapshot regardless.
+		if !s.db.NoTries() && (prev.Root == types.EmptyRootHash || s.db.TrieDB().IsVerkle()) {
```
buddh0: there is no relation between 's.db.NoTries()' and 's.db.TrieDB().IsVerkle()' , so I prefer
```
if (!s.db.NoTries() && prev.Root == types.EmptyRootHash) || s.db.TrieDB().IsVerkle() {
```

zlacfzy: Agreed.

NathanBSC:
this bug exists more than 2 years, no later than PR https://github.com/bnb-chain/bsc/pull/1840/files

NathanBSC:
after cancun, with EIP-6780, no wiping out old storages then.
so people download fastnode snapshots and run, this issue will not happen

buddh0 (approved):

buddh0:
fix issue https://github.com/bnb-chain/bsc/issues/3487

zzzckck (approved):

PR #3530, opened 2026-01-21
fix: Prevent overflow in keybytesToHex allocation size

odaysec:
Update keybytesToHex function to prevent overflow on 32-bit platforms by using uint64 for length calculation.

### DIFF
--- a/accounts/usbwallet/ledger.go
+++ b/accounts/usbwallet/ledger.go
@@ -509,9 +509,15 @@ func (w *ledgerDriver) ledgerSignTypedMessage(derivationPath []uint32, domainHas
 //	Optional APDU data       | arbitrary
 func (w *ledgerDriver) ledgerExchange(opcode ledgerOpcode, p1 ledgerParam1, p2 ledgerParam2, data []byte) ([]byte, error) {
 	// Construct the message payload, possibly split into multiple chunks
-	apdu := make([]byte, 2, 7+len(data))
+	// APDU body length is 5 bytes of header plus the data length. It is encoded
+	// as a 16-bit big-endian value, so enforce that it fits and avoid int overflow.
+	apduLen := 5 + len(data)
+	if apduLen < 5 || apduLen > 0xffff {
+		return nil, errors.New("APDU payload too large")
+	}
+	apdu := make([]byte, 2, 2+apduLen)
 
-	binary.BigEndian.PutUint16(apdu, uint16(5+len(data)))
+	binary.BigEndian.PutUint16(apdu, uint16(apduLen))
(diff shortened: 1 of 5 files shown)


odaysec (commented):
fixes addresses potential integer overflow vulnerabilities in allocation size calculations across multiple modules. The issue arises when computing buffer sizes using `len(data) + constant` without validating that the result fits within `int` limits, which could lead to allocation overflows on 32-bit architectures or with exceptionally large inputs.


fix follows a consistent approach: either (1) validate input size bounds before performing arithmetic, or (2) compute sizes in a wider integer type and verify the result fits within the target type before allocation.

#### `ledgerExchange`
- **Issue**: APDU body length calculation `5 + len(data)` could overflow `int` or exceed the encodable 16-bit length field.
- **Fix**: 
  - Compute `apduLen := 5 + len(data)` using `int` after validating that `len(data)` does not cause overflow.
  - Enforce maximum allowed `len(data)` such that `apduLen ≤ math.MaxUint16`.
  - Return an error (`"APDU payload too large"`) if limits are exceeded.
  - Use `apduLen` consistently for both `PutUint16` and buffer capacity:  
    `apdu := make([]byte, 2, 2+apduLen)`.

####  `trezorExchange`
- **Issue**: Buffer allocation uses `8 + len(data)` without overflow checks.
- **Fix**:
  - Define a reasonable maximum message size constant (`trezorMaxMessageSize`).
  - Validate `len(data)` against both `trezorMaxMessageSize` and `math.MaxInt-8`.
  - Return an error for oversized payloads before allocation.

#### `Encode`
- **Issue**: `len(b)*2 + 2` may overflow `int` for extremely large inputs.
- **Fix**:
  - Compute required length in `uint64`: `need := uint64(len(b))*2 + 2`.
  - If `need > math.MaxInt`, panic with a clear message (consistent with low-level helper behavior).
  - Otherwise, allocate using `int(need)`.

#### `processRequestsSystemCall`
- **Issue**: `len(ret) + 1` could overflow in `make([]byte, len(ret)+1)`.
- **Fix**:
  - Check `len(ret) ≤ math.MaxInt-1` before allocation.
  - Return `fmt.Errorf("system call response too large")` if the check fails.

#### `keybytesToHex`
- **Issue**: `l := len(str)*2 + 1` may overflow when passed to `make`.
- **Fix**:
  - Compute `l64 := uint64(len(str))*2 + 1`.
  - If `l64 > math.MaxInt`, panic (consistent with existing trie invariants).
  - Otherwise, use `l := int(l64)` for allocation.

buddh0:
IMO, these checks are not really need

PR #3520, opened 2026-01-14
BEP648: Enhanced Fast Finality via InMemory Voting Pool

zlacfzy:
### Description

[BEP-648: Enhanced Fast Finality via In-Memory Voting Pool](https://github.com/zlacfzy/BEPs/blob/one_block_finality/BEPs/BEP-648.md)

### Rationale

tell us why we need these changes...

### Example

add an example CLI or API response...

### Changes

Notable changes: 
* add each change in a bullet point here
* ...

PR #3547, opened 2026-02-04
BEP648: fix race condition in early finalization.

zlacfzy:
### Description
It's a fix for #3520
Fix a critical race condition in CheckFinalityAndNotify where the function relied on currentHeader matching targetBlockHash. Since currentHeader is updated later than highestVerifiedBlock in writeBlockAndSetHead, votes entering curVotes (via GetVerifiedBlockByHash check) would trigger CheckFinalityAndNotify before currentHeader was updated, causing the check to fail.

### Rationale

In writeBlockAndSetHead, highestVerifiedBlock is set before currentHeader. When votes arrive and GetVerifiedBlockByHash returns non-nil, they enter curVotes and trigger CheckFinalityAndNotify. However, currentHeader hasn't been updated yet, so the check `currentHeader.Hash() != targetBlockHash` fails, preventing early finalization from working correctly.

### Example

Before fix:
- CheckFinalityAndNotify skipped due to "currentHeader mismatch"
- Early finalization rarely triggered via VotePool

After fix:
- Target header retrieved directly by GetHeaderByHash
- Early finalization works reliably

### Changes

Notable changes:
* CheckFinalityAndNotify: Use GetHeaderByHash(targetBlockHash) instead of 
  checking currentHeader.Hash() == targetBlockHash
* VotePool: Add early filtering using HighestNotifiedFinal() to skip 
  already-finalized blocks
* BlockChain: Add HighestNotifiedFinal() getter for deduplication
* Parlia: Remove finalizedNotified LRU cache

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -265,10 +265,6 @@ type Parlia struct {
 	slashABI                   abi.ABI
 	stakeHubABI                abi.ABI
 
-	// finalizedNotified tracks blocks that have already triggered early finalization notification
-	// to avoid duplicate notifications
-	finalizedNotified *lru.Cache[common.Hash, struct{}]
-
 	// The fields below are for testing only
 	fakeDiff bool // Skip difficulty verifications
 }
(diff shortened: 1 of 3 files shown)


NathanBSC (approved):

allformless (approved):

galaio (approved):

maizas2504-code (approved):

maizas2504-code:
جيد

maizas2504-code (approved):

Issue #3548, opened 2026-02-04
Block synchronization issues


morexcess:
#### System information
Geth version: `geth version`
OS & Version: Linux
Commit hash : (if `develop`)
Arguments: 1.6.5 and 1.6.6

#### Description
- Steps to reproduce the behaviour
...After upgrading to version 1.6.5 and running normally for about half a month, nodes that were functioning normally suddenly stopped catching up with blocks. I recently upgraded the client to version 1.6.6, but the problem persists. I've reviewed the logs but haven't found any particularly significant errors. This issue usually resolves itself by restarting the client or waiting a long time, but it reappears at an unknown time in the future.

- Expected behaviour
...

- Actual behaviour
...

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

zlacfzy:
Could you upload full log?

morexcess:
Which part of the logs do you need? Do you need the logs that don't catch up to the points before and after the block? This is difficult to capture because it doesn't appear all the time; it only happens occasionally.

zlacfzy:
Yes, it will help; i couldn’t reproduce it locally.

morexcess:
Okay, please don't close it then. I'll send you the captured logs if any problems occur.

morexcess:
02-08|10:44:21.958 lvl=info msg="Imported new chain segment" number=80004140 hash=0xa6945f20035250370c89f758859b90045ce149ba3bc67a1765399b81fa07c95f miner=0x38944092685a336CB6B9ea58836436709a2adC89 blocks=1 txs=86 blobs=0 mgas=12.755491 elapsed=34.808ms mgasps=366.4476655165738 BAL=false triediffs="140.23 MiB" triedirty="253.68 MiB" trieimutabledirty="0.00 B"
t=02-08|10:44:22.408 lvl=info msg="Imported new chain segment" number=80004141 hash=0xf8c1080c18be08523205ef2f704f291f8f6a6948a6e591c0d9ba84b6ef855a99 miner=0x38944092685a336CB6B9ea58836436709a2adC89 blocks=1 txs=72 blobs=0 mgas=10.352535 elapsed=35.441ms mgasps=292.1004401143783 BAL=false triediffs="140.04 MiB" triedirty="253.81 MiB" trieimutabledirty="0.00 B"
[...]

Issue #3525, opened 2026-01-19
sync issue


zerochl:
#### System information
Geth version: `1.6.6`
OS & Version: Linux
Commit hash : (if `develop`)
Arguments:

#### Description
- Steps to reproduce the behaviour
when start the node, sync is ok, but run some time, will show the issue, restart will return to normal:
t=01-18|23:33:05.382 lvl=warn msg="Synchronisation failed, retrying" peer=6a7470e4e4721844e3bce9673ad2e43f62f9825f24d56715bcab5acafee6e39d err="peer is unknown or unhealthy"
t=01-18|23:33:36.444 lvl=warn msg="Synchronisation failed, retrying" peer=917f44826990a548f023fa360c1516a64f73d2d4fd29bce5b1918b45a0e19be4 err=timeout
t=01-18|23:34:11.518 lvl=warn msg="Synchronisation failed, retrying" peer=e710dad043bf8b6900f0442039927b32070083c89c612a10ab455c0f846d0f18 err=timeout
t=01-18|23:34:30.743 lvl=warn msg="Synchronisation failed, retrying" peer=5caf79211c0bb12c541c635782f0fab9b54b848688a95159844fa2f6c87a55ca err="shutting down"

zlacfzy:
This is normal. The warning occurs when a peer stops or shuts down.

zerochl:
However, the problem is that this issue occurs frequently, and once it happens, it freezes for about a minute, and no sync in 1minute. 
Can this be optimized?

zlacfzy:
Testnet or mainnet?

zerochl:
mainnet

zerochl:
你们节点绝对有问题，我经常跑着跑着突然就不同步了，此时要等很久很久，但是如果这个时候我重启节点，立马就恢复

zlacfzy:
Could you please upload the full log file for this case? It would help with the investigation.

morexcess:
我也遇到了同样的问题，正常运行一段时间以后，突然就追不上块，过一段时间大概半小时以后就又同步好了。能够正常运行一段时间

zerochl:
我最近魔改了你们节点，但是还是会有问题，仍旧是重启节点可以迅速恢复

zlacfzy:
We need more logs about this case. 

> 我也遇到了同样的问题，正常运行一段时间以后，突然就追不上块，过一段时间大概半小时以后就又同步好了。能够正常运行一段时间

[...]

PR #3556, opened 2026-02-10
eth/downloader, eth/sync: reduce sync stall on BSC's 0.45s block inte…

zlacfzy:
### Description

Reduce sync stalls on BSC after the block interval was 0.45s. For #3525 #3548
 
1. Cap critical-path header timeout – Introduce masterPeerHeaderTimeout (5s) to bound blocking fetchHeadersByHash / fetchHeadersByNumber. The dynamic TargetTimeout can exceed 10s, stalling the entire pipeline on one slow peer. Also add missing cancelCh handling in fetchHeadersByHash.

2. Preserve forced sync on failure – Keep cs.forced = true on doSync error, so the sync loop retries immediately with minPeers=1 instead of waiting up to 10s.

3. Shorten near-head wait – Replace blocking 2s time.Sleep with a non-blocking 500ms select in nextSyncOp, responsive to quit signals and peer events.

### Rationale

With 0.45s block intervals, the old parameters (designed for 3s/12s cadence) cause compounding delays: a single slow peer triggers a 10s+ timeout, followed by a 10s recovery wait — ~45 blocks of drift total. These changes reduce worst-case sync stall from ~20s to ~5s.

### Changes

* eth/downloader/fetchers.go: Add masterPeerHeaderTimeout = 5s; cap TTL in both header fetch functions; add cancelCh in fetchHeadersByHash
* eth/sync.go: Preserve cs.forced = true on sync failure for immediate retry; replace 2s blocking sleep with 500ms non-blocking select

### DIFF
--- a/eth/sync.go
+++ b/eth/sync.go
@@ -117,10 +117,16 @@ func (cs *chainSyncer) loop() {
 		select {
 		case <-cs.peerEventCh:
 			// Peer information changed, recheck.
-		case <-cs.doneCh:
+		case err := <-cs.doneCh:
 			cs.doneCh = nil
-			cs.force.Reset(forceSyncCycle)
-			cs.forced = false
+			if err != nil {
+				// Sync failed, keep forced=true so nextSyncOp can immediately
+				// retry with minPeers=1 instead of waiting for the 10s force timer.
+				cs.forced = true
+			} else {
+				cs.force.Reset(forceSyncCycle)
+				cs.forced = false
+			}
 		case <-cs.force.C:
 			cs.forced = true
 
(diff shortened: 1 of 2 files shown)


NathanBSC:
maybe can solve this issue by adjust  https://github.com/bnb-chain/bsc/blob/master/eth/downloader/downloader.go#L61 ？

eth/sync.go
```diff
@@ -184,11 +190,17 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
 		// 		log.Info("Disable transaction acceptance randomly for the delay exceeding 10 blocks.")
 		// 	}
 	} else if op.td.Cmp(new(big.Int).Add(ourTD, common.Big2)) <= 0 { // common.Big2: difficulty of an in-turn block
-		// On BSC, blocks are produced much faster than on Ethereum.
-		// If the node is only slightly behind (e.g., 1 block), syncing is unnecessary.
-		// It's likely still processing broadcasted blocks(such as including a big tx) or block hash announcements.
-		// In most cases, the node will catch up within 2 seconds.
-		time.Sleep(2 * time.Second)
```
flywukong, line 191: can we just change it to  time.Sleep(500 * time.Millisecond)

eth/sync.go
```diff
@@ -184,11 +190,17 @@ func (cs *chainSyncer) nextSyncOp() *chainSyncOp {
...
+		// Only slightly behind (~1 block). Wait briefly for block broadcast to
+		// catch up before starting a full sync. Use non-blocking select to stay
+		// responsive to quit signals and peer events.
+		waitTimer := time.NewTimer(500 * time.Millisecond)
+		defer waitTimer.Stop()
+		select {
+		case <-waitTimer.C:
+		case <-cs.handler.quitSync:
+			return nil
+		case <-cs.peerEventCh:
```
allformless, line 202: don't need trigger by this signal `peerEventCh`

PR #3584, opened 2026-03-06
core/vm: align super-instruction maxStack with raw bounds

MatusKysel:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vm/jump_table.go
+++ b/core/vm/jump_table.go
@@ -1145,14 +1145,14 @@ func createOptimizedOpcodeTable(tbl *JumpTable) *JumpTable {
 		execute:     opPush2Jump,
 		constantGas: GasFastestStep + GasMidStep,
 		minStack:    minStack(0, 0),
-		maxStack:    maxStack(0, 0),
+		maxStack:    maxStack(0, 1), // intermediate PUSH2 requires stack_len <= 1023
 	}
 
 	tbl[Push2JumpI] = &operation{
 		execute:     opPush2JumpI,
 		constantGas: GasFastestStep + GasSlowStep,
 		minStack:    minStack(1, 0),
-		maxStack:    maxStack(1, 0),
+		maxStack:    maxStack(0, 1), // intermediate PUSH2 requires stack_len <= 1023
 	}
 
 	tbl[Push1Push1] = &operation{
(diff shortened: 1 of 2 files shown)


annielz (dismissed):

core/vm/jump_table.go
```diff
@@ -1145,14 +1145,14 @@ func createOptimizedOpcodeTable(tbl *JumpTable) *JumpTable {
...
-		maxStack:    maxStack(0, 0),
+		maxStack:    maxStack(0, 1), // intermediate PUSH2 requires stack_len <= 1023
 	}
 
 	tbl[Push2JumpI] = &operation{
 		execute:     opPush2JumpI,
 		constantGas: GasFastestStep + GasSlowStep,
 		minStack:    minStack(1, 0),
-		maxStack:    maxStack(1, 0),
+		maxStack:    maxStack(0, 1), // intermediate PUSH2 requires stack_len <= 1023
```
NathanBSC, line 1155: maxStack(0, 1) --> maxStack(1, 1)？

core/vm/jump_table.go
```diff
@@ -1166,14 +1166,14 @@ func createOptimizedOpcodeTable(tbl *JumpTable) *JumpTable {
...
-		maxStack:    maxStack(1, 1),
+		maxStack:    maxStack(0, 1), // intermediate PUSH1 requires stack_len <= 1023
 	}
 
 	tbl[Push1Shl] = &operation{
 		execute:     opPush1Shl,
 		constantGas: 2 * GasFastestStep,
 		minStack:    minStack(1, 1),
-		maxStack:    maxStack(1, 1),
+		maxStack:    maxStack(0, 1), // intermediate PUSH1 requires stack_len <= 1023
```
NathanBSC, line 1176: why not maxStack(1, 1)?

MatusKysel: If you looked only at the net effect, you'd use `maxStack(1, 1)` = 1024, meaning the stack could be completely full (1024 items) when this super-instruction starts. But that's wrong: The intermediate state matters. When `PUSH2` executes first, it pushes 1 item onto the stack before `JUMP` pops anything. If the stack starts at 1024, `PUSH2` would push it to 1025 — exceeding the limit. So the correct constraint comes from the tightest intermediate bound, which is `PUSH2`'s requirement: the stack must be ≤ 1023 when it runs. That's maxStack(0, 1) = 1024 + 0 - 1 = 1023.

MatusKysel: sorry this is for the upper question, but logic is the same

allformless:
create a PR to this branch
[core/vm: align super-instruction minStack with raw bounds](https://github.com/bnb-chain/bsc/pull/3588)

allformless (approved):

annielz (approved):

PR #3588, opened 2026-03-10
core/vm: align super-instruction minStack with raw bounds

allformless:
### Description

core/vm: align super-instruction minStack with raw bounds

### Bug fixes

Five super-instructions had incorrect `minStack` values that allowed the pre-execution stack check to pass with fewer items than the raw opcode sequence actually requires. When execution then proceeded, an intermediate opcode would attempt to access stack slots that did not exist, causing a Go-level panic rather than a graceful `ErrStackUnderflow`. This is a consensus-safety bug: the super-instruction path diverges from the raw opcode path at stack underflow boundaries.

Fixed instructions:

| Instruction | Old minStack | Correct minStack | Tightest intermediate step |
|---|---|---|---|
| `AndSwap1PopSwap2Swap1` | 2 | 5 | SWAP2 (after AND+SWAP1+POP, delta=-2, needs ≥3) |
| `Swap1Pop` | 1 | 2 | SWAP1 (needs ≥2) |
| `PopJump` | 1 | 2 | JUMP (after POP, delta=-1, needs ≥1 → initial ≥2) |
| `AndDup2AddSwap1Dup2LT` | 1 | 2 | tightest intermediate requires ≥2 |
| `Swap1Push1Dup1NotSwap2AddAndDup2AddSwap1Dup2LT` | 1 | 2 | tightest intermediate requires ≥2 |

### maxStack alignment (no functional impact)

Several super-instructions had their `maxStack` adjusted to match the value derived from the tightest intermediate constraint. All adjusted values remain above 1024 and therefore do not affect runtime behavior — since the EVM stack starts empty and every instruction maintains the invariant that post-execution depth ≤ 1024, a `maxStack > 1024` check can never trigger in practice. The values are left above 1024 intentionally, consistent with how standard opcodes like `POP` (`maxStack=1025`) are defined.

### Test improvements (`TestSuperInstructionStackBoundary`)

- Renamed from `TestSuperInstructionMaxStackBoundary` to reflect that both `minStack` and `maxStack` are now verified.
- Cases reordered to match the declaration order in `createOptimizedOpcodeTable` (excluding `Nop`), making it easier to spot missing entries.
- Coverage expanded from 22 to 31 cases, now including all super-instructions defined in `createOptimizedOpcodeTable`. For each instruction the test computes the tightest `minStack`/`maxStack` by simulating the raw opcode sequence with cumulative delta tracking, then asserts the optimized table entry matches exactly.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vm/instructions_test.go
+++ b/core/vm/instructions_test.go
@@ -1779,13 +1779,22 @@ func TestOpSwap2Swap1Dup3SubSwap2Dup3GtPush2(t *testing.T) {
 	require.Equal(t, scope2.Memory.Data(), scope1.Memory.Data())
 }
 
-// TestSuperInstructionMaxStackBoundary verifies that super-instructions enforce
-// the same stack overflow behavior as their raw opcode sequences at the EVM
-// stack limit (1024). This is a regression test for a consensus-safety bug where
-// fused instructions used net-effect maxStack instead of the tightest
-// intermediate bound, allowing execution to succeed where raw semantics would
-// fail with ErrStackOverflow.
-func TestSuperInstructionMaxStackBoundary(t *testing.T) {
+// TestSuperInstructionStackBoundary verifies that every super-instruction in
+// createOptimizedOpcodeTable has the same minStack and maxStack as the tightest
+// intermediate constraint across its raw opcode sequence. A super-instruction
+// skips per-step stack checks, so its bounds must reflect the strictest
+// individual requirement; otherwise execution can diverge from the raw sequence
+// at stack boundaries (underflow or overflow), which is a consensus bug.
+//
+// For each raw op at cumulative stack delta d, the constraint on the initial
+// stack depth s is:
+//
+//	minStack[op] - d <= s <= maxStack[op] - d
+//
+// The tightest bounds are the max of all lower constraints and the min of all
+// upper constraints across the sequence.
+func TestSuperInstructionStackBoundary(t *testing.T) {
+	// log.SetDefault(log.NewLogger(log.NewTerminalHandlerWithLevel(os.Stderr, log.LevelDebug, true)))
 	baseTbl := newCancunInstructionSet()
 	optTbl := createOptimizedOpcodeTable(copyJumpTable(&baseTbl))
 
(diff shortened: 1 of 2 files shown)


asdfg43a-bit (approved):

annielz (approved):

core/vm/instructions_test.go
```diff
@@ -1795,74 +1804,85 @@ func TestSuperInstructionMaxStackBoundary(t *testing.T) {
...
+			if optOp.minStack != tightestMinStack {
+				t.Errorf("minStack %d != tightest raw constraint %d; super-instruction is more permissive than raw sequence at stack lower boundary",
+					optOp.minStack, tightestMinStack)
+			}
+			if optOp.maxStack != tightestMaxStack {
+				t.Errorf("maxStack %d != tightest raw constraint %d; super-instruction is more permissive than raw sequence at stack upper boundary",
+					optOp.maxStack, tightestMaxStack)
 			}
+
+			// log.Debug(tc.name, "minStack", tightestMinStack, "maxStack", tightestMaxStack)
```
MatusKysel: plz remove it

core/vm/jump_table.go
```diff
@@ -1186,22 +1186,22 @@ func createOptimizedOpcodeTable(tbl *JumpTable) *JumpTable {
...
+		maxStack:    maxStack(0, 0),
 	}
 
 	tbl[PopJump] = &operation{
 		execute:     opPopJump,
 		constantGas: GasQuickStep + GasMidStep,
-		minStack:    minStack(1, 0),
-		maxStack:    maxStack(1, 0),
+		minStack:    minStack(2, 0),
+		maxStack:    maxStack(1, 0), // tightest intermediate: PUSH2 at stack S-2 requires S <= 1025
```
MatusKysel: this comment is not right

core/vm/jump_table.go
```diff
@@ -1186,22 +1186,22 @@ func createOptimizedOpcodeTable(tbl *JumpTable) *JumpTable {
...
 	}
 
 	tbl[Pop2] = &operation{
 		execute:     opPop2,
 		constantGas: 2 * GasQuickStep,
 		minStack:    minStack(2, 0),
-		maxStack:    maxStack(2, 0),
+		maxStack:    maxStack(1, 0), // tightest intermediate: PUSH2 at stack S-2 requires S <= 1025
```
MatusKysel: also this one

core/vm/jump_table.go
MatusKysel: because there is not push

core/vm/instructions_test.go
```diff
@@ -1779,13 +1779,22 @@ func TestOpSwap2Swap1Dup3SubSwap2Dup3GtPush2(t *testing.T) {
...
+//
+// For each raw op at cumulative stack delta d, the constraint on the initial
+// stack depth s is:
+//
+//	minStack[op] - d <= s <= maxStack[op] - d
+//
+// The tightest bounds are the max of all lower constraints and the min of all
+// upper constraints across the sequence.
+func TestSuperInstructionStackBoundary(t *testing.T) {
+	// log.SetDefault(log.NewLogger(log.NewTerminalHandlerWithLevel(os.Stderr, log.LevelDebug, true)))
```
MatusKysel, line 1797: also remove this I guess

MatusKysel (approved):

core/vm/instructions_test.go
allformless: > also remove this I guess

maybe we can keep it and line:1885
1. many other test cases have similar code
2. if new super instruction is deployed, we can use these logs to output the right 'minStack' and 'maxStack'

PR #3591, opened 2026-03-12
consensus/parlia: increment nonce only when tx is included

allformless:
### Description

consensus/parlia: increment nonce only when tx is included

### Rationale
the effect is nearly to revert [consensus/parlia: set nonce before evm run](https://github.com/bnb-chain/bsc/pull/2185/changes)

The old code itself had a miner/validator inconsistency — when slash reverted, the miner's nonce behavior differed from the validator's, causing any such block to be rejected by all validators and become an orphan.

This actually proves:
1. Under the old code, a slash revert would inevitably orphan the block.
2. BSC mainnet has no slash-related orphan blocks, which means slash has never reverted.
3. Therefore this code path has never been executed on mainnet, and this change carries no consensus-split risk.
4. The new code is actually more correct — miner and validator behavior is now fully consistent.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/consensus/parlia/parlia.go
+++ b/consensus/parlia/parlia.go
@@ -2191,7 +2191,6 @@ func (p *Parlia) applyTransaction(
 		*receivedTxs = (*receivedTxs)[1:]
 	}
 	state.SetTxContext(expectedTx.Hash(), len(*txs))
-	*txs = append(*txs, expectedTx)
 
 	// Create a new context to be used in the EVM environment
 	context := core.NewEVMBlockContext(header, chainContext, nil)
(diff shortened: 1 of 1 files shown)


flywukong (dismissed):

zlacfzy (dismissed):

consensus/parlia/parlia.go
```diff
@@ -2193,6 +2191,7 @@ func (p *Parlia) applyTransaction(
 		*receivedTxs = (*receivedTxs)[1:]
 	}
 	state.SetTxContext(expectedTx.Hash(), len(*txs))
+	*txs = append(*txs, expectedTx)
```
zlacfzy: append tx here but receipts missed

consensus/parlia/parlia.go
```diff
@@ -2230,6 +2228,8 @@ func (p *Parlia) applyTransaction(
 		return err
 	}
 	*txs = append(*txs, expectedTx)
+	// increment nonce only when tx is included
+	state.SetNonce(msg.From, state.GetNonce(msg.From)+1, tracing.NonceChangeEoACall)
```
zlacfzy: just use nonce don't need get again?

zlacfzy (approved):

flywukong (approved):

PR #3600, opened 2026-03-18
metrics: replace time.Tick with time.NewTicker to prevent resource leaks

haoyu-haoyu:
## Summary

Replace all 6 instances of `time.Tick()` in the `metrics/` package with `time.NewTicker()` + `defer ticker.Stop()`.

`time.Tick()` creates a ticker whose underlying resources can never be released — there is no handle to call `Stop()`. This is flagged by `go vet` (SA1015) and is a recognized Go anti-pattern.

### Changes

| File | Function |
|------|----------|
| `metrics/debug.go` | `CaptureDebugGCStats` |
| `metrics/json.go` | `WriteJSON` |
| `metrics/log.go` | `LogScaled` |
| `metrics/opentsdb.go` | `OpenTSDBWithConfig` |
| `metrics/syslog.go` | `Syslog` |
| `metrics/writer.go` | `Write` |

### Pattern

```diff
-for range time.Tick(d) {
+ticker := time.NewTicker(d)
+defer ticker.Stop()
+for range ticker.C {
```

While these goroutines run for the process lifetime, using `NewTicker` + `defer Stop()` follows Go best practices and enables proper cleanup if these functions are ever modified to have return paths.

Note: this same issue exists in upstream go-ethereum `metrics/` package.

### DIFF
--- a/metrics/debug.go
+++ b/metrics/debug.go
@@ -22,7 +22,9 @@ var (
 // CaptureDebugGCStats captures new values for the Go garbage collector statistics
 // exported in debug.GCStats. This is designed to be called as a goroutine.
 func CaptureDebugGCStats(r Registry, d time.Duration) {
-	for range time.Tick(d) {
+	ticker := time.NewTicker(d)
+	defer ticker.Stop()
+	for range ticker.C {
 		CaptureDebugGCStatsOnce(r)
 	}
 }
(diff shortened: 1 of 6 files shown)


allformless:
'for range time.Tick(d)'  is a infinite loop, it never return , so there is no chance to call 'Stop func'

PR #3601, opened 2026-03-18
core: reject future chasing heads for DA checks

MatusKysel:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1396,9 +1396,6 @@ func (bc *BlockChain) SnapSyncCommitHead(hash common.Hash) error {
 
 // UpdateChasingHead update remote best chain head, used by DA check now.
 func (bc *BlockChain) UpdateChasingHead(head *types.Header) {
-	if head == nil || head.Number == nil {
-		return
-	}
 	if head.Time > uint64(time.Now().Unix()) {
 		log.Warn("Ignoring future chasing head", "number", head.Number, "time", head.Time)
 		return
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -1396,7 +1396,14 @@ func (bc *BlockChain) SnapSyncCommitHead(hash common.Hash) error {
 
 // UpdateChasingHead update remote best chain head, used by DA check now.
 func (bc *BlockChain) UpdateChasingHead(head *types.Header) {
-	bc.chasingHead.Store(head)
+	if head == nil || head.Number == nil {
```
NathanBSC: impossible
because it's checked when decode msg from peer

keep it is also ok

core/data_availability.go
```diff
@@ -66,11 +66,19 @@ func IsDataAvailable(chain consensus.ChainHeaderReader, block *types.Block) (err
...
 	highest := chain.ChasingHead()
 	current := chain.CurrentHeader()
-	if highest == nil || highest.Number.Cmp(current.Number) < 0 {
+	now := uint64(time.Now().Unix())
+	if highest == nil || highest.Number == nil || highest.Time > now {
 		highest = current
+	} else if current != nil && current.Number != nil && highest.Number.Cmp(current.Number) < 0 {
+		highest = current
+	}
+	if highest == nil || highest.Number == nil {
```
NathanBSC: so if highest == nil
firstly highest is assigned to current
then highest is assigned to block.Header()?

no need to modify IsDataAvailable IMP

core/blockchain.go
MatusKysel: you are right, I will remove it. We will have less code and less condition to check

core/data_availability.go
MatusKysel: fixed

allformless (approved):

flywukong (approved):

PR #3603, opened 2026-03-19
handler.go: add peer info in logs

zlacfzy:
### Description

print peer id when fetch block

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/handler.go
+++ b/eth/handler.go
@@ -314,15 +314,21 @@ func newHandler(config *handlerConfig) (*handler, error) {
 		return h.chain.InsertChain(blocks)
 	}
 
-	broadcastBlockWithCheck := func(block *types.Block, propagate bool) {
+	broadcastBlockWithCheck := func(peer string, block *types.Block, propagate bool) {
 		if propagate {
 			if !(block.Header().WithdrawalsHash == nil && block.Withdrawals() == nil) &&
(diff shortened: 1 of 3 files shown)


allformless (approved):

eth/handler.go
```diff
@@ -317,15 +314,21 @@ func newHandler(config *handlerConfig) (*handler, error) {
...
 			}
 			if err := core.IsDataAvailable(h.chain, block); err != nil {
-				log.Error("Propagating block with invalid sidecars", "number", block.Number(), "hash", block.Hash(), "err", err)
+				var peerAddr string
+				if p := h.peers.peer(peer); p != nil {
+					if addr := p.RemoteAddr(); addr != nil {
+						peerAddr = addr.String()
+					}
+				}
+				log.Error("Propagating block with invalid sidecars", "number", block.Number(), "hash", block.Hash(), "peer", peer[:16], "peerAddr", peerAddr, "err", err)
```
flywukong, line 331: if len(peerID) > 16 {
    peerID = peerID[:16]
}.         consider to check the length in case of future change , panic risk

zlacfzy: The peer ID is always a 64-character hex string, so it will never be shorter than 16

flywukong (approved):

allformless (approved):

PR #3618, opened 2026-03-24
miner: async blob tx validation during bid simulation

flywukong:
### Description

Async blob validation is launched at the beginning of simBid, right before prepareWork. The validation result is consumed lazily in commitTransaction when the first blob tx is encountered.

If a bid contains N transactions and the first blob tx is at position K, the effective parallel window equals the prepareWork setup time plus the execution time of the preceding K-1 non-blob transactions. With a typical simBid duration of ~14ms and KZG proof verification at ~1.8ms per blob, the parallel window is usually sufficient to fully absorb the verification cost — especially since blob transactions are unlikely to appear at the very beginning of a bid.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_simulator.go
+++ b/miner/bid_simulator.go
@@ -1007,9 +1007,8 @@ type BidRuntime struct {
 	packedBlockReward     *big.Int
 	packedValidatorReward *big.Int
 
-	finished      chan struct{}
-	duration      time.Duration
-	blobValidated bool // true after async blob validation result has been consumed
+	finished chan struct{}
+	duration time.Duration
 }
 
 func newBidRuntime(newBid *types.Bid, validatorCommission uint64) (*BidRuntime, error) {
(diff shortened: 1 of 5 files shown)


miner/bid_simulator.go
```diff
@@ -1080,9 +1084,14 @@ func (r *BidRuntime) commitTransaction(chain *core.BlockChain, chainConfig *para
 			return errors.New("cell proof is not supported yet")
 		}
 
-		// Validate blob sidecar commitment hashes and KZG proofs.
-		if err := txpool.ValidateBlobTx(tx, env.header, nil); err != nil {
-			return err
+		if ch, ok := r.bid.BlobValResults[tx.Hash()]; ok {
+			if err := <-ch; err != nil {
```
allformless, line 1089: if a bad bid contain one blobtx twice, will it stuck forever?

flywukong: Fixed by deleting the map entry after consuming the channel result. It may cause one extra synchronous re-validation,  the probability is extremely small

zlacfzy (approved):

allformless (approved):

PR #3623, opened 2026-03-26
core/vm: reject duplicate bridge validators at Pasteur

MatusKysel:
## Summary
  - add Pasteur-specific precompile variants for BLS signature verification (`0x66`) and CometBFT light block validation (`0x67`)
  - reject duplicate bridge validator identities at Pasteur for both BLS pubkey aggregation inputs and light-client validator sets
  - add regression coverage for Pasteur precompile selection and duplicate-validator rejection paths

  ## Details
  This change introduces Pasteur-only behavior for the bridge validator precompiles.

  For BLS signature verification, the Pasteur variant rejects inputs that repeat the same validator pubkey in the aggregated signer set. Pre-Pasteur behavior is preserved.

  For CometBFT light block validation, the Pasteur variant validates that validator sets do not contain duplicate identities. The validation covers validator address, consensus pubkey, BLS key, and relayer
  address, and is applied when decoding both the trusted consensus state and the incoming light block validator set.

  ## Why
  Duplicate bridge validators can allow malformed validator sets or signer lists to bypass uniqueness assumptions in bridge verification logic. Pasteur now enforces uniqueness at the precompile boundary instead
  of relying on callers to sanitize inputs.


mliu233 (approved):

annielz (dismissed):

core/vm/lightclient/v2/lightclient.go
```diff
@@ -36,6 +36,59 @@ type ConsensusState struct {
...
+func newValidatorDuplicateTracker(field string, size int) validatorDuplicateTracker {
+	return validatorDuplicateTracker{
+		field: field,
+		seen:  make(map[string]int, size),
+	}
+}
+
+func (t validatorDuplicateTracker) check(idx int, value []byte) error {
+	key := string(value)
+	if firstIdx, ok := t.seen[key]; ok {
```
NathanBSC, line 63: are bls key and RelayerAddress allowed to stay unset? represent by zero value

core/vm/lightclient/v2/lightclient.go
```diff
@@ -137,7 +190,7 @@ func (cs *ConsensusState) ApplyLightBlock(block *types.LightBlock, isHertz bool)
 // input:
 // | chainID   | height   | nextValidatorSetHash | [{validator pubkey, voting power, relayer address, relayer bls pubkey}] |
 // | 32 bytes  | 8 bytes  | 32 bytes             | [{32 bytes, 8 bytes, 20 bytes, 48 bytes}]                               |
-func DecodeConsensusState(input []byte) (ConsensusState, error) {
```
NathanBSC, line 140: how about 
1. func DecodeConsensusState(input []byte) (ConsensusState, error) 
 --> func DecodeConsensusState(input []byte, requireUniqueValidators bool) (ConsensusState, error) 
2. delete DecodeConsensusStateWithValidation

core/vm/lightclient/v2/lightclient.go
```diff
@@ -192,14 +245,27 @@ func DecodeConsensusState(input []byte) (ConsensusState, error) {
...
+}
+
+func DecodeConsensusStateWithValidation(input []byte) (ConsensusState, error) {
+	return decodeConsensusState(input, true)
+}
+
 // input:
 // consensus state length | consensus state | light block |
 // 32 bytes               |                 |             |
-func DecodeLightBlockValidationInput(input []byte) (*ConsensusState, *types.LightBlock, error) {
```
NathanBSC, line 202: how about 
1. func DecodeLightBlockValidationInput(input []byte) (*ConsensusState, *types.LightBlock, error) 
 --> func DecodeLightBlockValidationInput(input []byte, requireUniqueValidators bool) (*ConsensusState, *types.LightBlock, error) 
2. delete DecodeLightBlockValidationInputWithValidation

core/vm/contracts_lightclient.go
```diff
@@ -430,13 +439,25 @@ type cometBFTLightBlockValidateHertz struct {
...
 func (c *cometBFTLightBlockValidateHertz) Run(input []byte) (result []byte, err error) {
-	return c.run(input, true)
+	return c.run(input, true, false)
 }
 
 func (c *cometBFTLightBlockValidateHertz) Name() string {
 	return "COMET_BFT_LIGHT_BLOCK_VALIDATE_HERTZ"
 }
 
+type cometBFTLightBlockValidateHertzMendel struct {
```
allformless: how about 
cometBFTLightBlockValidateHertzMendel --> cometBFTLightBlockValidateMendel?
because when Mendel enabled, Hertz must be already enalbed

allformless:
delay to Pasteur HF after discussion

core/vm/lightclient/v2/lightclient.go
MatusKysel: fixed

core/vm/lightclient/v2/lightclient.go
MatusKysel: fixed

core/vm/lightclient/v2/lightclient.go
MatusKysel: Yes. Those two fields are allowed to be unset upstream. In our fixed-width decode path, "unset" is represented as zero-filled bytes, so I treat empty/nil and all-zero as the same unset state and exclude only
  that case from duplicate detection.

mliu233 (approved):

zlacfzy:
0x66 (blsSignatureVerify) is left unchanged — it is a generic BLS primitive and does not require unique pubkeys.

core/vm/contracts.go
```diff
@@ -350,13 +350,22 @@ var PrecompiledContractsOsaka = PrecompiledContracts{
 	common.BytesToAddress([]byte{0x1, 0x00}): &p256Verify{eip7951: true},
 }
 
+var PrecompiledContractsPasteur = func() PrecompiledContracts {
```
NathanBSC: list precompiles directly like others

core/vm/contracts.go
```diff
@@ -417,6 +426,9 @@ func init() {
 	for k := range PrecompiledContractsPrague {
 		PrecompiledAddressesPrague = append(PrecompiledAddressesPrague, k)
 	}
+	for k := range PrecompiledContractsPasteur {
```
NathanBSC: move after PrecompiledContractsOsaka

andyzhang2023 (approved):

NathanBSC (approved):

allformless (approved):

flywukong (approved):

PR #3628, opened 2026-03-31
core/vote: cap future vote hashes per height

MatusKysel:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


allformless:
too cost to do an attack, impossible

PR #3631, opened 2026-04-07
core/vote: fix deadlock in votepool when stop client

allformless:
### Description

core/vote: fix deadlock in votepool when stop client

### Rationale

### Example

### Changes

Notable changes: 
* ...


zlacfzy (approved):

flywukong (approved):

rodriguezvallerobertocarlos6-arch (commented):
Image

mikwespinoza-ux (approved):

PR #3641, opened 2026-04-12
refactor: use slices.Backward to simplify the code

chuanshanjida:
### Description

There is a [new function](https://pkg.go.dev/slices@go1.23.0#Backward)  added in the go1.23 standard library, which can make the code more concise and easy to read.

More info can see  https://github.com/golang/go/issues/61899

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2849,11 +2849,11 @@ func (bc *BlockChain) insertSideChain(block *types.Block, it *insertIterator, ma
 		blocks []*types.Block
 		memory uint64
 	)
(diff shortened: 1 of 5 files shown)


allformless:
Thanks for the patch! Honestly, your approach is cleaner.

That said, BSC shares a lot of code with go-ethereum and needs to stay in sync with upstream periodically. If a change is purely cosmetic, I'd lean toward not merging it here — otherwise we risk introducing unnecessary merge conflicts down the road.

For changes like this, I'd recommend submitting to go-ethereum first. Thanks!

PR #3660, opened 2026-04-28
all: post quantum migration poc

fynnss:
### Description

Migrates BSC fast-finality vote signatures from BLS-12-381 to **ML-DSA-44** (NIST FIPS 204),
and introduces a new `PQTxType (0x05)` transaction type whose sender is authenticated via an
on-chain PQ registry contract (`0x70`) rather than ECDSA recovery. Vote attestations are
aggregated using a STARK recursive proof instead of BLS linear aggregation. A new `bsc4` P2P
sub-protocol carries `PQVotesMsg` alongside the existing `bsc3` vote messages.

### Rationale

BLS-12-381 is broken in polynomial time by Shor's algorithm on a cryptographically relevant
quantum computer. Validator signatures are permanently on-chain, making BSC vulnerable to
**harvest-now, decrypt-later** attacks. NIST finalised ML-DSA (FIPS 204) in August 2024 as
the primary post-quantum signature standard. ML-DSA-44 was selected over ML-DSA-65 because
it matches the effective security level of BLS-12-381 (~128-bit post-quantum) at 35% smaller
key/signature size, and its single-core verification throughput (~1,500/s) leaves the 450 ms
block slot with ample headroom — verification is not a bottleneck.

### Changes

* **ML-DSA-44 transaction type** (`PQTxType = 0x05`): transactions carry an explicit 1312-byte
  public key and 2420-byte ML-DSA signature; sender address is derived as `keccak256(pubkey)[12:]`
  and resolved via the on-chain PQ Registry at `0x70`
* **PQ Registry contract (`0x70`)**: stores `addr → pubkey` mappings; protected against EIP-158
  storage wipe by pinning a non-zero nonce in genesis
* **STARK-based vote attestation** (`PQVoteAttestation`): replaces BLS aggregate signature with
  a STARK recursive proof over the validator committee; `pqAssembleVoteAttestation` /
  `pqVerifyVoteAttestation` gate on `IsPQFork`
* **PQVotePool / PQVoteManager**: pool stores and deduplicates `PQVoteEnvelope`s; manager signs
  votes with the local ML-DSA key and submits them when the node is an active PQ validator
* **`bsc4` P2P sub-protocol**: adds `PQVotesMsg` for broadcasting and syncing PQ votes across
  peers without conflicting with existing `bsc3` BLS vote messages
* **`--pqvotekey <path>` flag**: operator opt-in; loads raw ML-DSA private key and starts
  `PQVoteManager`; node always instantiates `PQVotePool` so received PQ votes are stored even
  without a local key
* **PQ registry cache warm-up**: on startup, `WarmPQRegistryCache` is seeded from the current
  validator set via `Parlia.CurrentValidators()` (snapshot-based, epoch-independent) to prevent
  vote failures immediately after restart
* **Snapshot `PQVoteAddress` back-fill**: `loadSnapshot` calls `backfillPQVoteAddresses()` to
  fill zero-value entries from the process-level cache; `pqVerifyVoteAttestation` applies a
  second point-of-use back-fill, preventing committee root mismatch on nodes restarted before
  cache warm-up completes


johnnyleevn85 (commented):
Stop changing my funds

PR #3669, opened 2026-05-07
miner: reduce local mining time for last block in one turn

allformless:
### Description

miner: reduce local mining time for last block in one turn

### Rationale

This PR fixes two issues:

1. **Empty first block in a validator's turn (local mining)**: When a validator takes
the full `BlockInterval` to mine the last block of its turn, the next validator receives
it too late and has little time left to build a non-empty block. The fix caps the
last-in-turn block's local mining time to `BlockInterval/4`, giving the next validator
sufficient lead time.

2. **Insufficient bid simulation window for the last block in a turn**: Previously the
mining time cap (`BlockInterval/2`) was applied uniformly inside `Delay()`, which also
shortened the bid simulation window for the last block. The fix moves the cap into a
local-mining-only wrapper in `worker.go`, so bid simulators always get the full
`BlockInterval` window regardless of turn position.

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -1208,6 +1208,26 @@ func (w *worker) generateWork(genParam *generateParams, witness bool) *newPayloa
 	}
 }
 
+// delay returns the time budget for local block building.
+// For the last block in a validator's turn it caps the budget to BlockInterval/4,
+// giving the next validator enough lead time to avoid producing an empty block.
+// (post-Fermi: 450ms interval, ~120ms one-way p2p delay)
+func (w *worker) delay(header *types.Header, leftOver *time.Duration) *time.Duration {
+	d := w.engine.Delay(w.chain, header, leftOver)
+	if d == nil {
+		return nil
+	}
+	if p, ok := w.engine.(*parlia.Parlia); ok && p.IsLastBlockInTurn(w.chain, header) {
+		blockInterval, err := p.BlockInterval(w.chain, header)
+		if err == nil {
+			if cap := time.Duration(blockInterval) * time.Millisecond / 4; *d > cap {
+				return &cap
+			}
+		}
+	}
+	return d
+}
+
 // commitWork generates several new sealing tasks based on the parent block
 // and submit them to the sealer.
 func (w *worker) commitWork(interruptCh chan int32, timestamp int64) {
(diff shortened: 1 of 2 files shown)


zlacfzy (approved):

miner/worker.go
```diff
@@ -1348,7 +1368,7 @@ LOOP:
 				log.Debug("commitWork interruptCh closed, new block imported or resubmit triggered")
 				return
 			case ev := <-txsCh:
-				delay := w.engine.Delay(w.chain, work.header, w.config.DelayLeftOver)
+				delay := w.delay(work.header, w.config.DelayLeftOver)
```
flywukong, line 1371: One remaining concern is the local mining cap semantics. After a fill finishes early,
  commitWork resets stopTimer to headerTime - DelayLeftOver, then may wait for new txs and
  trigger another fill round. Each round is capped, but repeated rounds can still keep local
  mining active beyond BlockInterval/4. If the intended behavior is total local mining time <=
  BlockInterval/4 for the last in-turn block, this probably needs an absolute deadline shared by
  the wait/rebuild loop.

miner/worker.go
```diff
@@ -1208,6 +1208,26 @@ func (w *worker) generateWork(genParam *generateParams, witness bool) *newPayloa
...
+// (post-Fermi: 450ms interval, ~120ms one-way p2p delay)
+func (w *worker) delay(header *types.Header, leftOver *time.Duration) *time.Duration {
+	d := w.engine.Delay(w.chain, header, leftOver)
+	if d == nil {
+		return nil
+	}
+	if p, ok := w.engine.(*parlia.Parlia); ok && p.IsLastBlockInTurn(w.chain, header) {
+		blockInterval, err := p.BlockInterval(w.chain, header)
+		if err == nil {
+			if cap := time.Duration(blockInterval) * time.Millisecond / 4; *d > cap {
```
flywukong, line 1223: If a future fork brings it below ~400ms, BlockInterval/4 drops below the typical one-way p2p latency,

miner/worker.go
allformless: later fill also base an empty env

miner/worker.go
allformless: In a long future, no plan to reduce block interval again.
besides, "drops below the typical one-way p2p latency" won't cause any issue

flywukong (approved):

PR #3671, opened 2026-05-09
eth/protocols/bsc: cap GetBlocksByRange response size

zlacfzy:
### Description
Cap `GetBlocksByRange` response size and reuse RLP bytes on the send path.

### Rationale
Today the handler accumulates only `block.Size() + block.BALSize()` and never bounds the total against the 10MB `maxMessageSize` hard cap. Sidecars are not counted at all, so a batch of 64 BAL/blob-rich blocks can overflow the limit and drop the peer — after the server has already paid the full encode cost.

This change mirrors eth/66's body-serve pattern: measure each entry's exact RLP size (sidecars and BAL included), stop before `softResponseLimit` (8MB) would overflow, and send `[]rlp.RawValue` to avoid a second encode pass in `p2p.Send`. Wire format is unchanged.

### Changes
- `handleGetBlocksByRange`: measure exact size via `rlp.EncodeToBytes`; break before `softResponseLimit` (8MB) would overflow; always include the start block, even when it alone exceeds the soft limit.
- Hoist `MaxRequestRangeBlocksCount` and `softResponseLimit` into a single `const (...)` block at the top of `handler.go`.
- Add `BlocksByRangeRLPPacket` (`Blocks []rlp.RawValue`) and use it on the send path. Wire bytes are identical to `BlocksByRangePacket`; receivers are unchanged.

### DIFF
--- a/eth/protocols/bsc/handler.go
+++ b/eth/protocols/bsc/handler.go
@@ -13,9 +13,20 @@ import (
 	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/p2p/enr"
+	"github.com/ethereum/go-ethereum/rlp"
 )
 
-const MaxRequestRangeBlocksCount = 64
+const (
+	// MaxRequestRangeBlocksCount is the maximum number of blocks to serve per
+	// GetBlocksByRange request. Mostly there to bound disk lookups; with BSC's
+	// typical block size, the practical limit will always be softResponseLimit.
+	MaxRequestRangeBlocksCount = 64
+
+	// softResponseLimit is the target maximum size of replies. The hard cap on
+	// the wire is maxMessageSize (10MB); 8MB leaves headroom for outer RLP/p2p
+	// framing. Each entry's measured size includes sidecars and BAL.
+	softResponseLimit = 8 * 1024 * 1024
+)
 
 // Handler is a callback to invoke from an outside runner after the boilerplate
 // exchanges have passed.
@@ -168,7 +179,7 @@ func handleGetBlocksByRange(backend Backend, msg Decoder, peer *Peer) error {
 	}
 
 	// Get requested blocks
-	blocks := make([]*BlockData, 0, req.Count)
+	blocks := make([]rlp.RawValue, 0, req.Count)
 	var block *types.Block
 	// Prioritize blockHash query, get block & sidecars from db
 	if req.StartBlockHash != (common.Hash{}) {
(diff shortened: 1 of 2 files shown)


allformless (dismissed):

eth/protocols/bsc/handler.go
```diff
@@ -179,19 +190,28 @@ func handleGetBlocksByRange(backend Backend, msg Decoder, peer *Peer) error {
...
+	for i := uint64(0); i < req.Count; i++ {
+		if i > 0 {
+			block = backend.Chain().GetBlockByHash(block.ParentHash())
+			if block == nil {
+				break
+			}
+		}
+		enc, err := rlp.EncodeToBytes(NewBlockData(block))
+		if err != nil {
+			log.Error("failed to encode BlockData", "hash", block.Hash(), "err", err)
```
flywukong, line 203: Behavior change: if encoding fails on the start block (i==0), we silently
reply with an empty Blocks list instead of an error. The caller may treat
this as "peer has no data" rather than a server-side failure. Consider
returning an error when len(blocks)==0 so the requester can retry another peer.

flywukong (dismissed):

eth/protocols/bsc/handler.go
```diff
@@ -177,21 +188,41 @@ func handleGetBlocksByRange(backend Backend, msg Decoder, peer *Peer) error {
...
+			if block == nil {
+				break // chain ends or pruned mid-walk; return what we have
+			}
+		}
+		enc, err := rlp.EncodeToBytes(NewBlockData(block))
+		if err != nil {
+			// Server-side failure. If we have nothing to send, surface the error so
+			// the requester doesn't mistake it for "peer has no data".
+			log.Error("failed to encode BlockData", "hash", block.Hash(), "err", err)
+			if len(blocks) == 0 {
```
NathanBSC: can't return error, otherwise connect will broken

eth/protocols/bsc/handler.go
```diff
@@ -177,21 +188,41 @@ func handleGetBlocksByRange(backend Backend, msg Decoder, peer *Peer) error {
 		block = backend.Chain().GetBlockByNumber(req.StartBlockHeight)
 	}
 	if block == nil {
-		return fmt.Errorf("msg %v, cannot get start block: %v, %v", GetBlocksByRangeMsg, req.StartBlockHeight, req.StartBlockHash)
```
NathanBSC, line 185: revert

eth/protocols/bsc/handler.go
zlacfzy: Encode failed should not return error, to disconnect peer

NathanBSC (approved):

allformless (approved):

flywukong (approved):

PR #3678, opened 2026-05-11
miner: remove optional transaction gas limit cap

allformless:
### Description

miner: remove optional transaction gas limit cap

### Rationale

EIP-7825 has been enabled on bsc,
so remove optional transaction gas limit cap,
along side remove useless 'SetMaxGas'

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/api_miner.go
+++ b/eth/api_miner.go
@@ -20,8 +20,6 @@ import (
 	"math/big"
 	"time"
 
-	"github.com/ethereum/go-ethereum/params"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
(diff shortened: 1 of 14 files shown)


zlacfzy (approved):

eth/api_miner.go
```diff
@@ -70,15 +68,6 @@ func (api *MinerAPI) SetGasPrice(gasPrice hexutil.Big) bool {
 	return true
 }
 
-// SetGasLimit sets the gaslimit to target towards during mining.
```
flywukong, line 73: this is not introduced  by gas limit cap,  but it should be  fine to be removed

flywukong (approved):

PR #3684, opened 2026-05-12
cmd, core, eth, tests: prevent state flushing in RPC (#33931)

allformless:
### Description

cmd, core, eth, tests: prevent state flushing in RPC (#33931)

### Rationale

upstream commit: https://github.com/ethereum/go-ethereum/commit/6d99759f0

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/eth/api_debug.go
+++ b/eth/api_debug.go
@@ -520,27 +520,3 @@ func (api *DebugAPI) ExecutionWitness(bn rpc.BlockNumberOrHash) (*stateless.ExtW
 
 	return result.Witness().ToExtWitness(), nil
 }
-
-func (api *DebugAPI) ExecutionWitnessByHash(hash common.Hash) (*stateless.ExtWitness, error) {
-	bc := api.eth.blockchain
-	block := bc.GetBlockByHash(hash)
-	if block == nil {
-		return &stateless.ExtWitness{}, fmt.Errorf("block hash %x not found", hash)
-	}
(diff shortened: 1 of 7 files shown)


zlacfzy (dismissed):

zlacfzy (dismissed):

eth/api_debug.go
```diff
@@ -496,39 +497,23 @@ func (api *DebugAPI) StateSize(blockHashOrNumber *rpc.BlockNumberOrHash) (interf
 	}, nil
 }
 
-func (api *DebugAPI) ExecutionWitness(bn rpc.BlockNumber) (*stateless.ExtWitness, error) {
+func (api *DebugAPI) ExecutionWitness(bn rpc.BlockNumberOrHash) (*stateless.ExtWitness, error) {
 	bc := api.eth.blockchain
-	block, err := api.eth.APIBackend.BlockByNumber(context.Background(), bn)
+	block, err := api.eth.APIBackend.BlockByNumberOrHash(context.Background(), bn)
```
flywukong, line 502: `BlockByNumberOrHash` may return `(nil, nil)` for an unknown block hash, so this can panic on `block.ParentHash()` before returning a not-found error. Please add an explicit `block == nil` check after the call.

zlacfzy (approved):

flywukong (approved):

PR #3725, opened 2026-06-16
feat: implement BEP-696 compatibility 

flywukong:
### Description

Implement [BEP-696](https://github.com/bnb-chain/BEPs/blob/master/BEPs/BEP-696.md) compatibility for repurposing ParentBeaconRoot and UncleHash as optional producer metadata after Pasteur.

Allow post-Pasteur ParentBeaconRoot to carry non-zero metadata while still requiring it to be present.
Skip BSC EIP-4788 beacon-root system calls because ParentBeaconRoot is not a consensus-layer beacon root.
Allow post-Pasteur UncleHash to carry producer metadata instead of committing to the uncle body.
Keep actual block bodies uncle-free across downloader, fetcher, broadcast, and import validation paths.
Preserve pre-Pasteur validation semantics for both fields.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -64,9 +64,8 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if err := v.bc.engine.VerifyUncles(v.bc, block); err != nil {
 		return err
 	}
-	// From Pasteur (BEP-696), UncleHash is repurposed to carry producer metadata and no
-	// longer commits to the uncle list. VerifyUncles above already guarantees the body
-	// carries no uncles, so the uncle-root match is only enforced before Pasteur.
+	// After Pasteur, UncleHash no longer commits to the body; VerifyUncles still
+	// guarantees the body has no uncles.
 	if !v.config.IsPasteur(block.Number(), block.Time()) {
 		if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
 			return fmt.Errorf("uncle root hash mismatch (header value %x, calculated %x)", header.UncleHash, hash)
--- a/core/block_validator.go
+++ b/core/block_validator.go
@@ -64,8 +64,13 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if err := v.bc.engine.VerifyUncles(v.bc, block); err != nil {
 		return err
 	}
-	if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
-		return fmt.Errorf("uncle root hash mismatch (header value %x, calculated %x)", header.UncleHash, hash)
+	// From Pasteur (BEP-696), UncleHash is repurposed to carry producer metadata and no
+	// longer commits to the uncle list. VerifyUncles above already guarantees the body
+	// carries no uncles, so the uncle-root match is only enforced before Pasteur.
+	if !v.config.IsPasteur(block.Number(), block.Time()) {
+		if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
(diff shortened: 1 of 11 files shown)


core/block_validator.go
```diff
@@ -64,8 +64,12 @@ func (v *BlockValidator) ValidateBody(block *types.Block) error {
 	if err := v.bc.engine.VerifyUncles(v.bc, block); err != nil {
 		return err
 	}
-	if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
-		return fmt.Errorf("uncle root hash mismatch (header value %x, calculated %x)", header.UncleHash, hash)
+	// After Pasteur, UncleHash no longer commits to the body; VerifyUncles still
+	// guarantees the body has no uncles.
+	if !v.config.IsPasteur(block.Number(), block.Time()) {
```
allformless, line 69: len(uncles) ==0 has been ensured by `VerifyUncles`

for bsc, no need to `CalcUncleHash`, so I prefer
```
if v.config.IsNotInBSC()) {
	if hash := types.CalcUncleHash(block.Uncles()); hash != header.UncleHash {
			return fmt.Errorf("uncle root hash mismatch (header value %x, calculated %x)", header.UncleHash, hash)
		}
	}
```

eth/protocols/eth/handlers.go
```diff
@@ -387,9 +387,18 @@ func handleNewBlock(backend Backend, msg Decoder, peer *Peer) error {
 		return err
 	}
 
-	if hash := types.CalcUncleHash(ann.Block.Uncles()); hash != ann.Block.UncleHash() {
-		log.Warn("Propagated block has invalid uncles", "have", hash, "exp", ann.Block.UncleHash())
-		return nil // TODO(karalabe): return error eventually, but wait a few releases
+	// After Pasteur, UncleHash is metadata; block bodies must still have no
+	// uncles. Before Pasteur, it remains the uncle-list commitment.
+	if backend.Chain().Config().IsPasteur(ann.Block.Number(), ann.Block.Time()) {
```
allformless, line 392: handleNewBlockethereum has no func `handleNewBlock`
so just need to consider in bsc senario, IMO, just
```
		if len(ann.Block.Uncles()) > 0 {
			log.Warn("Propagated block has uncles after Pasteur", "count", len(ann.Block.Uncles()))
			return nil // TODO(karalabe): return error eventually, but wait a few releases
		}
```
no need to check Pasteur, and no need to `CalcUncleHash`

eth/fetcher/block_fetcher.go
```diff
@@ -749,7 +753,13 @@ func (f *BlockFetcher) loop() {
 						if uncleHash == (common.Hash{}) {
 							uncleHash = types.CalcUncleHash(task.uncles[i])
 						}
```
allformless, line 755: blockfetcher only exist in bsc now
so just need to consider in bsc senario, IMO, just
```
		if len(task.uncles[i]) > 0 {
			continue
		}
```
no need to check Pasteur, and no need to `CalcUncleHash`

zlacfzy (approved):

eth/downloader/queue.go
```diff
@@ -802,8 +806,16 @@ func (q *queue) DeliverBodies(id string, hashes eth.BlockBodyHashes, bodies []et
 		if hashes.TransactionRoots[index] != header.TxHash {
 			return errInvalidBody
 		}
-		if hashes.UncleHashes[index] != header.UncleHash {
```
allformless, line 805: prefer
```
if hashes.UncleHashes[index] != types.EmptyUncleHash{
       if hashes.UncleHashes[index] != header.UncleHash {
              return errInvalidBody
       }
}
```
and check 
len(uncles)==0 in handleBlockBodies

PR #3726, opened 2026-06-16
core/vm: suspend legacy v1 Tendermint precompiles, bound v2 light block

zlacfzy:
### Description

The legacy v1 Tendermint light-client precompiles 0x64 (tmHeaderValidate) and 0x65 (iavlMerkleProofValidate) are only reachable via the deprecated BC<->BSC cross-chain stack (contracts/deprecated/), which no longer exists. Suspend them from the Pasteur fork via a new PrecompiledContractsPasteur set (the Osaka set with 0x64/0x65 mapped to the existing *Nano "suspend" variants). The live BSC<->Greenfield path uses 0x67 and 0x66, unaffected.

For the v2 cometBFT light-client (0x67, still live), bound the light-block validator set to the same maximum as the consensus state (maxValidators=99; the block's set becomes the next consensus state) and cap the light-block blob (maxLightBlockLength=32KB). The bound is fork-gated: 0x67 maps to a new cometBFTLightBlockValidatePasteur variant only in the Pasteur set, which uses DecodeLightBlockValidationInputBounded; the pre-Pasteur Hertz variant keeps the unbounded decode so earlier blocks still replay identically.

Adds regression tests (including that the unbounded decode does not apply the new bound). Existing v2 light-block tests still pass; a 99-validator light block encodes to ~17KB, well under the cap.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -112,8 +112,8 @@ var PrecompiledContractsNano = PrecompiledContracts{
 	common.BytesToAddress([]byte{0x8}): &bn256PairingIstanbul{},
 	common.BytesToAddress([]byte{0x9}): &blake2F{},
 
-	common.BytesToAddress([]byte{0x64}): &tmHeaderValidateNano{},
-	common.BytesToAddress([]byte{0x65}): &iavlMerkleProofValidateNano{},
+	common.BytesToAddress([]byte{0x64}): &tmHeaderValidateDeprecated{},
+	common.BytesToAddress([]byte{0x65}): &iavlMerkleProofValidateDeprecated{},
 }
 
 var PrecompiledContractsMoran = PrecompiledContracts{
@@ -371,8 +371,8 @@ var PrecompiledContractsPasteur = PrecompiledContracts{
 	common.BytesToAddress([]byte{0x10}): &bls12381MapG1{},
 	common.BytesToAddress([]byte{0x11}): &bls12381MapG2{},
 
-	common.BytesToAddress([]byte{0x64}): &tmHeaderValidate{},
-	common.BytesToAddress([]byte{0x65}): &iavlMerkleProofValidatePlato{},
+	common.BytesToAddress([]byte{0x64}): &tmHeaderValidateDeprecated{},
+	common.BytesToAddress([]byte{0x65}): &iavlMerkleProofValidateDeprecated{},
 	common.BytesToAddress([]byte{0x66}): &blsSignatureVerify{},
 	common.BytesToAddress([]byte{0x67}): &cometBFTLightBlockValidatePasteur{},
 	common.BytesToAddress([]byte{0x68}): &verifyDoubleSignEvidence{},
--- a/core/vm/contracts_lightclient.go
+++ b/core/vm/contracts_lightclient.go
@@ -134,33 +134,34 @@ func (c *iavlMerkleProofValidate) Name() string {
 	return "IAVL_MERKLE_PROOF_VALIDATE"
 }
 
-// tmHeaderValidateNano implemented as a native contract.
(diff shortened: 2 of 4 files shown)


core/vm/contracts.go
```diff
@@ -350,6 +350,38 @@ var PrecompiledContractsOsaka = PrecompiledContracts{
...
+	common.BytesToAddress([]byte{0x0b}): &bls12381G1Add{},
+	common.BytesToAddress([]byte{0x0c}): &bls12381G1MultiExp{},
+	common.BytesToAddress([]byte{0x0d}): &bls12381G2Add{},
+	common.BytesToAddress([]byte{0x0e}): &bls12381G2MultiExp{},
+	common.BytesToAddress([]byte{0x0f}): &bls12381Pairing{},
+	common.BytesToAddress([]byte{0x10}): &bls12381MapG1{},
+	common.BytesToAddress([]byte{0x11}): &bls12381MapG2{},
+
+	common.BytesToAddress([]byte{0x64}): &tmHeaderValidateNano{},
+	common.BytesToAddress([]byte{0x65}): &iavlMerkleProofValidateNano{},
```
allformless: tmHeaderValidateNano --> tmHeaderValidateDeprecated

```
func (c *tmHeaderValidateNano) Run(input []byte) (result []byte, err error) {
	return nil, errors.New("suspend")
}
```
-->
```
func (c *tmHeaderValidateNano) Run(input []byte) (result []byte, err error) {
	return nil, errors.New("deprecated")
}
```

core/vm/lightclient/v2/lightclient.go
```diff
@@ -27,6 +27,12 @@ const (
 	relayerBlsKeyLength        uint64 = 48
 	singleValidatorBytesLength uint64 = validatorPubkeyLength + validatorVotingPowerLength + relayerAddressLength + relayerBlsKeyLength
 	maxConsensusStateLength    uint64 = chainIDLength + heightLength + validatorSetHashLength + 99*singleValidatorBytesLength // Maximum validator quantity 99
+
+	// maxValidators caps the light block's validator set; it becomes the next
+	// consensus state, which is itself capped at 99.
+	maxValidators uint64 = 99
```
unclezoro: seems a big small. can't validator number impact the final gas usage?

core/vm/lightclient/v2/lightclient.go
```diff
@@ -27,6 +27,12 @@ const (
...
+	// maxLightBlockLength caps the light-block blob (a 99-validator block is well under it).
+	maxLightBlockLength uint64 = 32 * 1024
```
unclezoro: seems not a good idea to have a hard limitation here.  The idea is not to set limitation but to define a reasonable gas usage according to the real CPU/Memory comsumption

allformless (approved):

flywukong (approved):

core/vm/contracts_lightclient.go
```diff
@@ -134,33 +134,34 @@ func (c *iavlMerkleProofValidate) Name() string {
...
+// legacy v1 Tendermint header-validate precompile (returns an error for any input).
+type tmHeaderValidateDeprecated struct{}
 
-func (c *tmHeaderValidateNano) RequiredGas(input []byte) uint64 {
+func (c *tmHeaderValidateDeprecated) RequiredGas(input []byte) uint64 {
 	return params.TendermintHeaderValidateGas
 }
 
-func (c *tmHeaderValidateNano) Run(input []byte) (result []byte, err error) {
-	return nil, errors.New("suspend")
```
unclezoro, line 145: will this  change impact the consensus result?

PR #3732, opened 2026-06-18
consensus/parlia: reject zero turnLength in parseTurnLength (fix divide-by-zero DoS)

iakisme:
# consensus/parlia: reject zero turnLength in parseTurnLength (defense-in-depth vs divide-by-zero)

## Summary

`parseTurnLength` (`consensus/parlia/snapshot.go:564`) reads the turn-length byte directly from header extra-data and returns it **without any range check**. The value flows into `Snapshot.TurnLength` via `apply()`, and the subsequent `inturnValidator` / `backOffTime` computations perform `… / uint64(TurnLength)` — so a `0` here is an **integer divide-by-zero panic**.

The byte is *not* the on-chain value. The `BSCValidatorSet` contract hard-bounds `turnLength` to `{1} ∪ [3,64]` (and `getTurnLength()` maps a stored `0 → 1`), so an **honest** producer — which copies `getTurnLength()` in `prepareTurnLength` — never writes `0`. The risk is a **Byzantine validator** that crafts and signs an epoch header with the byte set to `0`, bypassing the contract entirely. The only guard that cross-checks the header byte against the on-chain value, `verifyTurnLength`, runs in `Finalize` (`parlia.go:1407`) and is **not on the header-only verification / header-sync path**, where the snapshot is built and the division is reached first.

## Threat model & impact

- Requires the crafting node to be a **current, staked validator** — the epoch header must survive `apply()`'s `ecrecover` + validator-membership check. Not triggerable by an arbitrary network peer.
- The malicious header is **signed → self-attributable → slashable**, so the attack is high-cost and low-stealth.
- Within that model, the panic propagates to every peer that verifies the header (the `VerifyHeaders` workers have no `recover()`), so a single Byzantine validator can halt verifying peers.

Reachable division path (re-verified against current code):

```
VerifyHeaders → verifyCascadingFields
  ├─ snapshot(...)                          parlia.go:710  → apply() sets snap.TurnLength = parsed byte (0)
  └─ blockTimeVerifyForRamanujanFork(...)   parlia.go:715  → backOffTime(...)
        → inturnValidator()  snapshot.go:469:  (s.Number+1) / uint64(s.TurnLength) % …   ← ÷0 panic
```

This is reached *before* `verifySeal`, and `Finalize` (hence `verifyTurnLength`) never runs on the header-sync path. `loadSnapshot` coerces a persisted `0` back to the default (`snapshot.go:130`), which masks the issue across restarts but does not prevent the in-memory panic during verification.

- **Severity:** Medium (Byzantine-validator–gated DoS; not arbitrary-network-triggerable)
- **Class:** CWE-369 (Divide By Zero), reachable from a validator-signed header
- **Affected division sites:** `snapshot.go:458/469/486/492`, `parlia.go:2409`

## Fix

This is **defense-in-depth**: validate the untrusted header byte at parse time rather than relying solely on the `Finalize`-time `verifyTurnLength`. Reject `turnLength == 0` in `parseTurnLength`, returning the existing `errInvalidTurnLength`.

```go
turnLength := header.Extra[pos]
if turnLength == 0 {
    return nil, errInvalidTurnLength
}
return &turnLength, nil
```

Note: this rejects only `0` — enough to remove the panic. The parse path still accepts out-of-contract values such as `2` or `200`, which `verifyTurnLength` rejects later in `Finalize`. Tightening `parseTurnLength` to the full `{1} ∪ [3,64]` set (so parse-time and `Finalize` agree) is a reasonable follow-up but is intentionally out of scope here to keep the change minimal.

## Tests

- `TestParseTurnLengthRejectsZero` — **fails before the fix** (parseTurnLength returns `(0, nil)`), passes after.
- `TestParseTurnLengthAcceptsValid` — guards that valid turn-lengths still parse.
- `TestSnapshotZeroTurnLengthPanics` — documents the divide-by-zero impact at the snapshot level.

```
$ go test ./consensus/parlia/ -run 'TestParseTurnLength|TestSnapshotZeroTurnLength'
ok  github.com/ethereum/go-ethereum/consensus/parlia
```

Full `consensus/parlia` package passes (no regression).

## Notes / follow-ups (not in this PR)

- Validate the full `{1} ∪ [3,64]` range in `parseTurnLength` to mirror the contract and `verifyTurnLength`.
- `bohrFork.go:36` truncates the contract-returned turn length to `uint8` (256→0), feeding the same divide-by-zero family — worth validating `IsInt64()` and range before truncation.
- Consider clamping `uint64(TurnLength)` to `max(1, …)` at the division sites as further defense-in-depth.

### DIFF
--- a/consensus/parlia/snapshot.go
+++ b/consensus/parlia/snapshot.go
@@ -576,6 +576,14 @@ func parseTurnLength(header *types.Header, chainConfig *params.ChainConfig, epoc
 		return nil, errInvalidTurnLength
 	}
 	turnLength := header.Extra[pos]
+	// A zero turn-length is never valid and, if accepted, would propagate into
+	// Snapshot.TurnLength and cause an integer divide-by-zero panic in
+	// inturnValidator/backOffTime. This check must live here (the header
+	// verification path) because verifyTurnLength only runs in Finalize and is
+	// bypassed by header-only sync.
+	if turnLength == 0 {
+		return nil, errInvalidTurnLength
+	}
 	return &turnLength, nil
 }
 
--- a/consensus/parlia/snapshot_test.go
+++ b/consensus/parlia/snapshot_test.go
@@ -2,12 +2,15 @@ package parlia
 
 import (
 	"bytes"
+	"math/big"
 	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/params"
 )
 
 func TestValidatorSetSort(t *testing.T) {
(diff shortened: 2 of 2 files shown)


consensus/parlia/snapshot_test.go
```diff
@@ -21,3 +24,73 @@ func TestValidatorSetSort(t *testing.T) {
...
+	cfg := bohrChainConfig()
+	header := buildTurnLengthHeader(defaultEpochLength, 1, 0)
+
+	_, err := parseTurnLength(header, cfg, defaultEpochLength)
+	assert.ErrorIs(t, err, errInvalidTurnLength,
+		"parseTurnLength must reject a zero turn-length from attacker-controlled extra-data")
+}
+
+// TestParseTurnLengthAcceptsValid ensures the fix does not break valid parsing.
+func TestParseTurnLengthAcceptsValid(t *testing.T) {
```
zlacfzy: Can just keep TestParseTurnLengthRejectsZero.

zlacfzy:
CI need fix

consensus/parlia/snapshot.go
```diff
@@ -576,6 +576,14 @@ func parseTurnLength(header *types.Header, chainConfig *params.ChainConfig, epoc
 		return nil, errInvalidTurnLength
 	}
 	turnLength := header.Extra[pos]
+	// A zero turn-length is never valid and, if accepted, would propagate into
```
NathanBSC, line 579: impossible
https://github.com/bnb-chain/bsc-genesis-contract/blob/master/contracts/BSCValidatorSet.sol#L678

PR #2770, opened 2024-11-25
p2p/discover: fix ENR filtering

MatusKysel:
### Description

Reintroducing ENR filtering back. This allows discv4 ENR filtering and add small support for discv5 that could be enabled later on based on testing.

PR #1320, opened 2023-02-23
p2p/discover: implement ENR node filtering

MatusKysel:
### Description

This PR implements [BEP194](https://github.com/bnb-chain/BEPs/pull/194)

### Rationale

This change should improve peer discovery time by filtering out peers from different networks e.g. ETH

### Changes

Notable changes: 
* adding peer do the table is now async as it also requires to get ENR record from the node
* this could also help to filter out nodes, that are not running latest HF

PR #3735, opened 2026-06-24
p2p: remove BSC table-layer ENR node filtering and dead bootnode bucket-size code

zlacfzy:
### Description

Remove BSC-specific discovery-table customizations that are either redundant or dead code.

## 1. Table-layer ENR node filtering (#1320, #2770, #3237)

Remove the discovery-table ENR node filtering feature introduced and evolved across [#1320](https://github.com/bnb-chain/bsc/pull/1320), [#2770](https://github.com/bnb-chain/bsc/pull/2770) and [#3237](https://github.com/bnb-chain/bsc/pull/3237).

**Rationale**: the dial-layer forkid filter in `eth/backend.go` `setupDiscovery` (`RandomNodes()` → `enode.AsyncFilter(RequestENR)` → `enode.Filter(eth.NewNodeFilter)` → `NewBufferIter` → discmix, upstream go-ethereum #31944) already provides the same network separation for full nodes, runs by default, and is more efficient. The table-layer filter is gated behind `EnableENRFilter`, which has no CLI flag and is never set true, so on default full nodes it never runs.

Removed: `Table.enrFilter`/`filterNode` and its checks in `table_reval.go`/`v5_udp.go`; `discover.NodeFilterFunc`/`ParseEthFilter`/`GetEthEntry`/`Config.FilterFunction`; `server.go` `EnableENRFilter` branch + `forkFilter`/`SetFilter`; `p2p.Config.EnableENRFilter`; `eth.StartENRFilter`; bootnode `-network` flag and its `ParseEthFilter`/`GetEthEntry` use.

Also reverts the #2770 asynchronous `handleAddNode` scaffolding back to the upstream synchronous form (it only existed to accommodate the blocking `RequestENR` inside the now-removed `filterNode`): drop `addNodeOp.syncExecution`/`addInboundNodeSync`, move `tab.mutex` back to the callers, restore `addInboundNode` to return bool.

**Kept**: the dial-layer `AsyncFilter` pipeline + `eth.NewNodeFilter`, the `RequestENR` transport method, and the [#3347](https://github.com/bnb-chain/bsc/pull/3347) `AsyncFilter` shutdown-deadlock fix.

## 2. Dead bootnode bucket-size code (#1727)

[#1727](https://github.com/bnb-chain/bsc/pull/1727) made bootnodes use a 256-entry bucket (`bootNodeBucketSize`) via a per-`Table` `bucketSize` field set when `IsBootnode` is true. The #2770 refactor rewrote `handleAddNode` to use the `bucketSize` constant (16) directly, so `Table.bucketSize` has been written but never read since — the larger bucket never actually took effect.

Verified on a **live official mainnet bootnode** (`cmd/bootnode`, `IsBootnode=true`) via dlv: `tab.bucketSize=256` but the busiest buckets cap at 16 entries, confirming the 256 bucket is inert. Removing the dead field/constant therefore has zero runtime impact. `Config.IsBootnode` is kept (still gates the bootnode "Discovery status" log).

### Verification

- `go build ./...`
- `go test -race ./p2p/discover/`
- `go test ./p2p/`


allformless (dismissed):

cmd/bootnode/main.go
```diff
@@ -129,17 +121,11 @@ func main() {
...
-	ethEntry, err := discover.GetEthEntry(*networkFilter)
-	if err != nil {
-		utils.Fatalf("-network: %v", err)
-	}
-	ln.Set(ethEntry)
 	printNotice(&nodeKey.PublicKey, *listenerAddr)
 	cfg := discover.Config{
-		PrivateKey:     nodeKey,
-		NetRestrict:    restrictList,
-		FilterFunction: filterFunction,
```
galaio, line 141: Was removing the bootnode-side ENR filtering intended, or is it collateral from the full-node cleanup?

The rationale ("gated behind `EnableENRFilter`, which has no CLI flag and is never set true") only holds for the **full-node** path (`p2p/server.go` -> `Config.EnableENRFilter`). The **bootnode** sets `Table.enrFilter` through a *different* path: `-network bsc` -> `discover.ParseEthFilter` -> `Config.FilterFunction`, which bypasses `EnableENRFilter` entirely. So on a bootnode launched with `-network bsc`, `filterNode` **is** active today.

Bootnodes also don't have the dial-layer fallback we rely on for full nodes: `cmd/bootnode` calls `discover.ListenUDP/ListenV5` directly and never builds the `AsyncFilter -> eth.NewNodeFilter` pipeline (that lives only in `eth/backend.go setupDiscovery`). A bootnode doesn't dial; it only answers FINDNODE from its DHT table. So `-network bsc`'s `filterNode` is the *only* thing keeping non-BSC (e.g. ETH mainnet) discv4 nodes out of the served table. Removing it means the bootnode table gets polluted with cross-network nodes, and bootstrapping BSC nodes get FINDNODE responses mixed with non-BSC peers -- slower discovery, which is the regression #1320/#2770/#3237 originally addressed.

Two asks before this lands:
1. **Can you confirm the live mainnet bootnode's actual launch args / `tab.enrFilter`?** The dlv check verified `bucketSize=256` was inert but didn't report whether `enrFilter` is nil. If prod runs `-network bsc`, then `enrFilter != nil` and this is a real behavioral removal, not dead-code cleanup.
2. If dropping bootnode filtering **is** intended, please make it explicit rather than implicit: update `README.md:256` (still documents `bootnode ... -network bsc`) and add a short comment / PR-description note stating that bootnodes will now accept non-BSC nodes and why that's acceptable. If it's **not** intended, consider keeping the bootnode `-network` / `FilterFunction` / `ln.Set(ethEntry)` path and scoping this PR to only the full-node `EnableENRFilter` branch + the dead `bootNodeBucketSize`.

zlacfzy: Confirmed it's intended, not collateral — prod bootnode runs -network bsc, so filterNode is active there and bootnodes have no dial-layer fallback.

1. The table-layer filter isn't needed. Every BSC full node already runs EnableENRFilter=false — the whole full-node population has always relied solely on the dial-layer forkid filter (eth.NewNodeFilter) and the network runs fine. Correctness doesn't depend on a bootnode's table purity; dropping it only costs slightly slower cold-start discovery, which is transient. This just makes bootnodes consistent with full nodes.
2. Rollout is gradual. Bootnodes keep running the current -network bsc binary and migrate over time.

galaio (dismissed):

flywukong (dismissed):

hilljessica8787-dot (approved):

joey0612 (approved):

allformless (approved):

PR #3737, opened 2026-06-25
feat: stop sending the bsc capability handshake message

zlacfzy:
### Description

Follow-up to #3483. That PR removed the **blocking** part of the bsc protocol handshake but still **sent** a `BscCapMsg` (and ignored inbound ones) for backward compatibility, noting it would "enable complete handshake removal in a future release". This PR is that next step: it stops sending `BscCapMsg` entirely.

### Rationale

The bsc subprotocol and its version are already negotiated by the devp2p (RLPx `Hello`) capability exchange before any bsc message is processed — `peer.version` is set at that point. The bsc-level `BscCapMsg` only re-confirmed that already-negotiated version, so it was redundant. #3483 removed the blocking read; the outbound send was kept only to satisfy peers still blocking on the handshake.

This also aligns `bsc` with `snap`, the other `eth` satellite protocol: `snap` has never had a subprotocol-level handshake — its `RunPeer` goes straight to `runSnapExtension`. After this PR, `bsc`'s `RunPeer` has the same shape. A satellite protocol carries no identity/version info that devp2p + the `eth` `StatusMsg` handshake don't already establish, so a dedicated `bsc` handshake was never necessary.

### Changes

- `eth/handler_bsc.go`: `RunPeer` no longer calls `SendBscCap()` (now mirrors `snapHandler.RunPeer`).
- `eth/protocols/bsc/handshake.go`: removed (it only contained `SendBscCap`).
- `eth/protocols/bsc/protocol.go`: removed `BscCapPacket` and `defaultExtra`; `BscCapMsg = 0x00` kept reserved (no longer sent).
- `eth/protocols/bsc/handler.go`: `handleBscCap` simplified to a no-op (payload is drained by `handleMessage`'s deferred `msg.Discard()`).

### Backward compatibility

`handleBscCap` is **kept registered** for code `0x00`. Peers still running the previous release continue to send `BscCapMsg` on connect; silently ignoring it prevents them from being torn down via `errInvalidMsgCode`. Wire codes (`VotesMsg=0x01`, `GetBlocksByRange=0x02`, `BlocksByRange=0x03`) and `protocolLengths` are unchanged.

Once the whole network runs a release that no longer sends `BscCapMsg`, the `0x00` code and `handleBscCap` can be removed entirely in a future PR.


NathanBSC:
does the reth client has stopped waiting the handshake msg？
we can only remove this safely after that IMO

will-2012:
> does the reth client has stopped waiting the handshake msg？ we can only remove this safely after that IMO

Yep, `reth-bsc` currently relies on this message; otherwise, the handshake process will be affected. `reth-bsc` needs to implement compatibility logic and be deployed before `geth-bsc` to avoid compatibility issues.

hilljessica8787-dot (approved):

hilljessica8787-dot (approved):

NathanBSC:
reth develop ready for this PR
but we should make sure https://github.com/bnb-chain/reth-bsc/pull/474 is included in a hardfork release
and the hardfork passed
then this PR can be merged

joey0612 (approved):

PR #3739, opened 2026-06-29
core/types: extract bid and block mev info into builder subpackage

flywukong:
### Description

refactor bid/bidblocks go file path , move to builder package
 support builder integration doc  to support bep-675 builder integration

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/core/types/builder/bid.go
+++ b/core/types/builder/bid.go
@@ -1,4 +1,4 @@
-package types
+package builder
 
 import (
 	"fmt"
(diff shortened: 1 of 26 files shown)


aweneagle (approved):

andyzhang2023 (approved):

allformless (approved):

core/types/builder/bid.go
```diff
@@ -1,4 +1,4 @@
-package types
+package builder
```
zlacfzy, line 1: Just use `buildertypes` ?

zlacfzy (approved):

PR #3763, opened 2026-07-17
cmd/geth: fix database command help typos

futurehua:
### Description

Fixes a few typos and grammar issues in `geth db` command help text.

This only updates CLI descriptions and does not change command behavior.

### Rationale

### Example

### Changes

Notable changes: 
* ...


hilljessica8787-dot (approved):

zlacfzy:
Thanks! Since the inspect text comes from upstream go-ethereum, fixing it only here adds a cosmetic diff we'd carry through every merge — for a docs-only change I'd rather send that upstream so it stays in sync, so I'll pass on merging this as-is. 🙏

PR #3774, opened 2026-07-31
 internal/ethapi: support gRPC sendBidBlock API

flywukong:
### Description
This PR adds an optional BEP-675 BidBlock gRPC endpoint through`BidBlockService.SendBidBlock`.

The gRPC endpoint transports the `BidBlock` as RLP-encoded bytes, with the signature and validator hostname carried as protobuf fields. This avoids JSON parsing and hex expansion for large blocks and blob sidecars.

After transport decoding, both gRPC and JSON-RPC ingress reuse the existing `MevSentry.sendBidBlock` business path, including signature recovery, builder allowlist validation, validator routing, timeout handling, and forwarding.

The existing JSON-RPC endpoint remains fully supported and protocol-compatible. Both ingress paths share the process-wide concurrency budge
```
gRPC BidBlockService
   ├─ pre-decode admission / timeout
   ├─ protobuf unmarshal
   └─ MevGRPCServer.SendBidBlock
      ├─ RLP decode BidBlock
      └─ MevAPI.SendBidBlock(ctx, BidBlockArgs)
                 │
                 ▼
       EthAPIBackend.SendBidBlock
```
### Rationale

### Example

### Changes

Notable changes: 
* ...


allformless (dismissed):

allformless (approved):

internal/ethapi/mev_grpc_service.go
```diff
@@ -0,0 +1,209 @@
...
+
+	healthServer.Shutdown()
+	gracefulDone := make(chan struct{})
+	go func() {
+		server.GracefulStop()
+		close(gracefulDone)
+	}()
+	select {
+	case <-gracefulDone:
+	case <-time.After(mevGRPCShutdownTimeout):
```
zlacfzy, line 154: 15s  maybe too long?

zlacfzy (approved):

internal/ethapi/mev_grpc_service.go
flywukong: > 15s maybe too long?

15s is only the upper bound; shutdown returns immediately when no RPC is in flight. The default request timeout is 10s, so the additional 5s leaves enough time for an admitted request to observe cancellation and finish cleanup.

PR #3792, opened 2026-08-13
core, params: implement BEP-706 millisecond timestamp precompile

joey0612:
### Description

This PR implements [BEP-706](https://github.com/bnb-chain/BEPs/pull/706) behind a new BSC-only hardfork named **Jenner** (activation time not scheduled yet, `jennerTime` is nil for both mainnet and Chapel).

It introduces a new precompiled contract at address `0x70` that returns the **millisecond-precision timestamp of the executing block** (`Header.MilliTimestamp()`, available at consensus level since BEP-520) as a left-padded 32-byte big-endian integer. The precompile costs a fixed **20 gas** and ignores calldata entirely. Before the fork activates, calling `0x70` behaves exactly like calling an empty account (success, empty return data) — this falls out of the fork-gated precompile map machinery, no special-case code.

Since the existing `PrecompiledContract` interface is stateless (`Run(input []byte)`), this PR adds an optional `BlockContextPrecompiledContract` interface and threads `vm.BlockContext` through `RunPrecompiledContract`. The ~30 existing precompiles are untouched.

**RPC block-override semantics on this client (needs a BEP-706 supplement):** the `prevRandao` block-override field maps to the header's `MixDigest`, which BSC repurposes as the **millisecond remainder of the block timestamp** (BEP-520) — consensus enforces a value below 1000 on real headers and never exposes it to the EVM as randomness. Block overrides in `eth_call` / `eth_estimateGas` / `debug_traceCall` / `eth_simulateV1` follow the same semantics: the value is validated (`< 1000`, rejected otherwise — callers must NOT pass arbitrary random values) and is assembled into the millisecond timestamp reported by the precompile: `milliTimestamp = time*1000 + prevRandao`. This behavior should be added to the BEP-706 spec as a follow-up.

### Rationale

BEP-520 lowered BSC block times below one second: the sub-second remainder lives in `Header.MixDigest` and is validated by consensus, but EVM contracts can only read whole seconds via the `TIMESTAMP` opcode. DEXs, oracles, auctions and games need the on-chain millisecond value. BEP-706 exposes the already-consensus-validated value through a precompile without changing `block.timestamp` semantics.

The fork is gated BSC-only: `IsJenner()` embeds an explicit `IsInBSC()` (Parlia) check, so a non-Parlia config can never activate the precompile even if `jennerTime` is set. Jenner is defined to always be the newest fork; `CheckConfigForkOrder` rejects any `jennerTime` earlier than another scheduled fork (equal is allowed).

### Example

```bash
# start a node with a manual activation time (both `geth init` and runtime paths supported)
geth --override.jenner 1800000000 ...
```

```bash
# read the executing block's millisecond timestamp
curl -s -X POST -H 'Content-Type: application/json' --data \
  '{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[{"to":"0x0000000000000000000000000000000000000070"},"latest"]}' \
  http://127.0.0.1:8545
# → {"result":"0x000000000000000000000000000000000000000000000000000001a3185c5000"}   (= 1800000000000, block time 1800000000.000s)

# simulate a future time with a 123ms remainder (prevRandao carries the millisecond remainder on BSC)
curl -s ... '"params":[{"to":"0x…0070"},"latest",null,{"time":"0x713fb300","prevRandao":"0x…007b"}]'
# → 1900000000123

# invalid remainder is rejected
curl -s ... '"params":[…,{"prevRandao":"0x…03e8"}]'
# → error: block override "prevRandao" on BSC carries the millisecond remainder of the block
#   timestamp (BEP-520/BEP-706) and must be less than 1000, got 1000
```

```solidity
(bool ok, bytes memory data) = address(0x70).staticcall("");
uint256 milliTimestamp = abi.decode(data, (uint256)); // e.g. 1800000000123
```

### Changes

Notable changes:
* `params`: define the Jenner fork — `JennerTime` on `ChainConfig` (nil placeholders for mainnet/Chapel/Rialto), `IsJenner()` with an explicit BSC-only gate, `Rules`, `LatestFork`, `Timestamp()`, `checkCompatible`, and hard fork-ordering enforcement


params/forks/forks.go
```diff
@@ -52,6 +52,7 @@ const (
 	BPO4
 	BPO5
 	Amsterdam
+	Jenner
```
allformless: move definition next to Pasteur？

params/config.go
```diff
@@ -1848,6 +1877,10 @@ func (c *ChainConfig) LatestFork(time uint64) forks.Fork {
 	london := c.LondonBlock
 
 	switch {
+	// Jenner is defined to always be the latest BSC fork (enforced by the
+	// fork ordering check in CheckConfigForkOrder), so it stays on top.
+	case c.IsJenner(london, time):
```
allformless: move to above 'case c.IsPasteur(london, time)'?

params/config.go
```diff
@@ -1583,6 +1605,10 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 		{name: "bpo4", timestamp: c.BPO4Time, optional: true},
 		{name: "bpo5", timestamp: c.BPO5Time, optional: true},
 		{name: "amsterdam", timestamp: c.AmsterdamTime, optional: true},
+		// Jenner must always be the latest fork: keeping it last in this list
+		// makes the ordering check reject any JennerTime earlier than another
+		// defined fork (equal is allowed).
+		{name: "jenner", timestamp: c.JennerTime, optional: true},
```
allformless: move next to '{name: "pasteurTime", timestamp: c.PasteurTime}'?

core/vm/contract_milli_timestamp.go
```diff
@@ -0,0 +1,54 @@
...
+// the BLS precompiles). Do NOT copy that pattern for milliTimestamp: call
+// RunWithBlockContext instead. Returning a loud, explicit error here is
+// deliberate: it fails safely (no state change, no silently-wrong value)
+// rather than fabricating a plausible-looking but fake timestamp.
+func (c *milliTimestamp) Run(input []byte) ([]byte, error) {
+	return nil, errors.New("milliTimestamp: must be dispatched with block context, direct Run() is not supported")
+}
+
+func (c *milliTimestamp) Name() string {
+	return "milliTimestamp"
```
allformless: upper format like others?

eth/catalyst/simulated_beacon.go
```diff
@@ -103,6 +103,14 @@ type SimulatedBeacon struct {
 
 func payloadVersion(config *params.ChainConfig, time uint64) engine.PayloadVersion {
 	switch config.LatestFork(time) {
+	case forks.Jenner:
```
allformless: move into line 116?

eth/catalyst/simulated_beacon.go
```diff
@@ -209,7 +217,9 @@ func (c *SimulatedBeacon) sealBlock(withdrawals []*types.Withdrawal, timestamp u
 		Random:                random,
 		BeaconRoot:            &common.Hash{},
 	}
-	if c.eth.BlockChain().Config().LatestFork(timestamp) == forks.Amsterdam {
```
allformless, line 212: no need to change

params/config.go
```diff
@@ -1937,6 +1970,11 @@ func (c *ChainConfig) ActiveSystemContracts(time uint64) map[string]common.Addre
 // the fork isn't defined or isn't a time-based fork.
 func (c *ChainConfig) Timestamp(fork forks.Fork) *uint64 {
 	switch {
+	// Keep Jenner first: it is defined to be the latest fork, and
+	// logForkReadiness in core/blockchain.go relies on Timestamp(current+1)
+	// resolving for the fork enum value right after the current one.
+	case fork == forks.Jenner:
```
allformless: also change if change order in 'forks.go'

params/forks/forks.go
joey0612: fixed

params/config.go
joey0612: fixed

params/config.go
joey0612: fixed

core/vm/contract_milli_timestamp.go
joey0612: fixed

eth/catalyst/simulated_beacon.go
joey0612: fixed

eth/catalyst/simulated_beacon.go
joey0612: fixed

params/config.go
joey0612: fixed

allformless (dismissed):

params/config.go
```diff
@@ -1576,6 +1598,10 @@ func (c *ChainConfig) CheckConfigForkOrder() error {
 		{name: "osakaTime", timestamp: c.OsakaTime},
 		{name: "mendelTime", timestamp: c.MendelTime},
 		{name: "pasteurTime", timestamp: c.PasteurTime},
+		// Jenner (BEP-706) is the BSC fork right after Pasteur; keeping it here
+		// makes the ordering check require JennerTime to be at or after Pasteur
+		// and at or before the later (BPO/Amsterdam) forks.
+		{name: "jenner", timestamp: c.JennerTime, optional: true},
```
zlacfzy: jennerTime？

params/config.go
```diff
@@ -1949,6 +1982,11 @@ func (c *ChainConfig) Timestamp(fork forks.Fork) *uint64 {
 		return c.BPO2Time
 	case fork == forks.BPO1:
 		return c.BPO1Time
+	// Jenner sits right after Pasteur in the fork order (see forks.go);
+	// logForkReadiness relies on Timestamp(current+1) resolving for the fork
+	// enum value right after the current one, so keep this order aligned.
```
will-2012: The "ponytail skill" allows for some comments to be simplified.

will-2012 (dismissed):

internal/ethapi/simulate.go
```diff
@@ -301,6 +301,12 @@ func (sim *simulator) processBlock(ctx context.Context, block *simBlock, header,
 		header.ExcessBlobGas = &excess
 	}
 	blockContext := core.NewEVMBlockContext(header, sim.newSimulatedChainContext(ctx, headers), nil)
+	// The simulated header's MixDigest follows the BSC header semantics
+	// (BEP-520): it is the millisecond remainder of the block timestamp —
+	// either zero or a prevRandao override validated to be < 1000 in
+	// sanitizeChain. Header.MilliTimestamp() therefore assembles the exact
+	// millisecond timestamp the BEP-706 precompile must report.
+	blockContext.MilliTimestamp = header.MilliTimestamp()
```
zlacfzy: core.NewEVMBlockContext already sets MilliTimestamp

internal/ethapi/override/override.go
zlacfzy, line 266: Please keep MakeHeader on the same rule as Apply: when Time is overridden and PrevRandao is nil, zero MixDigest.

core/vm/contract_milli_timestamp.go
```diff
@@ -0,0 +1,54 @@
...
+// RunWithBlockContext is the actual entry point, dispatched from
+// RunPrecompiledContract when the block context carrying the millisecond
+// timestamp is available. It falls back to Time*1000 if MilliTimestamp is
+// zero: some non-live-path BlockContext constructors (core/vm/runtime,
+// evm t8n, various tests) don't fill the new field, and a second-precision
+// value padded with .000 is "degraded but correct", unlike a near-1970
+// garbage value.
+func (c *milliTimestamp) RunWithBlockContext(blockCtx BlockContext, input []byte) ([]byte, error) {
+	ts := blockCtx.MilliTimestamp
+	if ts == 0 {
```
zlacfzy, line 50: ts can't be 0

internal/ethapi/override/override.go
```diff
@@ -127,19 +127,51 @@ func (diff *StateOverride) Apply(statedb *state.StateDB, precompiles vm.Precompi
...
+// of a randomness beacon. Consensus enforces the same bound on real headers
+// (MilliTimestamp()/1000 must equal Time).
+const MaxBSCMilliRemainder = 1000
+
+// BSCMilliRemainder interprets a "prevRandao" override the BSC way: the
+// 32-byte value is the millisecond remainder and must be below
+// MaxBSCMilliRemainder, exactly like the MixDigest of a real BSC header.
+func BSCMilliRemainder(prevRandao *common.Hash) (uint64, error) {
+	ms := new(big.Int).SetBytes(prevRandao[:])
+	if ms.Cmp(big.NewInt(MaxBSCMilliRemainder)) >= 0 {
```
zlacfzy: In parlia, MilliTimestamp() only cmp uint64, may keep same?

flywukong (dismissed):

internal/ethapi/override/override.go
```diff
@@ -166,6 +203,14 @@ func (o *BlockOverrides) Apply(blockCtx *vm.BlockContext) error {
...
 	if o.PrevRandao != nil {
+		// prevRandao is the millisecond remainder on BSC — validate it and
+		// assemble it into the millisecond timestamp on top of the (possibly
+		// overridden) seconds.
+		ms, err := BSCMilliRemainder(o.PrevRandao)
+		if err != nil {
+			return err
+		}
+		blockCtx.MilliTimestamp = blockCtx.Time*1000 + ms
 		blockCtx.Random = o.PrevRandao
```
zlacfzy, line 222: In bsc blockCtx.Random should always be nil, but now PrevRandao will not be nil.

core/vm/contract_milli_timestamp.go
joey0612: Yes, it is a safety check, I think we can keep the code.

internal/ethapi/override/override.go
joey0612: Setting it here is intentional: it keeps the Apply path (eth_call / eth_estimateGas / debug_traceCall) consistent with eth_simulateV1. In some simulated scenarios, users need to know the exact millisecond timestamp because some time-sensitive contracts may behave differently.

params/config.go
joey0612: fixed

params/config.go
joey0612: Changed

internal/ethapi/simulate.go
joey0612: fixed

internal/ethapi/override/override.go
joey0612: fixed

internal/ethapi/override/override.go
joey0612: fixed

zlacfzy (approved):

will-2012 (approved):

PR #3795, opened 2026-08-17
eth/protocols/eth: disconnect on NewBlock body-hash mismatch

joey0612:
### Description

`handleNewBlock` re-derives a propagated block's body commitments (`CalcUncleHash`, `DeriveSha` over every transaction) to check them against the header. When they don't match it only logged a `Warn` and returned `nil` — leaving a `TODO(karalabe): return error eventually` — so the connection stayed up and the peer could send the same malformed block again.

This PR resolves the TODO: a body/header-commitment mismatch now returns an error, which `handleMessage` turns into a disconnect. A malformed block can never enter the chain, so forcing the recipient to keep re-hashing it is pure wasted work; the sender is treated as a bad peer.

### Rationale

Without a disconnect, a peer can repeatedly send blocks (up to `maxMessageSize`, ~10MiB) whose `TxHash`/`UncleHash` don't match the body, making the victim pay `Decode` + `DeriveSha` on every message with no throttle and no penalty — a cheap, repeatable CPU drain against any inbound-accepting node (public full nodes, sentries, RPC backends). The sibling paths in the same function (`Decode`, `sanityCheck` failures) already disconnect; the mismatch path was the odd one out.

Upstream go-ethereum resolved this TODO by removing the whole `NewBlock` broadcast path post-merge (PoS propagates blocks via the consensus layer). BSC (Parlia PoSA) still relies on eth `NewBlock` gossip to propagate blocks, so the handler must stay — the fix is to make the mismatch an explicit disconnect signal instead. A single `handleNewBlock` is shared by both the `eth68` and `eth70` dispatch tables, so one change covers every eth version BSC negotiates.

### Example

NA

### Changes

Notable changes: 
* `eth/protocols/eth/handlers.go`: in `handleNewBlock`, return an `errDecode`-wrapped error (instead of `nil`) on `UncleHash` and `TxHash` mismatch, so the connection is disconnected; resolves the `TODO(karalabe)`. Added a comment explaining why BSC keeps the handler (Parlia gossip) rather than removing the path like upstream.
* `eth/protocols/eth/handler_test.go`: add `TestHandleNewBlockBadBody`, asserting both mismatch cases (TxHash / UncleHash) return an `errDecode`-wrapped error.


zlacfzy (dismissed):

will-2012 (dismissed):

eth/protocols/eth/handlers.go
```diff
@@ -397,13 +397,24 @@ func handleNewBlock(backend Backend, msg Decoder, peer *Peer) error {
...
+	// mismatch must be an explicit disconnect signal.
 	if hash := types.CalcUncleHash(ann.Block.Uncles()); hash != ann.Block.UncleHash() {
 		log.Warn("Propagated block has invalid uncles", "have", hash, "exp", ann.Block.UncleHash())
-		return nil // TODO(karalabe): return error eventually, but wait a few releases
+		return fmt.Errorf("%w: propagated block has invalid uncle hash: have %x, exp %x", errDecode, hash, ann.Block.UncleHash())
 	}
 	if hash := types.DeriveSha(ann.Block.Transactions(), trie.NewStackTrie(nil)); hash != ann.Block.TxHash() {
 		log.Warn("Propagated block has invalid body", "have", hash, "exp", ann.Block.TxHash())
-		return nil // TODO(karalabe): return error eventually, but wait a few releases
+		return fmt.Errorf("%w: propagated block has invalid body: have %x, exp %x", errDecode, hash, ann.Block.TxHash())
```
flywukong, line 417: fetchRangeBlocks (eth/handler.go:326) doesn't verify TxHash/UncleHash, and the fetcher propagates before insertChain validates the body — so a node with EnableQuickBlockFetching could relay a body-mismatched block and get disconnected here for someone else's fault. Most sentry nodes has enable EnableQuickBlockFetching = true and may meet this issue

joey0612: Confirmed the path and fixed it. fetchRangeBlocks only SanityChecks (no TxHash/UncleHash) and the fetcher broadcasts with propagate=true once verifyHeader passes, ahead of InsertChain — so an honest relay with EnableQuickBlockFetching (default on sentries) can forward a body-mismatched block it got from a bad upstream, and disconnecting on receipt would misattribute the fault and could cascade through a relay topology.

Dropped the disconnect: on a mismatch the block is now discarded before it reaches backend.Handle — so this node never re-broadcasts it and never pays the downstream insert — while the connection stays up, with a new eth/propagated/block/malformed meter as the signal. That still fixes the original issue (the old code logged and then forwarded the block anyway); the only thing given up is punishing a genuinely malicious peer, and the residual cost is one body-hash derivation per message, which is unavoidable since that is what reveals the mismatch. Tests updated: malformed blocks are asserted dropped-and-still-connected, plus a positive control that a well-formed block is still forwarded.

PR #3796, opened 2026-08-17
miner: persist BidBlock revokes across restarts

joey0612:
### Description

BidBlock `SendBidBlock` revokes (BEP-675 Layer 2) were held only in an in-memory map in `BidBlockPermissionManager`, reconstructed empty on every `miner.New`. A validator that revoked a builder for 24h and then restarted within that window (upgrade, crash, OOM) came back with an empty map, so the builder was immediately allowed to `SendBidBlock` again — even though the permission RPC still advertised `resetAt = revokedAt + duration` as a wall-clock promise.

This PR persists the revoke map to the chain database and restores it on startup:

* The whole map is mirrored to a single database key (`bidBlockRevokes`), overwritten on every change. It is JSON-encoded — RLP is geth's codec for consensus-critical / large / frequent data, but supports neither maps nor `time.Time`; this state is tiny, written rarely, never on-chain, and benefits from JSON's schema tolerance and readability, matching the existing `WriteChainConfig` precedent.
* On startup the manager loads the persisted map and keeps only revokes still within their window, dropping any whose lockout elapsed while the node was offline. So a revoke survives a restart within its window; one that fully expired during downtime is not resurrected.
* Persistence is **asynchronous and best-effort**: mutations snapshot the map under the lock and hand it to a fire-and-forget writer, so the mining path never blocks on disk I/O. Losing the newest write on a crash is acceptable; what must not happen is a disk hiccup stalling `Revoke`/`SetAllowed`.
* Writers are serialized and carry a monotonic sequence number. The seq guard is advanced **before** the write and **regardless of its outcome**, so the seq that actually reaches `db.Put` is strictly increasing: once a newer snapshot has entered the writer, an older one is dropped even if the newer write failed — an older snapshot can never overwrite a newer one and resurrect stale state (e.g. a revoke overwriting a later manual clear).

The manager keeps a nil-database path (unchanged in-memory behavior) for callers and tests that don't wire a database.

### Rationale

A revoke that silently vanishes on restart lets a misbehaving builder resume `SendBidBlock` (bypassing the weaker `SendBid` fallback) before its lockout should have ended, and makes the RPC's `resetAt` a promise the node doesn't keep. Durability of the lockout across restarts is what the advertised 24h window requires.

Scope is deliberately limited to durability across restarts: revoke triggers, the 24h / 450s durations, the RPC semantics and the per-validator (non-shared) nature of the lockout are all unchanged.

### Example

NA

### Changes

Notable changes:
NA

### DIFF
--- a/miner/bid_block_permission.go
+++ b/miner/bid_block_permission.go
@@ -7,12 +7,12 @@ package miner
 
 import (
 	"encoding/json"
+	"os"
 	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	buildertypes "github.com/ethereum/go-ethereum/core/types/builder"
-	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
 )
 
@@ -28,18 +28,20 @@ const (
 	bidBlockGasPriceLowRevokeDuration = 450 * time.Second
 )
 
-// bidBlockRevokesKey is the database key under which the whole revoke map is
-// persisted (a single blob, overwritten on every change) so that lockouts
-// survive a process restart within their window.
+// The revoke map is persisted as a single JSON file under the node's datadir
+// (a whole-map snapshot, atomically replaced on every change) so that lockouts
+// survive a process restart within their window. It deliberately lives OUTSIDE
+// chaindata: revokes are validator-local MEV policy, not chain data — they must
+// not be wiped by a chaindata resync nor carried along when chaindata is cloned.
+// This mirrors how other node-local state is stored (vote journal, txpool
+// journal).
 //
-// The blob is JSON-encoded rather than RLP. RLP is geth's codec for
+// The file is JSON-encoded rather than RLP. RLP is geth's codec for
 // consensus-critical and large/frequent data, but it supports neither maps nor
 // time.Time, so it would require flattening the map and converting timestamps
 // through a separate wire struct. This state is tiny, written rarely (only on
 // bad bids), never on-chain, and benefits from JSON's forgiving schema
-// evolution and readability — matching the existing rawdb precedent for small
-// metadata blobs (e.g. WriteChainConfig).
-var bidBlockRevokesKey = []byte("bidBlockRevokes")
+// evolution and readability.
 
 // BidBlockRevokeRecord holds one active revoke event.
 type BidBlockRevokeRecord struct {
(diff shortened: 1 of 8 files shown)


miner/bid_block_permission.go
```diff
@@ -35,19 +51,119 @@ type BidBlockRevokeRecord struct {
...
+//
+// persistedSeq is advanced BEFORE the write and regardless of its outcome. This
+// makes the seq that actually reaches db.Put strictly monotonic: once a newer
+// snapshot has entered here, every older one is dropped even if the newer
+// write failed. Otherwise a newer write failing could let an older snapshot
+// win the disk afterwards and resurrect stale state (e.g. a revoke overwriting
+// a later manual clear). Losing the newest write on failure is acceptable
+// (best-effort); an older snapshot overwriting a newer one is not.
+func (m *BidBlockPermissionManager) persistAsync(seq uint64, snapshot map[common.Address]BidBlockRevokeRecord) {
+	m.persistMu.Lock()
```
flywukong, line 153: Consider following the VoteJournal approach by persisting the JSON under a dedicated <datadir>/bidBlockRevokeJournal/ directory instead of chaindata, since BidBlock revocations are validator-local MEV policy rather than chain data: https://github.com/bnb-chain/bsc/blob/master/core/vote/vote_journal.go.

joey0612: Agreed — moved it out of chaindata. Revokes now journal to a single atomic JSON file under the datadir (<datadir>/bidblockrevokes.json, path resolved via stack.ResolvePath like the txpool journal), and the ChainDb() addition to miner.Backend is gone. I kept it a snapshot file rather than a VoteJournal-style WAL dir: the WAL fits append-heavy per-block votes, while this is a rarely-mutated small map with self-expiring (≤24h) entries — a whole-map atomic replace is simpler and crash-safe (temp file + rename). Same restore-on-restart semantics and ordering guarantees, now with file-based tests.

will-2012 (dismissed):

miner/bid_block_permission.go
```diff
@@ -35,19 +53,128 @@ type BidBlockRevokeRecord struct {
...
-	return &BidBlockPermissionManager{
-		revoked: make(map[common.Address]BidBlockRevokeRecord),
-		clock:   time.Now,
+// NewBidBlockPermissionManager returns a manager journalling to journalPath
+// (may be empty for a purely in-memory manager). Any revokes persisted by a
+// previous process that are still within their lockout window are restored;
+// expired ones are dropped.
+func NewBidBlockPermissionManager(journalPath string) *BidBlockPermissionManager {
+	m := &BidBlockPermissionManager{
+		revoked:     make(map[common.Address]BidBlockRevokeRecord),
```
flywukong, line 113: Consider adding explicit json: tags and a version envelope to the journal.

. The journal  may change format (may need to carry a per-builder revoke count as well to support new feature like dynamic reovke time computiing) , so the lockout duration can scale with the number of offenses — adding fields for that is then a clean v1→v2 migration rather than a compat branch.

joey0612: fixed

flywukong (approved):

will-2012 (approved):

PR #3799, opened 2026-08-20
core/vm: CAS20 stateful precompile host

zlacfzy:
### Description

Implements [BEP-702](https://github.com/bnb-chain/BEPs/blob/new_token/BEPs/BEP-702.md) — a native token standard whose tokens are stateful precompiles rather than deployed contracts.

Two parts:

1. **A stateful-precompile host** in `core/vm`. `StatefulPrecompiledContract` receives a `PrecompileContext` carrying the StateDB, the caller, the call-mode guards and dynamic gas metering, so a precompile can read and write consensus state. Additive: `RunPrecompiledContract` and every existing precompile are untouched.
2. **The token family** built on it — a factory, two variants (Asset and Stablecoin), a shared PolicyRegistry and an ActivationRegistry. A token holds no executable bytecode, only a one-byte `0xEF` sentinel; it is identified by its address, routed by the address prefix, and isolated by the state stored under that address.

Nothing is reachable until governance opens it. Routing and the fork hook both require Jenner **and** a configured activation admin, and every built-in config still names a placeholder — so on `develop` today this changes no behaviour and writes no state.

### Rationale

A BEP-20 transfer spends most of its gas on bytecode interpretation rather than on the state it touches. Moving the standard token into the client removes that overhead, makes one audited implementation serve every token of a variant, and lets compliance state (allow/blocklists) be shared across tokens instead of reimplemented per issuer.

The design follows Base's B20, whose reference implementation [base-std](https://github.com/base/base-std) is public and MIT. Aligning rather than inventing means integrators and auditors can reuse what they know, and it gives conformance something external to check against.

### Example

Creation goes through the factory at `0x20bF000000000000000000000000000000000000`. The
bootstrap window lets `initCalls` grant roles and mint before any role holder exists,
so a token can be created and initialised in one transaction:

```
createB20(uint8 variant, bytes32 salt, bytes params, bytes[] initCalls) -> address
```

The returned address is an ERC-20 from then on — `transfer`, `approve`, `balanceOf`
and the rest answer at it, with no code deployed there:

```bash
# 1 for the sentinel, and the code hash is keccak256(0xEF)
$ geth attach --exec 'eth.getCode("0x20B0...")'
"0xef"
```

A network that wants the switch held by a key it controls names it at startup:

```bash
geth --override.jenner <timestamp> --override.b20activationadmin 0x…
```

### Changes

Notable changes:

* `core/vm/contracts_stateful.go` — the host: `StatefulPrecompiledContract`, `PrecompileContext`, and the dispatcher `evm.runPrecompile` routes to. Existing precompiles keep the stateless path.
* `core/vm/b20*.go` — the factory, both variants, the two registries, RBAC, pause, supply cap, transfer policies, permit (EIP-2612) and the Asset extensions (multiplier, batch mint, announcements).
* Gas is produced by existing EVM cost functions applied to the work performed — EIP-2929 warm/cold, EIP-2200 net metering with EIP-3529 refunds, log and keccak prices — so a token operation is never cheaper than the same state accesses through bytecode. No new gas constant is introduced.
* Storage uses [ERC-7201](https://eips.ethereum.org/EIPS/eip-7201) namespaces with Solidity-identical slot derivation, specified in BEP-702 §3.17 and generated from `core/vm/testdata/b20_layout.json`, which a test holds against the constants.
* `params`, `core/genesis.go`, `core/systemcontracts/upgrade.go` — the activation admin is chain configuration, seeded once at the Jenner boundary and part of that block's state root, so it participates in `CheckCompatible`.
* `core/vm/testdata/basestd_surface.json` — base-std's published ABI at a pinned commit. `TestB20SurfaceMatchesBaseStd` diffs our whole registered surface against it in both directions; divergences must be listed with a written reason.

Not in this PR: the `B20` → `N20` rename that BEP-702 now uses, the real activation multisig, and a dedicated fork flag (Jenner is the gate today).


core/vm/b20.go
```diff
@@ -0,0 +1,274 @@
+// Copyright 2024 The go-ethereum Authors
+// This file is part of the go-ethereum library.
```
will-2012: The B20 implement is not part of the go-ethereum library?

core/vm/b20.go
```diff
@@ -0,0 +1,274 @@
...
+	if !ctx.chargeCodeWrite(ctx.Self, b20MarkerCode) {
+		return
+	}
+	ctx.StateDB.SetCode(ctx.Self, b20MarkerCode, tracing.CodeChangeContractCreation)
+}
+
+// b20EnterCall applies the guards every B20 entry point shares: direct calls
+// only, nonpayable, and calldata charged once (BEP-702 3.14). The value check
+// precedes the charge, so a value-bearing call consumes no gas.
+func b20EnterCall(ctx *PrecompileContext, input []byte) error {
```
will-2012: b20EnterCheck?

core/vm/evm.go
```diff
@@ -44,8 +44,44 @@ type (
...
-	p, ok := evm.precompiles[addr]
-	return p, ok
+	if p, ok := evm.precompiles[addr]; ok {
+		return p, ok
+	}
+	// B20 native token family: the factory has a fixed address and every token
+	// address is resolved dynamically from its prefix (there is no fixed-address
+	// entry for tokens, so they cannot live in the static precompiles map).
+	if evm.b20Enabled() {
+		return resolveB20(addr)
```
joey0612: **The reserved address space is only recognised on the CALL path — consider a CREATE-side guard, and an operator tool to verify the space is clean before activation.**

This is the only site in the tree that recognises the reserved space: `IsB20Address` / `resolveB20` are referenced from `precompile()` and nowhere else. `evm.create` derives an address, runs the standard collision check (`nonce != 0 || non-empty code`) and has no notion of the B20 prefix, so a CREATE/CREATE2 landing inside `0x20B0…` is not refused as such.

To be clear about the severity: **I don't think this is a reachable vulnerability today.** Squatting a *specific* B20 address needs a full 20-byte CREATE2 preimage (2^160). Hitting *some* address in the space only needs the 80-bit prefix, but token addresses derive from `keccak(abi.encode(creator, salt))[:9]`, so an attacker cannot choose which one they land on — the payoff is an address nobody was going to use. And even if code did end up there, `precompile()` shadows it (calls route to the B20 handler, the deployed code is unreachable) and `b20AddressOccupied` refuses `createB20` on any code, so the failure mode is fail-safe rather than a takeover.

So the suggestion is defence-in-depth, not a fix. Two reasons I still think it earns its place:

1. **It makes the invariant enforced rather than derived.** "The reserved space is owned exclusively by the factory" currently holds only if you chain together the address derivation, 2^160, the `precompile()` shadowing and `b20AddressOccupied`'s fail-safe. One explicit check makes it readable.
2. **It protects the invariant against future change.** Today's safety rests on the prefix width and the derivation. Anyone who later shortens the prefix, adds a variant with a different derivation, or widens `IsB20Address` silently weakens the cryptographic argument; an explicit check does not.

The check itself is small, and `createB20` writes the sentinel via `ctx.StateDB.SetCode` rather than `evm.create`, so the factory is not caught by it:

```go
// in evm.create, before the collision check
if evm.b20Enabled() && IsB20Address(address) {
    gas.Exhaust()
    return nil, common.Address{}, gas, ErrB20ReservedAddress
}
```

It is consensus-affecting, so the gate has to be exactly `evm.b20Enabled()` (`IsJenner && B20Scheduled()`) — same predicate as this line — or nodes disagree.

joey0612: **Separately, and I think more important than the guard: could we get an operator tool that audits the reserved space between the fork and activation?**

A CREATE guard cannot answer the question the tool would, for two reasons: it only constrains CREATEs *after* the fork, and the cheap way to dirty the space needs no CREATE at all. Anyone can send BNB to a `0x20B0…` address today — no key, no compute — producing an account with a balance, no code and no nonce. After the fork `b20AddressOccupied` only inspects code (`hadNoCode`), so that address is *not* occupied, `createB20` succeeds there, and the new token inherits the balance. B20 never touches native balance, so the BNB is inert but permanently unrecoverable — there is no withdrawal path.

Between the fork block and the first `activate()` there is a natural window to check this once, and right now there is no tooling for it (the only B20-related CLI surface is `--override.b20activationadmin`). What would help:

- accounts under `0x20B0…` (and the two registry singletons) carrying **code** — those addresses can never host a token
- accounts carrying a **balance** — a token created there would strand it
- accounts with a **non-zero nonce**
- confirmation that both registry sentinels and the activation admin were actually seeded at the boundary

The last two also cover a related gap: `SeedB20Activation` has a single production caller (`core/systemcontracts/upgrade.go`) gated on `IsOnJenner`, which only fires on the transition block — so on a chain whose genesis is already past Jenner (`jennerTime: 0`) nothing is ever seeded, the activation admin stays zero, and `requireAdmin` then rejects everyone, leaving no way to open a feature. A tool that asserts "sentinels present, admin non-zero" would catch that before it matters.

A `geth` subcommand iterating the state for the prefix would be enough; it does not need to be online. Happy to help scope it if useful.

One related nit while we're here: `b20AddressOccupied` only checks code, while `evm.create`'s collision check is `nonce != 0 || non-empty code`. Aligning the two would be more consistent. I'd deliberately *not* treat a balance as "occupied" though — otherwise anyone could permanently block a derivable address for 1 wei, which trades one problem for a cheaper DoS.

params/config.go
```diff
@@ -733,7 +736,19 @@ type ChainConfig struct {
...
+	// at the fork that ships the precompiles (BEP-702 3.15). It is the only
+	// account that can ever open a B20 feature.
+	//
+	// The choice is one-shot. Seeding runs on the fork boundary block and writes
+	// the admin only into an empty slot; setAdmin then requires the caller to be
+	// the current admin. So an admin that cannot originate a call — the zero
+	// address, or a contract with no path to the registry — can never be replaced
+	// by any transaction, and only a further hard fork can install one. Verify the
+	// configured account can actually reach the registry before the fork ships.
+	B20ActivationAdmin *common.Address `json:"b20ActivationAdmin,omitempty"`
```
joey0612: Should the activation switch be a config-supplied key at all, rather than an existing system authority? or the admin adress is managed by GOV?

core/vm/b20_activation.go
```diff
@@ -0,0 +1,240 @@
...
+func (r activationReg) setAdmin(a common.Address) {
+	r.s.setWord(actSlot(actSlotAdmin), addrKey(a))
+}
+
+// requireAdmin reverts unless the caller is the (non-zero) activation admin.
+// Both clauses are needed and neither is about the caller being zero: a zero
+// caller with an admin set fails the second, and one on a registry with no admin
+// fails the first — which is the clause that stops zero from matching zero
+// (BEP-702 3.15).
+func (r activationReg) requireAdmin(ctx *PrecompileContext) error {
```
joey0612: **The activation authority is state that has to be seeded, and there is a case where it never is.**

`requireAdmin` reads the admin out of a storage slot, so the authority only exists once something writes it. That write is `SeedB20Activation`, whose single production caller is gated on `IsOnJenner` — true only on the block that crosses the fork boundary. On a chain whose genesis is already past Jenner (`jennerTime: 0`) no such block exists, so the seeding never runs and the slot stays zero.

`b20Enabled()` is `IsJenner && B20Scheduled()` and does not consult the registry, so the reserved space is routed regardless. The result is a chain where `0x20B0…` is taken over by B20 and no feature can ever be opened, since `requireAdmin` rejects everyone against a zero admin. The tests already sidestep this by calling `SeedB20Activation` directly instead of going through the fork hook (`core/vm/b20_policy_test.go`).

Separately, the field's own doc comment notes the choice is one-shot: seeding only writes into an empty slot and `setAdmin` requires the caller to already be the admin, so a misconfigured admin *"can never be replaced by any transaction, and only a further hard fork can install one."*

**Possible fix: make the authority a constant instead of state** — check the caller against an existing system authority, e.g. `onlyGov` → `GovHubContract` (`0x…1007`), or the governor/timelock path (`0x…2006`):

```go
if ctx.Caller != systemcontracts.GovHubContract {
    return revB20("Unauthorized(address)", errSelUnauthorizedAddr, addrKey(ctx.Caller))
}
```

Then nothing needs seeding: the admin slot goes away along with `B20ActivationAdmin`, its placeholder and `setAdmin`, and the sentinels do not need seeding either since `activate` already plants one lazily (`ctx.ensureSentinel()`, *"only the activating write creates state worth keeping"*) and the registry holds no storage before that. `SeedB20Activation` and its fork hook can go entirely, which removes the gap above by construction rather than by fix.

joey0612:
**Test coverage: a few whole features are untested rather than thinly tested**

First, credit where it's due — the suite is substantial and several parts are better than typical: dispatch goes through the real `evm.Call` path rather than hand-built contexts, the storage layout is pinned against a fixture with swapped-key counter-assertions (`TestB20PolicyMemberSlot`), the string encoding has a full directed boundary matrix with a stale-slot scan, and `TestB20GasNeverCheaperThanBytecode` pins the property BEP-702 3.14 actually cares about. The gaps below are not "this is tested shallowly"; they are features with no behavioural test at all, where the suite stays green if the implementation is removed.

**Tier 1 — a deletion that the suite would not notice**

1. **The scheduled multiplier (ERC-8056) has no behavioural test.** `updateUIMultiplier` is exercised only for its value ceiling; `settleMatured`, `effectiveMultiplier`, `cancelUIMultiplierUpdate`, `newUIMultiplier`, `effectiveAt`, `totalSupplyUI`, `balanceOfUI`, `toUIAmount`, `fromUIAmount` and the two events are never called or asserted. Nothing schedules an update and advances `block.timestamp` past `effectiveAt`. Since maturation is deliberately silent — no transaction, no event (`b20_asset.go:128-136`) — a regression here means the revaluation simply never happens, with every node still agreeing with itself.

2. **`settleMatured` specifically.** The comment at `b20_asset.go:157-164` says dropping it "would silently revert the token to the value from before it matured". Removing the call at `:507` leaves the suite green and quietly revalues every holder on the next scheduled update.

3. **Composite policies (UNION / INTERSECT) are functionally untested.** `createCompositePolicy`, `updateComposite`, `validateChildren`, `children`/`setChildren`, `compositePolicyChildIds`, the empty-INTERSECT rule and the no-nesting rule that bounds `isAuthorized`'s recursion — grep finds them only as selector hashes in `b20_abi_test.go`. Nothing ever constructs one. `setChildren`'s tail-clearing loop exists to keep the state root matching a reference `uint64[]` layout, and nothing observes the slots it writes.

4. **The EIP-2929 warm/cold policy is untested, and `b20_gas.go:31-34` points at a file that does not exist** (`b20_warming_test.go`). None of BEP-702 3.17's clauses are covered: no B20 address starting warm, transaction access lists applying to B20 slots, warmth keyed on `(address, slot)`, or a foreign-slot read warming the foreign address. B20 addresses are absent from the static precompile map today, so a future refactor that adds them would make every one start warm and silently reprice every call.

5. **The BURN and SEIZE pause gates never block anything.** `b20PauseBurn` / `b20PauseSeize` appear in tests only as values passed to `pause()` and read back from `pausedFeatures()`. Delete the `isPaused` check in `burn` (`b20_admin.go:540`), `seizeWithMemo` (`:575`) or `batchMint` (`b20_asset.go:582`) and everything passes — a paused token keeps burning and seizing.

6. **The three gates the bootstrap must never skip are unasserted.** No `initCalls` bundle in any test contains `selUpdatePolicy`, `selPause` or `selUpdateSupplyCap`, so the claim at `b20_admin.go:511-517` ("enforced even during privileged bootstrap") is unbacked. Adding `!t.privileged` to `mintCore`'s policy check would let `createB20` mint past a compliance list and past the supply cap in the same transaction that sets them.

7. **`ctx.adminRenounced` has zero test references.** Drop `|| t.ctx.adminRenounced` from `ensureRoleMutable` and a bootstrap bundle can call `renounceLastAdmin()` — emitting `LastAdminRenounced`, which is advertised as permanent — and then grant itself the role back.

8. **The three activation feature ids are pinned to nothing.** Every test refers to them by Go variable; neither the canonical strings nor their keccak values are asserted anywhere. Renaming `"bsc.b20_asset"` — which `b20_activation.go:37-43` explicitly flags as confusable with the `"bsc.b20.asset"` storage namespace — changes a consensus-visible `bytes32`, breaks every integrator's `activate()`, and silently reverts already-activated features to inactive.

9. **`burn` has no unauthorized-caller test**; every burn in the suite is by a `BURN_ROLE` holder, so removing `ensureRole(roleBurn)` is green.

10. **The ERC-20 zero-address boundary set is entirely absent** — transfer to zero, `transferFrom` from zero, `approve` to zero, `approve` by zero. None of `errSelInvalidReceiver` / `InvalidSender` / `InvalidSpender` / `InvalidApprover` appears in any test, so `move()`'s zero checks are reached by nothing.

11. **`applyMultiplier` / `removeMultiplier` overflow is untested.** With a multiplier near `type(uint128).max`, `scaledBalanceOf` and `totalSupplyUI` must revert `Panic(0x11)` rather than wrap; nothing approaches the boundary.

12. **The pending-multiplier slot's lane packing is unpinned.** `setPending` / `pending` round-trip through each other, so swapping the mul/effectiveAt bit positions is invisible to any behavioural test while changing consensus-visible storage — exactly the hazard `TestB20PolicyLanePositions` was written for on the policy side.

13. **Nothing asserts that a failing `initCall` reverts the whole creation** — that the sentinel is absent afterwards, that `InitCallFailed(uint256)` carries the right index, or that a sub-4-byte entry yields `InternalCallMalformed(bytes)`. A half-created token would be indistinguishable from success.

14. **`CheckCompatible` on `B20ActivationAdmin` after Jenner is untested** (`params/config.go:1887`); `TestCheckCompatible_Jenner` covers only the timestamp. The admin is part of the boundary block's state root, so this check is what stops an operator editing a stored config and forking themselves off.

**Tier 2 — user-visible, lower blast radius** (happy to expand any of these): three of the five memo entry points (`transferFromWithMemo`, `mintWithMemo`, `burnWithMemo`) are never dispatched; seize's mandated `Transfer → Memo → Seized` ordering is unasserted, as is `Announcement` / `EndAnnouncement` being emitted at all; `supportsInterface` and the five advertised interface ids are never called or derived; permit's zero-spender branch, low-s / v-range enforcement, and "a rename invalidates an outstanding signature" (the separator is shown to change, but no test signs, renames, and submits); `setRoleAdmin` / `getRoleAdmin` are never invoked through the ABI although `ensureRoleMutable` routes all authority through `roleAdmin(role)`; infinite allowance never being decremented (`maxU256` appears in no test); `batchMint`'s empty batch and mid-batch failure; `SeedB20Activation`'s idempotency guard; and the initial `supplyCap` never being asserted after creation.

**There is no end-to-end test.** Every B20 test tops out at `evm.Call` inside `core/vm`; nothing constructs a real transaction, goes through `ApplyTransaction`, or produces/imports a block. So the transaction lifecycle (intrinsic gas, nonce, balance deduction, receipts), the block gas pool, the fork boundary actually being crossed, and the logs as they appear through receipts are all uncovered. One chain-level test would cover the lot:

```
block N-1 (pre-Jenner) -> block N (crosses the boundary, seeding runs)
-> block N+1: admin activates the asset feature
-> block N+2: user creates a token
-> block N+3: transfer / mint / a policy rejection
assert: state root agrees on re-import, receipts and logs are correct
```

Worth noting that this test would also have surfaced a gap in the seeding path: `SeedB20Activation` has a single production caller gated on `IsOnJenner`, true only on the transition block, so a chain whose genesis is already past Jenner never seeds at all. The current tests sidestep it by calling `SeedB20Activation` directly (`b20_policy_test.go`).

**One meta-observation.** `b20_basestd_diff_test.go:136` asserts `b20CobaltGaps = {0,0,0}` with the comment that "the scheduled multiplier and composite policies are both implemented", so nothing is deferred. That test diffs *signatures* only — and those two features are precisely the two largest behavioural holes above. The green there reads as coverage and isn't.

Happy to help write any of these if that's useful — items 1-3 and the chain-level test seem like the highest value for the effort.

---

*Reviewed by @joey0612 with Claude (Opus 5). Every absence above was confirmed by grep against the branch at `3c4eecb8` before being reported.*

joey0612:
**Suggestion: an aggregated token-info RPC endpoint**

Reading one B-20 token's basic metadata currently costs 6–8 separate `eth_call` round trips, all against the same address and all hitting adjacent storage:

| Call | Applies to |
|---|---|
| `name()` | both variants |
| `symbol()` | both |
| `decimals()` | both |
| `contractURI()` | both |
| `totalSupply()` | both |
| `supplyCap()` | both |
| `pausedFeatures()` | both |
| `policyId(scope)` × 6 | both — TRANSFER_SENDER / RECEIVER / EXECUTOR, MINT_RECEIVER, SEIZE_HOLDER / RECEIVER |
| `multiplier()` (+ `newUIMultiplier()` / `effectiveAt()`) | Asset only |
| `currency()` | Stablecoin only |

That is up to ~15 calls for a complete picture, and it is the picture a wallet or an explorer needs before it can render a single token correctly — the variant byte decides which balance call is even valid (`balanceOf` vs `scaledBalanceOf`), and the paused/policy state decides whether the Send button should be enabled at all.

Every one of those reads is a precompile dispatch against the same account, so the work is dominated by round-trip latency rather than by execution. A single method returning the whole set would collapse it to one:

```
b20_getTokenInfo(address, blockNumberOrHash) -> {
  variant,              // derived from byte 10 of the address, no state read
  name, symbol, decimals, contractURI,
  totalSupply, supplyCap,
  pausedFeatures,       // the bitmask, or the decoded feature list
  policies: { transferSender, transferReceiver, transferExecutor,
              mintReceiver, seizeHolder, seizeReceiver },
  multiplier,           // Asset only: current effective value
  pendingMultiplier,    // Asset only: { value, effectiveAt } or null
  currency,             // Stablecoin only
}
```

A batch form over an address list (`b20_getTokenInfoBatch`) would serve the "portfolio screen" case, which is the same problem multiplied by the number of tokens a user holds.

Two reasons I think this is worth doing here rather than leaving to integrators:

1. **The scheduled multiplier makes it more than a convenience.** Maturation is silent — the value flips at a timestamp with no transaction and no event (`b20_asset.go:128-136` says so explicitly, and calls out that a log-only indexer will diverge). Anything displaying an Asset balance therefore has to *poll* rather than follow logs, and polling three separate calls (`multiplier`, `newUIMultiplier`, `effectiveAt`) to reconstruct one consistent view is both slower and racy across blocks unless the caller pins a block. One endpoint at one block height removes the race.

2. **Pause state has no event either.** `isPaused` / `pausedFeatures` are poll-only by design, so the same argument applies: every wallet showing a B-20 token is going to poll this, forever, for every token.

Worth noting this would be genuinely additive rather than catching up: Base's node implementation has no B-20 RPC surface at all — none of its 123 custom RPC methods touch B-20, and no RPC crate references it — so precompile reads there go through plain `eth_call` exactly as they do here. This is a gap on both sides, not a divergence.

Scope-wise this is read-only, off the consensus path, and derives everything from state at a given block, so it does not need to ship with the fork — it could follow once the surface settles. Flagging it now mainly so the decision is deliberate rather than defaulted.

PR #3806, opened 2026-08-27
miner: revoke bidblock builder on cross-validator bad block evidence

flywukong:
## Description

BEP-675 currently revokes a builder only after it has burned **this node's own** block slot: `handleBidBlockResult` revokes when the sealed BidBlock fails `InsertChain`. A builder shipping broken blocks — a code bug, not necessarily malice — therefore costs every validator one slot before each of them independently blacklists it. On a 45-validator set that is 45 wasted slots.

This PR lets a validator act on the failing blocks **other** validators already sealed, so it can revoke such a builder before its own turn comes. Local revoke stays as the backstop; this only moves the revoke earlier. A builder is revoked once its failing blocks have spread across enough distinct validators — evidence other validators produced, not just this node's own.

- **Evidence**: a block carrying the BEP-675 header tag that fails execution counts as one vote against its builder, attributed to the validator that sealed it.
- **One vote per `(builder, sealer)`**, expiring on a 24h sliding window.
- **Revoke at N/2+1 distinct cabinet validators, or N/2+1  distinct validators of any role** → 24h lockout, the same window local revoke uses. The builder falls back to `SendBid`, which is simBid-protected.
- **Cabinet vs. candidate** is decided by the sealer's index in the mining validator set at the bad block's parent; the rotating tail is conservatively counted as candidates, so cabinet votes are never over-counted.

Thresholds are fixed rather than proportional. On Chapel the 15-validator tier is simply unreachable and the 6-cabinet tier applies, so there is no chain-specific code path.

### Rationale

### Example

### Changes

Notable changes: 
* ...

### DIFF
--- a/miner/bid_block.go
+++ b/miner/bid_block.go
@@ -236,6 +236,79 @@ func (w *worker) enqueueBidBlockTask(task *task, systemTxs int) {
 	}
 }
 
+// badBidBlockLoop revokes builders that have already burned enough other
+// validators, so this node need not be burned in turn.
+func (w *worker) badBidBlockLoop() {
+	defer w.wg.Done()
+	defer w.badBidBlockSub.Unsubscribe()
+
+	for {
+		select {
+		case ev := <-w.badBidBlockCh:
+			w.handleBadBidBlockEvidence(ev)
+		case <-w.badBidBlockSub.Err():
+			return
+		case <-w.exitCh:
+			return
+		}
+	}
+}
+
+// handleBadBidBlockEvidence records one validator's failing BidBlock and revokes
+// the builder once the evidence covers enough distinct validators.
+func (w *worker) handleBadBidBlockEvidence(ev core.BadBidBlockEvent) {
+	// Nil-checked: chain events reach nodes that never ran ApplyDefaultMinerConfig.
+	if !isTrue(w.config.Mev.Enabled) || !isTrue(w.config.Mev.BidBlockEnabled) {
+		return
+	}
+	if !w.permMgr.IsAllowed(ev.Builder) {
+		return
+	}
+	p, ok := w.engine.(*parlia.Parlia)
+	if !ok {
+		return
+	}
+	// At the parent, the state that decided this sealer's role. Not head: delivery
+	// is async, and head may already have crossed a breathe block.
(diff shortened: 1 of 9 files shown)


Jamstar00 (approved):

allformless (dismissed):

miner/bid_block.go
```diff
@@ -236,6 +236,79 @@ func (w *worker) enqueueBidBlockTask(task *task, systemTxs int) {
...
+	}
+	if role == parlia.ValidatorRoleNone {
+		// The seal was already verified, so this means a stale view, not a forgery.
+		log.Debug("[BID BLOCK EVIDENCE] sealer not in working validator set",
+			"sealer", ev.Sealer, "builder", ev.Builder, "number", ev.Number)
+		return
+	}
+
+	cabinetVotes, totalVotes := w.badBidBlocks.add(ev.Builder, ev.Sealer, role == parlia.ValidatorRoleCabinet)
+	if cabinetVotes < badBidBlockCabinetThreshold && totalVotes < badBidBlockTotalThreshold {
```
NathanBSC: then how to test it in devnet or testnet？

miner/bid_block_evidence.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	// Distinct validators that must have sealed a builder's failing blocks before
+	// it is revoked, counted over the cabinet alone and over every role. The
+	// cabinet threshold sits above the Byzantine tolerance of a 21-seat cabinet
+	// (f = 6), so framing a builder needs a colluding set larger than consensus
+	// already assumes could exist. Do not lower it back to f or below.
+	badBidBlockCabinetThreshold = 7
```
NathanBSC: N/2+1

miner/bid_block.go
flywukong: changed to n/2 +1

miner/bid_block_evidence.go
flywukong: fixed

NathanBSC (approved):

allformless (approved):

zlacfzy (approved):

PR #3810, opened 2026-09-01
 miner: record revoke count and add grpc flag

flywukong:
### Description

Simplify BidBlock revocation and enable the MEV gRPC service by default.

- ViolationCount is logged and persisted for observability, but does not affect revoke duration.

MEV gRPC
```
- Use port 8552 by default.
- Start only when Mev.BidBlockEnabled and Mev.Enable is enabled.
- Add --mev.grpc.disable to explicitly disable the service.
```

### Rationale

### Example

### Changes

Notable changes: 
* ...


miner/bid_block.go
```diff
@@ -278,6 +291,23 @@ func (w *worker) handleBidBlockResult(block *types.Block, task *task) {
...
+		// Escalate only when the chain rejected the block. InsertChain also fails
+		// on this node's own state or database trouble, which must not cost the
+		// builder a persisted 7-day lockout.
+		var (
+			duration       time.Duration
+			violationCount int
+		)
+		if w.chain.IsRejectedBidBlock(hash) {
+			duration, violationCount = w.revokeBidBlockBuilderEscalating(task.bidBlockInfo.builder, reason, hash, block.NumberU64())
+		} else {
```
NathanBSC: how is passible?

miner/bid_block_permission.go
NathanBSC: any where use this function?

NathanBSC:
too complex
let's keep it simple
not suggest to merge it

miner/bid_block.go
flywukong: yes, not allowed in bep-675, fixed

miner/bid_block_permission.go
flywukong: removed

allformless (approved):

zlacfzy (approved):

PR #3813, opened 2026-09-03
core/vm, params: CAS20 Compliant Asset Standard (BEP-702)

zlacfzy:
### Description

Spec context:
- BEP: [BEP-702: Compliant Asset Standard on BNB Smart Chain](https://github.com/bnb-chain/BEPs/blob/new_token/BEPs/BEP-702.md)

This PR adds BEP-702 CAS20 — the Compliant Asset Standard — to BSC: a fungible-token standard whose tokens are stateful precompiles rather than deployed contracts, with compliance policy part of the standard rather than reimplemented per issuer.

At a high level, it:
- adds a stateful-precompile host, so a precompile can read and write consensus state with metered gas
- adds a factory that derives a token's address from `(variant, creator, salt)` and marks it with a one-byte `0xEF` sentinel
- adds two token variants — Asset and Stablecoin — over a shared ERC-20, permit, role and pause surface
- adds two singletons: a PolicyRegistry for shared allow/blocklists, and an ActivationRegistry holding the per-feature switch
- gates all of it behind Jenner, with governance appointing the account that may open a feature

### Rationale

A BEP-20 transfer spends most of its gas on bytecode interpretation rather than on the state it touches. Moving the standard token into the client removes that overhead, lets one audited implementation serve every token of a variant, and lets compliance state be shared across tokens instead of reimplemented per issuer.

Nothing is reachable when the fork ships. The fork installs the switch and no feature starts activated, so this changes no behaviour until governance opens one.

### Activation authority

Governance appoints, the appointed admin operates:

```
governance --updateParam("admin", <20 bytes>)--> appoints the activation admin
admin      --activate / deactivate(bytes32)---> opens or closes a feature
```

Neither tier can do the other's job. Opening or closing is immediate rather than waiting out a voting period, while the authority behind it stays with chain governance — a lost key is replaced by a parameter-change proposal, not by a fork.

The first design instead took the admin from `ChainConfig` and seeded it at the fork boundary. Review found two defects: a chain whose genesis is already past the fork never crosses that boundary, so no feature could ever be opened; and the choice was one-shot, because only the incumbent could replace it. Both needed another fork to correct, and both disappear once the authority is a constant.

### Structure at a glance

```mermaid
flowchart TB
    subgraph HOST["Stateful-precompile host"]
        CTX["core/vm/contracts_stateful.go<br/>PrecompileContext, frame gas accounting"]
        DISP["core/vm/evm.go<br/>address resolution + dispatch"]
    end

    subgraph FAMILY["CAS20 family"]
        FAC["cas20_factory.go<br/>address derivation, create, initCalls"]
        TOK["cas20_token.go / _admin / _metadata / _permit<br/>ERC-20, roles, pause, EIP-2612"]
        VAR["cas20_asset.go / _stablecoin.go<br/>variant surfaces"]
        STO["cas20_storage.go / _gas.go<br/>ERC-7201 layout, EIP-2929/2200 metering"]
    end

    subgraph SINGLETONS["Singletons"]
        POL["cas20_policy.go<br/>PolicyRegistry"]
        ACT["cas20_activation.go<br/>ActivationRegistry"]
    end

    subgraph GATE["Fork gating"]
        PAR["params<br/>Jenner, GovHub constant"]
        UPG["core/systemcontracts<br/>plant registry sentinels"]
    end

    DISP --> CTX
    CTX --> STO
    DISP --> FAC
    FAC --> TOK
    TOK --> VAR
    TOK --> POL
    FAC --> ACT
    POL --> ACT
    PAR --> DISP
    PAR --> UPG
    UPG --> ACT
```

### Example

Creation and initialisation happen in one transaction — `initCalls` run inside a bootstrap window, so roles can be granted and supply minted before any role holder exists:

```
createCAS20(uint8 variant, bytes32 salt, bytes params, bytes[] initCalls) -> address
```

The returned address answers as an ERC-20 from then on, with no code deployed there beyond the sentinel.

### Reading token state

`eth_getCAS20TokenInfo(address, blockNrOrHash?)` returns a token's whole configuration from one state — name, symbol, decimals, contractURI, supply and cap, paused features, the six policy bindings, the EIP-712 domain separator, and for an Asset the *effective* multiplier with a pending schedule reported only while it is live; `eth_getCAS20TokenInfoBatch` does the same over up to twenty addresses, answering `null` for one that holds no token. It reads through the same storage accessors the precompile uses, off the consensus path, and replaces the dozen `eth_call`s a wallet otherwise issues per token.

### Changes

The commits split along the boundaries that compile independently:
- `params` — the Jenner fork gate and the GovHub constant the switch is reached through; `IsOnJenner` is written the way `IsOnPasteur` is, so a chain born after the fork carries the registry sentinels in its genesis allocation rather than through a block-1 special case
- `core/vm` — the host and the CAS20 family; one commit because the host's `AddLog` meters through a charge helper the family defines
- `core/systemcontracts` — the fork hook, which plants the two registry sentinels and nothing else; it is the only writer of a registry's sentinel, since GovHub tests `extcodesize` before forwarding the first appointment and a lazily planted byte could never be reached
- `internal/ethapi` — the two `eth_getCAS20TokenInfo*` methods

The rest answer review: the precompiles decode before any state check and revert empty on a decode failure, as Solidity does; `newUIMultiplier` falls back to the effective multiplier when nothing is scheduled; `getStringAt` grows its buffer only behind paid reads; each entry of an `announce` bundle or an `initCalls` array pays a warm `CALL` plus its own copy before it runs, so aliased entries are no longer free to dispatch; native storage access reaches the prestate and access-list tracers through a new `OnStorageRead` hook, and an `eth_call` code override on a CAS20 address disables its routing; every token method applies its checks in the order [BEP-702 §3.6](https://github.com/bnb-chain/BEPs/blob/new_token/BEPs/BEP-702.md#36-shared-token-interface) now tabulates; and the comments were cut to the why a name cannot carry.

### Notes

- Nothing is reachable at the fork: no feature starts activated, and only governance can appoint the account that opens one.
- A registry's sentinel is written only by the fork transition, at the start of the fork block in the same pass as `HistoryStorageAddress`; a token's only by `createCAS20`. Neither writes a nonce.
- Not in this PR: `CAS20Created` carrying `creator` and `salt` (a possible follow-up for indexers verifying a token's derivation); the node-deploy genesis seeding for chains born after the fork lives in that repository.


core/vm/cas20.go
```diff
@@ -0,0 +1,258 @@
...
+// isolated by the state stored under that address (BEP-702 §3.3).
+var cas20MarkerPrefix = [2]byte{0xca, 0x50}
+
+const (
+	cas20VariantAsset      = 0x00
+	cas20VariantStablecoin = 0x01
+
+	// cas20VariantMax is the highest ordinal the variant enum defines. A word above
+	// it is outside the enum, which is a different failure from a word inside it
+	// that no handler claims — see createCAS20.
```
will-2012: The annotations are a bit wordy; deleting them doesn't seem to affect understanding～

core/vm/cas20.go
```diff
@@ -0,0 +1,258 @@
...
+// hadNoCode reports whether addr carries no code at all — the condition under
+// which writing code owes the account-creation cost.
+func hadNoCode(state StateDB, addr common.Address) bool {
+	ch := state.GetCodeHash(addr)
+	return ch == (common.Hash{}) || ch == types.EmptyCodeHash
+}
+
+// ensureSentinel keeps a registry's storage out of reach of EIP-161 clearing. It
+// plants the marker only on an account with no code at all (BEP-702 3.16).
+func (ctx *PrecompileContext) ensureSentinel() {
```
will-2012: Isn't seed-sentinel a one-time operation performed during the hard fork?

core/vm/cas20_genesis.go
```diff
@@ -0,0 +1,28 @@
...
+// change, and the failed proposal would look like a successful one.
+func SeedCAS20Activation(state StateDB) {
+	seedCAS20Sentinel(state, CAS20ActivationRegistryAddress)
+	seedCAS20Sentinel(state, CAS20PolicyRegistryAddress)
+}
+
+// seedCAS20Sentinel gives a registry account the marker code that keeps it out of
+// reach of EIP-161 state clearing, unless it already carries code.
+func seedCAS20Sentinel(state StateDB, addr common.Address) {
+	if !hadNoCode(state, addr) {
```
will-2012, line 19: Log a Error log. IMO, this situation shouldn't happen—unless there's an address conflict.

params/config.go
```diff
@@ -733,7 +733,8 @@ type ChainConfig struct {
 	BPO4Time       *uint64 `json:"bpo4Time,omitempty"`       // BPO4 switch time (nil = no fork, 0 = already on bpo4)
 	BPO5Time       *uint64 `json:"bpo5Time,omitempty"`       // BPO5 switch time (nil = no fork, 0 = already on bpo5)
 	AmsterdamTime  *uint64 `json:"amsterdamTime,omitempty"`  // Amsterdam switch time (nil = no fork, 0 = already on amsterdam)
-	UBTTime        *uint64 `json:"ubtTime,omitempty"`        // UBT switch time (nil = no fork, 0 = already on UBT)
+
```
NathanBSC: why add empty line

params/config.go
```diff
@@ -1459,6 +1460,29 @@ func (c *ChainConfig) IsPasteur(num *big.Int, time uint64) bool {
...
+// where no false-to-true transition exists to find. Without it such a chain would
+// never plant the registries' account sentinels, and because GovHub refuses a
+// target carrying no code, the activation admin could then never be appointed —
+// leaving the reserved address space routed to precompiles that nothing can ever
+// open, correctable only by another fork.
+func (c *ChainConfig) IsOnJenner(currentBlockNumber *big.Int, lastBlockTime uint64, currentBlockTime uint64) bool {
+	if !c.IsJenner(currentBlockNumber, currentBlockTime) {
+		return false
+	}
+	if currentBlockNumber.Cmp(big.NewInt(1)) == 0 {
```
NathanBSC: strange

params/config.go
```diff
@@ -1459,6 +1460,29 @@ func (c *ChainConfig) IsPasteur(num *big.Int, time uint64) bool {
...
+// IsOnJenner reports whether this block is the one that crosses into Jenner, the
+// single block on which the fork's state changes are applied.
+//
+// Block 1 counts as that block on a chain whose genesis is already Jenner-active,
+// where no false-to-true transition exists to find. Without it such a chain would
+// never plant the registries' account sentinels, and because GovHub refuses a
+// target carrying no code, the activation admin could then never be appointed —
+// leaving the reserved address space routed to precompiles that nothing can ever
+// open, correctable only by another fork.
+func (c *ChainConfig) IsOnJenner(currentBlockNumber *big.Int, lastBlockTime uint64, currentBlockTime uint64) bool {
```
NathanBSC: maybe we can follow other defines:
1. move behind function `IsJenner`
2. follow the code `IsOnPasteur`

core/vm/cas20_storage.go
```diff
@@ -0,0 +1,504 @@
...
+	if length < 32 {
+		return "", true
+	}
+	base := s.stringDataRoot(slot)
+	// Before the allocation, not after: the loop below checks each iteration, but
+	// make() runs once with the stored length whatever the budget says.
+	if s.ctx != nil && s.ctx.OutOfGas() {
+		return "", false
+	}
+	out := make([]byte, 0, length)
```
flywukong: This allocates the full persisted string length before the tail-slot reads are charged. A caller can forward insufficient gas and repeatedly trigger a large Go allocation before the call runs out of gas, so EVM gas does not bound node resource usage. Please grow the buffer only after each paid read succeeds, or charge all required reads before allocating.

joey0612:
**Suggestion: an RPC that returns a CAS20 token's full metadata in one call — for both variants**

Reading one token's metadata today takes ~10–15 separate `eth_call`s against the same address, and there is no aggregated view in the precompile surface either (no `getTokenInfo` / `metadata()` — every field is its own selector: `name`, `symbol`, `decimals`, `contractURI`, `totalSupply`, `supplyCap`, `pausedFeatures`, `variantOf`, `eip712Domain`, `policyId(scope)` ×6, plus `multiplier` / `uiMultiplier` / `newUIMultiplier` / `effectiveAt` / `totalSupplyUI` for Asset and `currency` for Stablecoin).

Two properties of the standard make this more than a convenience:

1. **The scheduled multiplier matures with no transaction and no event** (spec §3.12; the code calls it out too). Anything showing an Asset balance therefore has to *poll*, and reconstructing one consistent view from `multiplier` + `newUIMultiplier` + `effectiveAt` across separate calls is racy across blocks unless the caller pins a block height.
2. **Pause and policy state have no events either** — `pausedFeatures` / `policyId` are poll-only by design.

So every wallet/explorer polls this, forever, for every token, over many round trips.

**Suggestion:** add a read-only JSON-RPC, e.g. `cas20_getTokenInfo(address, blockNumberOrHash)`, that returns the whole set in one call at one block height, **variant-tagged so a single response shape covers both an RWA/Asset and a Stablecoin**:

```
{
  variant,                 // from address byte[10], no state read
  name, symbol, decimals, contractURI,
  totalSupply, supplyCap,
  pausedFeatures,          // bitmask, or decoded feature list
  policies: { transferSender, transferReceiver, transferExecutor,
              mintReceiver, seizeHolder, seizeReceiver },
  eip712Domain,
  // Asset only:
  multiplier,
  pendingMultiplier: { value, effectiveAt } | null,
  totalSupplyUI,
  // Stablecoin only:
  currency
}
```

A batch form over an address list (`cas20_getTokenInfoBatch`) would serve the portfolio screen — the same problem multiplied by the number of tokens a user holds.

Scope-wise this is read-only, off the consensus path, and derives everything from state at a given block, so it does **not** need to ship with the fork — it can follow once the ABI surface settles. Worth noting it would be genuinely additive rather than catching up: the reference implementation has no CAS20/B20 RPC surface either, so metadata reads there go through plain `eth_call` exactly as they do here. Flagging it now mainly so the decision is deliberate rather than defaulted.

---
*Suggested by @joey0612 with Claude.*

will-2012 (dismissed):

joey0612 (approved):

joey0612 (dismissed):

will-2012 (dismissed):

joey0612 (dismissed):

joey0612 (approved):

will-2012 (approved):

PR #3816, opened 2026-09-07
chore: simply model

will-2012:
### Description

### Rationale

### Example

### Changes

Notable changes: 
* ...


core/stateless.go
```diff
@@ -62,6 +62,7 @@ func ExecuteStateless(ctx context.Context, config *params.ChainConfig, vmconfig
 		config:      config,
 		chainDb:     memdb,
 		headerCache: lru.NewCache[common.Hash, *types.Header](256),
+		numberCache: lru.NewCache[common.Hash, uint64](256),
```
will-2012: mark.

PR #3822, opened 2026-09-14
consensus/parlia: restore the one-second initial backoff for Jenner (BEP-714)

zlacfzy:
### Description
Client side of [BEP-714](https://github.com/bnb-chain/BEPs/pull/714) for the Jenner fork: return `InitialBackOffTime` from 2 s to 1 s on both backup paths. Client-only; no system contract changes. Activation time stays unscheduled.

### Rationale
BEP-520 raised the backup delay to 2 s at a 1.5 s block interval to limit reorgs from slow propagation. The interval has since dropped to 0.45 s while the delay stayed at 2 s, so an eight-block turn of an unavailable validator now costs ~19.6 s of affected intervals, 16 s of which is waiting. Restoring 1 s cuts that to ~11.6 s.

### Example
```bash
go test ./consensus/parlia -run 'Jenner|BackOff' -count=1
```

### Changes
* `consensus/parlia/parlia.go`: after Jenner (selected by parent time, like Lorentz) use `defaultInitialBackOffTime` for the first backup rank on both the normal path and the recently-signed in-turn path; rank spacing unchanged.
* `consensus/parlia/parlia_test.go`: backoff and timestamp-validation boundary at parent times just before, at, and after the fork, for both paths.


consensus/parlia/parlia.go
```diff
@@ -2428,11 +2434,12 @@ func (p *Parlia) backOffTime(snap *Snapshot, parent, header *types.Header, val c
 		})
 
 		if delay == 0 && isParerntLorentz {
```
will-2012, line 2436: ```if delay == 0 && (isParentLorentz || isParentJenner) ``` maybe more readable.

will-2012 (dismissed):

joey0612 (approved):

will-2012 (approved):

PR #3826, opened 2026-09-17
core/vm, internal/ethapi: CAS20 creation log order and bootstrap caller, drop dead StateOverride copy

zlacfzy:
### Description
Three CAS20 follow-ups from the BEP-702 fix review: `CAS20Created` is emitted before the initial admin grant and `initCalls`, the bootstrap window runs with the factory as `msg.sender`, and the unused `StateOverride` copy in `internal/ethapi/api.go` is removed (FR-05).

### Rationale
* The creation event is the token's identity record; emitting it last put it after the `RoleGranted` and `initCalls` logs it logically precedes.
* `initCalls` are calls the factory makes into the new token, so the token should see the factory as `msg.sender`, as for any contract the factory calls. Running them as the creator let the factory act in the creator's name (`approve`, `transfer`, `burn`, the renounce methods), a power no ordinary caller has.
* The `api.go` `StateOverride` copy was unreachable from RPC and lacked the `DisabledPrecompile` handling of the live `override.Apply`; a future caller pointed at it would silently lose CAS20 override routing. Its only test already exists in `internal/ethapi/override`.

Spec side: bnb-chain/BEPs PR for BEP-702 (linked in comments).

### Example
```
go test ./core/vm/ ./internal/ethapi/ ./internal/ethapi/override/ -count=1
```
`TestCAS20CreatedEvent` now pins: `CAS20Created` is the first log, the initial admin's `RoleGranted` follows, and its `sender` is the factory.

### Changes
* `core/vm/cas20_factory.go`: emit `CAS20Created` right after the token identity is written; `spawnBootstrap(addr, CAS20FactoryAddress)`; creation `RoleGranted` sender is the factory.
* `core/vm/cas20_factory_test.go`: pin log order and sender.
* `internal/ethapi/api.go`, `api_test.go`: drop the dead `OverrideAccount`/`StateOverride`/`Apply` copy, its duplicate test and the stub only it used.


zlacfzy:
Spec side: bnb-chain/BEPs#716

will-2012 (approved):
Logs will affect receipt root.

joey0612 (approved):

